PR #16, opened 2025-05-21
feat: Add comment_id as output to the Action

Texarkanine:
# Description

Expose the comment ID that the action creates, as an Action output.

This will allow users to identify the comment and interact with it in later steps in the workflow(s) where the action is used. This could allow hiding old comments, reacting to comments, filtering out Claude-created comments from other processing, notifying external services about Claude activities, etc.

Some of the above (or other) behaviors may make sense to include in the claude-code-action eventually, but simply exposing the created comment's ID is a backwards-compatible, nondisruptive, code-minimal way to enable users to build those use-cases *now*, without requiring any significant design decisions in the claude-code-action itself.

It's also probably a nice thing to do in general!

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,10 +84,10 @@ jobs:
 
 ## Outputs
 
-| Output           | Description                                   |
-| ---------------- | --------------------------------------------- |
-| `comment_id`     | ID of the comment created by Claude           |
-| `execution_file` | Path to the Claude Code execution output file |
+| Output           | Description                                             |
+| ---------------- | ------------------------------------------------------- |
+| `comment_ids`    | Newline-separated list of comment IDs created by Claude |
+| `execution_file` | Path to the Claude Code execution output file           |
 
 ## Examples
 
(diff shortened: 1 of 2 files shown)


ashwin-ant (changes requested):
@Texarkanine this is great, thank you! One suggestion: instead of `comment_id`, can you call this `comment_ids` and make it an array? I realize there's only one comment right now, but in the future we may allow Claude to post multiple comments, so making it an array now will make that transition easier.

Texarkanine:
GitHub Actions & workflows don't really have typed returns - they're all strings. "Array" could be represented in one of several ways:

* a space-separated list, e.g. `12345 67890 101123`
* a newline-separated list, e.g. `12345\n67890\n101123`
* a JSON array of... string? e.g. `["12345", "67890", "101123"]`

Personally, I think newline-separated will be the easiest for people to understand; I've updated and pushed that.

PR #48, opened 2025-05-25
Add OpenTelemetry environment variable support

ivancasco:
Pass OTEL configuration environment variables through to claude-code-base-action step to enable observability and tracing.

Supported OTEL env vars from Claude Code documentation:
- CLAUDE_CODE_ENABLE_TELEMETRY (required to enable telemetry)
- OTEL_METRICS_EXPORTER
- OTEL_EXPORTER_OTLP_PROTOCOL
- OTEL_EXPORTER_OTLP_ENDPOINT
- OTEL_EXPORTER_OTLP_HEADERS
- OTEL_EXPORTER_OTLP_METRICS_CLIENT_KEY
- OTEL_EXPORTER_OTLP_METRICS_CLIENT_CERTIFICATE
- OTEL_METRIC_EXPORT_INTERVAL
- OTEL_METRICS_INCLUDE_SESSION_ID
- OTEL_METRICS_INCLUDE_VERSION
- OTEL_METRICS_INCLUDE_ACCOUNT_UUID


sid374 (changes requested):
We should lean on the settings.json files (docs here: https://docs.anthropic.com/en/docs/claude-code/settings) to set up these env variables instead of defining them explicitly in the action. 

The env vars you have listed here are only a small subset of the supported OTEL env vars, and we'd end up writing more config here every time we discover a new one that's useful. Instead the flow should be:
1. Create a .claude/settings.json in your repo and check it in or create it as a separate workflow step before calling the claude action if you'd like to use Secrets
2. Add this to your settings.json + any additional OTEL env vars
```
  "env": {
    "CLAUDE_CODE_ENABLE_TELEMETRY": "1",
    "OTEL_METRICS_EXPORTER": "otlp"
  }
```
3. Continue...

We will also update our docs to reflect this.

Issue #14, opened 2025-05-21
[Bug] .mcp.json not picked up and detected


andaag:
Hi

I created an .mcp.json file in a test repository, but it's not picked up. I see claude-code-base-action has a mcp_config that can be set - but that can't be set in this repository.

How is this meant to be wired together?

andaag:
prepareMcpConfig seems to me to override any mcp config in the directory when setting up it's tools.

shai347:
I'm experiencing the same issue. I tried to integrate the Brave Search API with MCP and spent the whole day struggling with it until I gave up. I tried every possible setup, but Claude still isn't able to use MCP.

ashwin-ant:
Hi all, we're aware of this issue and working on a fix. Appreciate the patience here.

ashwin-ant:
This should be fixed now in the latest release (if you're on `beta` you'll get it immediately). `.mcp.json` files should now be respected properly.

PR #65, opened 2025-05-27
Merge existing MCP server configurations from workspace

ShintaroaSuzuki:
Add functionality to read and merge existing .mcp.json configurations from the GitHub workspace, preserving user-defined MCP servers while ensuring our default GitHub server configuration takes precedence.

#14

### DIFF
--- a/src/mcp/install-mcp-server.ts
+++ b/src/mcp/install-mcp-server.ts
@@ -1,10 +1,12 @@
 import * as core from "@actions/core";
+import * as fs from "fs";
+import * as path from "path";
 
 export async function prepareMcpConfig(
   githubToken: string,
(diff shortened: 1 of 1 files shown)


ShintaroaSuzuki:
Confirmed that the MCP server connection works when the following `.mcp.json` is placed in the root directory
```json
{
  "mcpServers": {
    "figma-developer-mcp": {
      "command": "npx",
      "args": ["-y", "figma-developer-mcp", "--stdio"]
    }
  }
}
```
image

ashwin-ant:
Thanks for this PR! This is actually a bug in Claude Code that we have a fix on the way for. So this PR won't be necessary.

PR #94, opened 2025-05-30
Add issue labeled event

piersonmarks:
This PR adds support for the `ISSUE_LABELED` event. Now Claude Code can detect when a Github issue has a label added and handles it properly. This is useful for creating label-based workflows. For example, adding a "auto-fix" label to trigger a Claude Code session.

### Changes

-  Updated code to recognize and process labeled issue events.
-  Added a new type for labeled issues to keep things organized.

## Testing 
- Tested on our own repository and correctly saw labeled issues trigger Claude Code and not throw the error. Correctly extracts the label name and logs `Event type: ISSUE_LABELED with '<name>'`

### DIFF
--- a/src/create-prompt/index.ts
+++ b/src/create-prompt/index.ts
@@ -260,6 +260,10 @@ export function prepareContext(
           claudeBranch,
         };
       } else if (eventAction === "labeled") {
+        const payload = context.payload;
+        const labelName =
+          "label" in payload && payload.label ? payload.label.name : "";
+
         eventData = {
           eventName: "issues",
           eventAction: "labeled",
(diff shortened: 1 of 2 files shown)


ashwin-ant:
Thanks for submitting this! I think this isn't quite sufficient though, you'll need to update the trigger check logic in https://github.com/anthropics/claude-code-action/blob/main/src/github/validation/trigger.ts#L13-L60. That's where we check for an `@-claude` mention. You may want to do something similar to the `assignee_trigger` handling there, and then add that as an input to the action like we do [here](https://github.com/anthropics/claude-code-action/blob/main/action.yml#L12-L14). The default value could be `claude`.

piersonmarks:
ah thanks for the guidance - it was working but maybe wasn't through enough. Will fix!

ashwin-ant:
Done in https://github.com/anthropics/claude-code-action/pull/177

PR #115, opened 2025-06-03
fix: skip SHA computation for deleted files

rjwalters:
When processing PRs that delete files, the action attempts to compute SHA for files that no longer exist, resulting in errors:

```
fatal: could not open '.github/workflows/claude-review-type-annotations.yml' for reading: No such file or directory
Failed to compute SHA for .github/workflows/claude-review-type-annotations.yml
```

While these errors are non-fatal (caught and handled), they create confusing output and could mask real issues.

## Solution

This PR checks the file's `changeType` from the GitHub GraphQL response before attempting to compute its SHA. Deleted files are skipped with a special "deleted" SHA value.

## Changes

- Check if `file.changeType === 'DELETED'` before computing SHA
- Return a special SHA value of "deleted" for deleted files
- Add informative console message when skipping deleted files

## Testing

- [x] Formatting passes (`bun run format:check`)
- [x] Type checking passes (`bun run typecheck`)
- [x] Test passes (`bun run test`)

### DIFF
--- a/src/github/data/fetcher.ts
+++ b/src/github/data/fetcher.ts
@@ -1,17 +1,17 @@
 import { execSync } from "child_process";
+import type { Octokits } from "../api/client";
+import { ISSUE_QUERY, PR_QUERY } from "../api/queries/github";
 import type {
-  GitHubPullRequest,
-  GitHubIssue,
   GitHubComment,
   GitHubFile,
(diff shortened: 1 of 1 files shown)


ashwin-ant (approved):
Thank you!

PR #145, opened 2025-06-07
fix: Remove deprecated tools from allowed_tools to align with what is available in github-mcp-server

sepehr500:
The current examples for allowed_tools include mcp__github__add_pull_request_review_comment, which is no longer provided by [github/github-mcp-server](https://github.com/github/github-mcp-server). This causes confusion for the GitHub Action, as the tool has been deprecated.

This tool was replaced by several new tools, as outlined in [this PR](https://github.com/github/github-mcp-server/pull/410). I’ve updated the examples and the "Auto Review PRs" action to use the new tools accordingly.

I discovered this issue after multiple failed attempts to get Claude to add inline comments—it kept throwing errors due to the missing tool. Updating the allowed_tools list resolved the problem.

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -29,4 +29,4 @@ jobs:
 
             Be constructive and specific in your feedback. Give inline comments where applicable.
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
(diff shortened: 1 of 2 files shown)


noachr (approved):

ashwin-ant (approved):
Thank you!

PR #166, opened 2025-06-12
fixed file ingestion

dioptre:
fixed breaking commits for binary files

### DIFF
--- a/src/mcp/github-file-ops-server.ts
+++ b/src/mcp/github-file-ops-server.ts
@@ -506,6 +506,7 @@ server.tool(
 
       const octokit = new Octokit({
         auth: githubToken,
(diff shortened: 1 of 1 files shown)


derekbreden:
I confirmed this change fixes an issue I was having with binary files. Apologies for not being more helpful, but I just wanted to share the anecdotal data point that there does seem to be an issue with binary files, and this does seem to fix it.

Thanks @dioptre 

https://github.com/derekbreden/truce-web/pull/102/files#diff-192bd4ecbfca54b19db79b65406a0db05870cf118af93c34a21ba2778e8a2e50

ashwin-ant:
@dioptre this looks good but can you fix the CI issues? `bun run format` and `bun run typecheck`

PR #201, opened 2025-06-25
chore(README): document `base_branch` parameter

stefanoamorelli:
[`base_branch` is already configurable](https://github.com/anthropics/claude-code-action/blob/main/src/github/context.ts#L37) but not documented in `README.md` - in this PR we add the documentation.


ashwin-ant (approved):
Thanks!

PR #215, opened 2025-06-29
feat(examples): add auto-update workflow for CLAUDE.md files

ps428:
Add new example workflow `claude-update-claude-md.yml` that automatically updates CLAUDE.md documentation files when code changes are pushed to main/master branches.

## What it does

- Automatically finds all CLAUDE.md files in the repository
- Analyzes commit messages and diffs to determine which sections need updating
- Updates only relevant documentation sections while preserving existing structure
- Creates PRs with the documentation updates for review

## Use case

This workflow is particularly useful for:
- Keeping documentation in sync with code changes
- Monorepos with multiple nested CLAUDE.md files
- Teams that want to automate documentation updates

## Implementation

The workflow triggers on pushes to main/master (excluding markdown changes to prevent loops) and uses Claude to intelligently update documentation based on the nature of the changes.

Resolves the need for manual documentation updates after code changes.

### DIFF
--- a/examples/claude-update-claude-md.yml
+++ b/examples/claude-update-claude-md.yml
@@ -1,0 +1,68 @@
+name: Auto-Update CLAUDE.md Files
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+
+jobs:
+  update-claude-md:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+      id-token: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - name: Update CLAUDE.md files
+        uses: anthropics/claude-code-action@beta
+        with:
+          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
+          timeout_minutes: "30"
+          branch_prefix: "claude-md-update/"
+          direct_prompt: |
+            Analyze the pull request changes and update all CLAUDE.md files in the repository to reflect these changes.
+
+            IMPORTANT INSTRUCTIONS:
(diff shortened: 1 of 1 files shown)


examples/claude-update-claude-md.yml
```diff
@@ -0,0 +1,75 @@
...
+
+on:
+  push:
+    branches:
+      - main
+      - master
+    # Ignore documentation changes to prevent infinite loops
+    paths-ignore:
+      - "**/*.md"
+      - ".github/workflows/claude-update-claude-md.yml"
```
ashwin-ant: Have you tested this? I'm not sure this will actually work (the Claude app doesn't run on `push` event triggers).

examples/claude-update-claude-md.yml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - name: Update CLAUDE.md files
+        uses: anthropics/claude-code-action@beta
+        with:
+          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
+          timeout_minutes: "30"
```
Dadarkp3, line 24: Just for consistence 'timeout_minutes' should be passed as a number, not a string.

examples/claude-update-claude-md.yml
ps428: Hey Ashwin, thanks for the review!

I updated the code to run the workflow on PR [create & sync] instead of push. I am testing it on a personal repo. I'll update again once it works properly on that.

I'm making this PR a draft until it's ready.

examples/claude-update-claude-md.yml
ps428: Well spotted Daiane,
I'll update that, thanks!

PR #211, opened 2025-06-27
feat: add `sticky_comment` input to reduce GitHub comment spam

valorkin:
## 🎯 Summary
Adds a new `sticky_comment` input that enables Claude to use a single persistent comment instead of creating multiple comments, reducing notification spam and improving user experience.

## ✨ What's Changed
- ✅ Added `sticky_comment` input parameter (default: `false`)
- ✅ Updated environment variable handling (`STICKY_COMMENT`)
- ✅ Enhanced comment logic to reuse existing Claude comments when enabled
- ✅ Updated documentation in README.md

## 🔧 Technical Details
- **Input**: `sticky_comment: true/false` 
- **Behavior**: When enabled on PRs, Claude updates existing comments instead of creating new ones
- **Fallback**: Graceful fallback to regular comments if no existing comment found
- **Scope**: Only affects PR comments (issues still get new comments as normal)

## 🧪 Testing
- [x] TypeScript compilation passes
- [x] All existing functionality preserved
- [x] Non-breaking change (backward compatible)

## 📚 Usage Example
```yaml
- uses: anthropics/claude-code-action@main
  with:
    sticky_comment: true  # Enable sticky comment mode
    # ... other inputs
```

## 🎉 Benefits
- 📢 **Reduces notification spam** for PR participants  
- 🧹 **Cleaner comment threads** with single updating comment
- ⚡ **Faster reviews** with less comment clutter
- 🔄 **Backward compatible** - existing workflows unchanged

Ready for review! 🚀

PR #226, opened 2025-07-03
fix sticky comment variable name

ashwin-ant:
followup to https://github.com/anthropics/claude-code-action/pull/211/


chrislloyd (approved):

domenkozar:
Note that if you delete all claude comments, the tool fails with an error.

ashwin-ant:
> Note that if you delete all claude comments, the tool fails with an error.

It should be creating a new comment in that case. Are you not seeing that?

PR #227, opened 2025-07-03
fix(github): fixing claude login user name

ryok90:
Simply fixing claude's login name change to match when doing sticky comments.

Following a fix from a [previous PR](https://github.com/anthropics/claude-code-action/pull/211) to allow sticky comments, one unwanted change was applied, and this PR aims to revert that change.

### DIFF
--- a/src/github/operations/comments/create-initial.ts
+++ b/src/github/operations/comments/create-initial.ts
@@ -14,6 +14,8 @@ import {
 } from "../../context";
 import type { Octokit } from "@octokit/rest";
 
+const CLAUDE_APP_BOT_ID = 209825114;
+
 export async function createInitialComment(
   octokit: Octokit,
   context: ParsedGitHubContext,
@@ -37,7 +39,7 @@ export async function createInitialComment(
         issue_number: context.entityNumber,
       });
       const existingComment = comments.data.find((comment) => {
-        const idMatch = comment.user?.id === 209825114;
+        const idMatch = comment.user?.id === CLAUDE_APP_BOT_ID;
         const botNameMatch =
           comment.user?.type === "Bot" &&
           comment.user?.login.toLowerCase().includes("claude");
--- a/src/github/operations/comments/create-initial.ts
+++ b/src/github/operations/comments/create-initial.ts
@@ -36,11 +36,15 @@ export async function createInitialComment(
         repo,
         issue_number: context.entityNumber,
       });
-      const existingComment = comments.data.find(
-        (comment) =>
-          comment.user?.login.indexOf("claude") !== -1 ||
-          comment.body === initialBody,
-      );
+      const existingComment = comments.data.find((comment) => {
+        const idMatch = comment.user?.id === 209825114;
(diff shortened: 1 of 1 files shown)


src/github/operations/comments/create-initial.ts
```diff
@@ -38,7 +38,7 @@ export async function createInitialComment(
       });
       const existingComment = comments.data.find(
         (comment) =>
-          comment.user?.login.indexOf("claude[bot]") !== -1 ||
+          comment.user?.login.indexOf("claude") !== -1 ||
```
ndouglas: Will this work if Claude is using another GITHUB_TOKEN? For instance, in my setup I have Claude using a token from a different GitHub App.

ryok90: It should. Using just `claude` instead of `claude[bot]` will match various Claude-related usernames:
- claude[bot] (default GitHub App)
- claude-bot (custom GitHub App)
- claude-assistant (personal access token user)
- Any username containing "claude"

The code also has a fallback that matches by comment body content (`comment.body === initialBody`), so
  even if the username doesn't match, it can still find existing comments.

ndouglas: Would it be appropriate to use the `trigger_phrase` input here?

ashwin-ant: Hm, I'd rather not simply match on any username containing claude. How about if we check for both `claude` and also `user.type === 'Bot'`? That way we won't trigger on people who happened to be named Claude.

@ndouglas I think the `trigger_phrase` match is an interesting idea. Maybe we'd check to see if the name contains the trigger phrase (after stripping any non-alphanumeric characters) and then also check to make sure it's a bot.

ryok90: Got it.
I will check to see if I can make these changes work.
Thank you!

ashwin-ant: @ryok90 another option to consider is user id matching. The Claude bot user's id is `209825114`. So an input to the action like `claude_user_id` that defaults to `209825114` might be okay.

ryok90: How about now?

src/github/operations/comments/create-initial.ts
```diff
@@ -36,11 +36,15 @@ export async function createInitialComment(
         repo,
         issue_number: context.entityNumber,
       });
-      const existingComment = comments.data.find(
-        (comment) =>
-          comment.user?.login.indexOf("claude[bot]") !== -1 ||
-          comment.body === initialBody,
-      );
+      const existingComment = comments.data.find((comment) => {
+        const idMatch = comment.user?.id === 209825114;
```
ashwin-ant: Cool! Let's move this to a constant at the top of the file like `CLAUDE_APP_BOT_ID`

ryok90: Done!

ashwin-ant (approved):
Thank you!

PR #245, opened 2025-07-10
chore: run cd command before git commands

ogadra:
When running this action on a Self-hosted runner, the following error occurs:

```bash
fatal: not a git repository (or any of the parent directories): .git
```

This error happens because git commands were executed in the wrong directory. To resolve this issue, the git commands are explicitly executed in the directory specified by process.env.GITHUB_WORKSPACE.

This fix ensures the git operations always run within the correct repository context, eliminating the error and ensuring compatibility with self-hosted runner environments.

### DIFF
--- a/src/github/operations/branch.ts
+++ b/src/github/operations/branch.ts
@@ -53,6 +53,11 @@ export async function setupBranch(
         `PR #${entityNumber}: ${commitCount} commits, using fetch depth ${fetchDepth}`,
       );
 
+      if (!process.env.GITHUB_WORKSPACE) {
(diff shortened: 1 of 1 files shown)


ad-m-ss:
Why do you think this is a correct assumption? `actions/checkout@v4` supports the "path" parameter to clone code to a subdirectory.

ogadra:
@ad-m-ss I'm sorry, there was no issue with `claude-code-action` itself; the problem was caused by a step executed earlier in the workflow. I've confirmed it works without this PR, so I'll close it.

Issue #223, opened 2025-07-02
Claude Code Action fails with "couldn't find remote ref" error for PRs from forks


gabrielguimaraes:
**Describe the bug**
The Claude Code Action fails when triggered on pull requests from forked repositories. The action attempts to fetch a branch that doesn't exist on the origin remote, resulting in the error: `fatal:
couldn't find remote ref [branch-name]`. This occurs because PRs from forks only exist as PR refs (e.g., `refs/pull/22/head`) and not as regular branches on the origin remote.

**To Reproduce**

1. Create a fork of a repository that uses Claude Code Action for PR reviews
2. Create a new branch in the fork and make changes
3. Open a pull request from the fork to the original repository
4. The Claude Code Action workflow triggers automatically
5. See error: `fatal: couldn't find remote ref [branch-name]`

**Expected behavior**
The Claude Code Action should successfully checkout and review PRs from forked repositories, just as it does for PRs from branches within the same repository. The action should use GitHub's PR refs
(`refs/pull/*/head`) instead of trying to fetch non-existent branch names.

**Screenshots**
Error from GitHub Actions logs:
This is an open PR, checking out PR branch...
PR #22: 1 commits, using fetch depth 20
fatal: couldn't find remote ref adding_claude_code_action
Error: Prepare step failed with error: Failed with exit code 128
Error: Process completed with exit code 1.

**Workflow yml file**
```yaml
name: Claude Code Review

on:
  pull_request:
    types: [opened, synchronize]

jobs:
  claude-review:
    runs-on: ubuntu-latest
    permissions:
      contents: read
      pull-requests: write
      issues: write
      id-token: write

    steps:
      - name: Checkout repository
        uses: actions/checkout@v4
        with:
          fetch-depth: 1

      - name: Run Claude Code Review
        id: claude-review
        uses: anthropics/claude-code-action@beta
        with:
          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
          github_token: ${{ secrets.GITHUB_TOKEN }}

          direct_prompt: |
            Please review this pull request and provide feedback on:
            - Code quality and best practices
            - Potential bugs or issues
````

**API Provider**

[x] Anthropic First-Party API (default)
[ ] AWS Bedrock
[ ] GCP Vertex

**Additional context**
- The issue appears to be in the branch checkout logic at https://github.com/anthropics/claude-code-action/blob/main/src/github/operations/branch.ts
[...]

PR #253, opened 2025-07-12
fix: use `gh` to checkout PR (which works for forks)

### LINKED ISSUE #46: Comments in (forked) PR trigger claude-code-action with wrong branch
I am using the https://github.com/anthropics/claude-code-action/blob/main/examples/claude.yml workflow.

According to https://github.com/orgs/community/discussions/25917, 

> Comments on issues and pull requests will have the ref [set to the default branch](https://help.github.com/en/articles/events-that-trigger-workflows).  However, you should be able to inspect the full [github event context ](https://help.github.com/en/articles/contexts-and-expression-syntax-for-github-actions#contexts)for the information about the pull request in question.

Workflow triggered by issue_comment (thus a comment in PR) will checkout to the base branch, which provides wrong context for claude-code to work with.

What is the suggested way to fix this?

- Prompt Claude to let it checkout to the PR branch automatically?
- Make the action `actions/checkout@v4`  checkout to the correct branch if on PR comments?

huweiATgithub:
resolves #46, #223

This PR is incomplete now. Because, to make `gh` to work, environment variables `GH_TOKEN` is also needed.
If this PR is ok, I can setup those.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -129,6 +129,11 @@ runs:
       id: prepare
       shell: bash
       run: |
+        # Change to CLAUDE_PREPARE_DIR if set (for running in custom directories)
+        if [ -n "$CLAUDE_PREPARE_DIR" ]; then
+          echo "Changing directory to CLAUDE_PREPARE_DIR: $CLAUDE_PREPARE_DIR"
+          cd "$CLAUDE_PREPARE_DIR"
+        fi
         bun run ${GITHUB_ACTION_PATH}/src/entrypoints/prepare.ts
       env:
         MODE: ${{ inputs.mode }}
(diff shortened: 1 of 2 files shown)


coolwhipmike:
https://github.com/anthropics/claude-code-action/blob/main/src/mcp/github-file-ops-server.ts would also need to get updated to handle PRs from forked repos.

huweiATgithub:
> https://github.com/anthropics/claude-code-action/blob/main/src/mcp/github-file-ops-server.ts would also need to get updated to handle PRs from forked repos.

I have gone through the codebase. It seems that the entire codebase has the implicit assumption of using repo's branches.
It is not easy to make it work with PR from a forked branch.
One needs to at least modify:

- the prepared context: https://github.com/anthropics/claude-code-action/blob/main/src/github/context.ts#L105
- related MCP servers
- generated prompts

Personally, I don't think this can be done (correctly and in a maintainable way) without abstracting concepts like head branch and base branch.

PR #259, opened 2025-07-13
Add Squid proxy network restrictions for claude-code-action

km-anthropic:
## Summary
Adds Squid proxy network restrictions to the Claude workflow to enhance security by limiting network access to only approved domains.

## Changes
- Added Squid proxy setup to `.github/workflows/claude.yml` with domain whitelisting

## Testing
Extensively tested on fork:
- Claude successfully responds to mentions and performs file operations


.github/workflows/claude.yml
```diff
@@ -29,6 +29,60 @@ jobs:
...
+          # GitHub Container Registry
+          ghcr.io
+
+          # Package registries
+          registry.npmjs.org
+          bun.sh
+
+          # Azure storage for GitHub Actions cache
+          .blob.core.windows.net
+          EOF
```
ashwin-ant: Can we add a way for users to specify additional domains? Also wondering if we should make it so you can start from no domains at all (for users who use Bedrock rather than 1st party Anthropic API, for example).

I'm also wondering if this needs to be on by default, since that's going to break existing configs. Instead, how about we make it so that it's disabled by default, but people can add a list of domains as an input. We can provide this list as an example in the readme and in https://github.com/anthropics/claude-code-action/tree/main/examples.

ashwin-ant (changes requested):

.github/workflows/claude.yml
km-anthropic: So users can just add additional domains by writing it from line 38 - line 58 (but we can make it a github env var)

I think what we can do is the following:

1. Make it disabled by default
2. Add necessary installation ones to `NO_PROXY`
3. I can double check what the endpoints are for bedrock / GCP and so users can just copy the filters they need

test/network-restrictions.test.ts
ashwin-ant, line 1: These tests wouldn't actually test the change, would they? The test job doesn't run with the network restriction on.

action.yml
```diff
@@ -142,6 +146,55 @@ runs:
...
         USE_COMMIT_SIGNING: ${{ inputs.use_commit_signing }}
 
+    - name: Setup Network Restrictions
+      if: steps.prepare.outputs.contains_trigger == 'true' && inputs.allowed_domains != ''
+      shell: bash
+      run: |
+        # Install and configure Squid proxy
+        sudo apt-get update && sudo apt-get install -y squid
+
+        echo "${{ inputs.allowed_domains }}" > /tmp/whitelist.txt
```
ashwin-ant: Instead of `/tmp` let's use `RUNNER_TEMP`

.github/workflows/claude.yml
km-anthropic: Resolved!

test/network-restrictions.test.ts
km-anthropic: So the main reason here was to instruct @.claude to use that test (to confirm restrictions were working). It can be thrown away now

action.yml
```diff
@@ -100,6 +100,10 @@ inputs:
     description: "Enable commit signing using GitHub's commit signature verification. When false, Claude uses standard git commands"
     required: false
     default: "false"
+  allowed_domains:
```
ashwin-ant: Should we call it `experimental_allowed_domains` for now so that it's clear that this is still pretty new?

ashwin-ant (dismissed):

.github/workflows/claude.yml
```diff
@@ -31,9 +31,14 @@ jobs:
...
         id: claude
-        uses: anthropics/claude-code-action@beta
+        uses: ./
         with:
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
           allowed_tools: "Bash(bun install),Bash(bun test:*),Bash(bun run format),Bash(bun typecheck)"
           custom_instructions: "You have also been granted tools for editing files and running bun commands (install, run, test, typecheck) for testing your changes: bun install, bun test, bun run format, bun typecheck."
           model: "claude-opus-4-20250514"
+          # Test network restrictions
+          allowed_domains: |
```
ashwin-ant: nit: update to new name

ashwin-ant (commented):
CI is failing due to formatting. Also left a minor comment

ashwin-ant (approved):
Excited for this!

PR #266, opened 2025-07-14
feat: Add configuration option to delete the progress/results comment at the end of the workflow

seamuswn:
We are using the Claude code action to automatically generate the PR body/description for developers by leveraging `direct_prompt` to run `gh pr edit --body`.

As a result the comment left by this workflow has very little value for us and just takes up space / makes noise on the PR screen.

This changes provides the ability to have the claude-code-action workflow delete the comment at the end of the workflow when users set `delete_comment_on_completion: true`, so the resulting experience is that they see progress reports in the comment while the workflow runs, but it's cleaned up by the end.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -96,6 +96,10 @@ inputs:
     description: "Enable commit signing using GitHub's commit signature verification. When false, Claude uses standard git commands"
     required: false
     default: "false"
+  delete_comment_on_completion:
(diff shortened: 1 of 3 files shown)


ashwin-ant:
An alternative approach here is modes: https://github.com/anthropics/claude-code-action/discussions/225. The `tag` mode (which is essentially the only mode we support today) is really the only one that needs a tracking comment. Other modes wouldn't have a comment.

seamuswn:
@ashwin-ant until such a time where we have modes / separate actions set up for this repo - is it okay to get this reviewed and merged?

PR #284, opened 2025-07-16
Avoid ambiguous name that can be either a branch or a file

ujihisa:
git-checkout takes either branch, commit, or pathspec. https://git-scm.com/docs/git-checkout
The line in this pull request tries to checkout a branch, but when you have a branch which name is the same as an existing file/dir it can cause a runtime error. It's recommended to use `--` for argument disambiguation[^1].

[^1]: https://git-scm.com/docs/git-checkout#_argument_disambiguation


ashwin-ant:
Fixed by https://github.com/anthropics/claude-code-action/pull/306, but thank you!

ujihisa:
Thanks! Oops the position here was wrong :sweat_smile:

PR #298, opened 2025-07-18
Add override prompt variable

km-anthropic:
Add the override prompt variable that will also swap in referenced items like $PR_COMMENTS, with the actual value as determined by the user.

@claude


km-anthropic:
@claude can you review?

km-anthropic:
@claude can you review?

PR #301, opened 2025-07-18
Add override prompt variable

km-anthropic:
Add the override prompt variable that will also swap in referenced items into overridePrompt, with the actual value as determined by the user.

@claude


src/create-prompt/index.ts
```diff
@@ -460,11 +462,65 @@ function getCommitInstructions(
...
+  context: PreparedContext,
+  githubData: FetchDataResult,
+): string {
+  const { contextData, comments, reviewData, changedFilesWithSHA } = githubData;
+  const { eventData } = context;
+
+  const variables: Record<string, string> = {
+    REPOSITORY: context.repository,
+    PR_NUMBER: eventData.isPR ? (eventData as any).prNumber || "" : "",
+    ISSUE_NUMBER: !eventData.isPR ? (eventData as any).issueNumber || "" : "",
```
ashwin-ant: Can we avoid `any`s?

src/create-prompt/index.ts
```diff
@@ -460,11 +462,65 @@ function getCommitInstructions(
...
+      ? formatReviewComments(reviewData, githubData.imageUrlMap)
+      : "",
+    CHANGED_FILES: eventData.isPR
+      ? formatChangedFilesWithSHA(changedFilesWithSHA)
+      : "",
+    TRIGGER_COMMENT: (eventData as any).commentBody || "",
+    TRIGGER_USERNAME: context.triggerUsername || "",
+    BRANCH_NAME:
+      (eventData as any).claudeBranch || (eventData as any).baseBranch || "",
+    BASE_BRANCH: (eventData as any).baseBranch || "",
```
ashwin-ant: Here too, let's try to be typesafe.

ashwin-ant (changes requested):
couple of small comments on the TS style

src/create-prompt/index.ts
katchu11: resolved

src/create-prompt/index.ts
katchu11: resolved

ashwin-ant (approved):

PR #306, opened 2025-07-19
fix: git checkout disambiguate error

### LINKED ISSUE #305: Action fail on PR when a directory of same name with branch exists
**Describe the bug**
Suppose that name of the branch is `foo`. when there is a change on direcory `foo` on pull request, action fails with below error message on git checkout

```
fatal: 'foo' could be both a local file and a tracking branch.
Please use -- (and optionally --no-guess) to disambiguate
```

**To Reproduce**
minimal example: https://github.com/palindrom615/claude-code-issue-report/pull/1

**Expected behavior**
git checkout command should be unambigous to avoid the error

**Screenshots**
-

**Workflow yml file**
included in minimal example above

**API Provider**

[ ] AWS Bedrock
[ ] GCP Vertex

**Additional context**
The cause of error is due to `git checkout ${branchName}` is disambiguate. check https://git-scm.com/docs/git-checkout#_argument_disambiguation

palindrom615:
Fix #305

### DIFF
--- a/src/github/operations/branch.ts
+++ b/src/github/operations/branch.ts
@@ -55,7 +55,7 @@ export async function setupBranch(
 
       // Execute git commands to checkout PR branch (dynamic depth based on PR size)
       await $`git fetch origin --depth=${fetchDepth} ${branchName}`;
-      await $`git checkout ${branchName}`;
+      await $`git checkout ${branchName} --`;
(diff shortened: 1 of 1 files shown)


ashwin-ant (approved):
Thank you!

Issue #303, opened 2025-07-18
claude should be able to cd to the repo directory


berenddeboer:
**Describe the bug**
I've asked Claude to go to the repo directory, but it complains with:

```
 "cd to '/home/runner/work/monorepo/monorepo' was blocked.
For security, Claude Code may only change directories to child directories of the allowed working directories
for this session (including '/home/runner/work/_actions/anthropics/claude-code-action/beta/base-action')."
```

**Expected behavior**
No clue why this is not allowed. Now I see it making "git -C" calls.

**Workflow yml file**
Pretty much the default from the `example/claude.yml`.

**API Provider**

[X] Anthropic First-Party API (default)
[ ] AWS Bedrock
[ ] GCP Vertex

**Additional context**
Using my Max subscription, not an API key at the moment.

pirate:
I'm seeing this as well on all the runs today: https://github.com/browser-use/browser-use/actions/runs/16381788009

<img width="2717" height="740" alt="Image" src="https://github.com/user-attachments/assets/a807ffbf-bb44-4445-996d-427b4b160ce1" />

berenddeboer:
Related to this is Claude code review, it's working in its own directory, and spending thousands of tokens figuring out why it can't access my files:

```
{
        "type": "thinking",
        "thinking": "I see, I'm currently in the `base-action` directory which is part of the action itself, 
not the repository being reviewed. The files I need to review are in the actual repository that's running this action. 
Let me check if there's a way to access the repository files. Let me try looking for the files without the base-action prefix."
      }
```

And failing. Needs bash now.

yamitzky:
@berenddeboer I ran into a similar issue and managed to get it working on my end. 

```yaml
      - name: Run Claude PR Action
        uses: anthropics/claude-code-action@beta
        with:
          settings: |
            {
              "permissions": {
                "allow": ["Bash(git:*)"],
                "additionalDirectories": ["/home/runner/_work"]
              }
            }
```

Hope this helps you out.

berenddeboer:
A thanks!! However the directory for me is "/home/runner/work" without the underscore so far. Just retrying this.

berenddeboer:
Ah seeing a bit more, this is one of the first lines:

```
Prompt file size: 13529 bytes
Running Claude with prompt from file: /home/runner/work/_temp/claude-prompts/claude-prompt.txt
Path /home/runner/_work was not found.
```

[...]

PR #312, opened 2025-07-19
fix: run Claude from workflow directory instead of base-action directory

ashwin-ant:
Changed the action to cd back to the original directory after installing dependencies, ensuring Claude runs in the context of the user's workflow rather than the base-action subdirectory.

fixes https://github.com/anthropics/claude-code-action/issues/303


wolffiex (approved):

arunjish:
Thanks

PR #318, opened 2025-07-21
add typescript types for claude code allowedTools and disallowedTools

btoo:


ashwin-ant:
This isn't correct. Allowed and disallowed tools can be any tool name, not just built-in tools in Claude Code.

PR #333, opened 2025-07-23
Add mode support

km-anthropic:
Add mode support + default to 'tag' mode. Structure support adding addtional modes as well.


src/create-prompt/index.ts
```diff
@@ -788,31 +789,36 @@ f. If you are unable to complete certain steps, such as running a linter or test
...
+    // Generate the prompt directly
     const promptContent = generatePrompt(
       preparedContext,
       githubData,
       context.inputs.useCommitSigning,
     );
 
+    await mkdir(`${process.env.RUNNER_TEMP}/claude-prompts`, {
+      recursive: true,
+    });
```
ashwin-ant: Why was this moved below?

src/entrypoints/prepare.ts
```diff
@@ -67,7 +75,7 @@ async function run() {
     const branchInfo = await setupBranch(octokit, githubData, context);
 
     // Step 9: Configure git authentication if not using commit signing
-    if (!context.inputs.useCommitSigning) {
+    if (!context.inputs.useCommitSigning && commentData?.user) {
```
ashwin-ant: I think we should still configure git auth. Maybe we can pass in the user as null for now and handle it in the function? Otherwise we won't be able to commit and push, I think.

src/modes/types.ts
```diff
@@ -0,0 +1,56 @@
+import type { ParsedGitHubContext } from "../github/context";
+import type { ModeName } from "./registry";
+
+export interface ModeContext {
```
ashwin-ant: nit: prefer `type` over `interface`,  here and elsewhere

src/modes/types.ts
```diff
@@ -0,0 +1,56 @@
...
+  /**
+   * Returns additional tools that should be allowed for this mode
+   * (base GitHub tools are always included)
+   */
+  getAllowedTools(): string[];
+
+  /**
+   * Returns tools that should be disallowed for this mode
+   */
+  getDisallowedTools(): string[];
```
ashwin-ant, line 50: Not sure I totally get why these are functions? Aren't they always going to be the same regardless of mode?

ashwin-ant (changes requested):

src/modes/types.ts
km-anthropic: Well my vision was for different modes to have different allowed / disallowed tools as needed. e.g. the agent / freeform tool might have bash tools as as default, whereas the review mode might only have github tools (no MCPs), etc.

This can be overridden in `claude.yml` though.

ashwin-ant: Hmm, gotcha. Yeah I think that makes sense.

src/entrypoints/format-turns.ts
```diff
@@ -3,21 +3,21 @@
 import { readFileSync, existsSync } from "fs";
 import { exit } from "process";
 
-export interface ToolUse {
+export type ToolUse = {
```
ashwin-ant, line 6: nice catch 😅

src/modes/registry.ts
```diff
@@ -0,0 +1,79 @@
...
+ * This function is called lazily to avoid circular dependencies.
+ * @internal
+ */
+function initializeRegistry(): void {
+  if (!initialized) {
+    const { tagMode } = require("./tag/index");
+    modeRegistry.set("tag", tagMode);
+    initialized = true;
+  }
+}
```
ashwin-ant: Can we simplify a bit? Since modes are all just statically defined, it might be simpler to just let callers reference constant objects rather than calling a function on a registry.

src/create-prompt/index.ts
km-anthropic: ah that's a miss - fixing it now

ashwin-ant (approved):
🔥

PR #350, opened 2025-07-28
ci: bump Claude Code version in action.yml files

### LINKED ISSUE #349: The bump-claude-code-version.yml workflow is updating the wrong action.yml file
**Describe the bug**

The [bump-claude-code-version.yml](https://github.com/anthropics/claude-code-action/blob/7c5a98d59d2464ae73f453c3e649f5b813374481/.github/workflows/bump-claude-code-version.yml)  workflow is only updating base-action/action.yml but not the root action.yml file. Both files should be updated to maintain consistency.

**To Reproduce**

1. Go to '.github/workflows/bump-claude-code-version.yml'
2. Check the workflow configuration
3. Notice it only targets 'base-action/action.yml'
4. See that version bumps don't affect the root action.yml file
5. Observe version inconsistency between the two action.yml files

**Expected behavior**
The bump version workflow should update both the root `action.yml` file AND the `base-action/action.yml` file to ensure version consistency across the entire action.

**Screenshots**

**Workflow yml file**
If it's not sensitive, consider including a paste of your full Claude workflow.yml file.

**API Provider**

[x] AWS Bedrock
[x] GCP Vertex

**Additional context**
The current bump version workflow is incomplete because it only updates one of the two action.yml files, which means:

Version bumps create inconsistency between root action.yml and base-action/action.yml
The root action.yml file retains outdated version references
This could lead to users getting different claude-code versions depending on which action.yml file is being referenced
Both files should maintain the same claude-code version for consistency

The workflow should be modified to update BOTH the root action.yml file AND the base-action/action.yml file to ensure that version bumps are properly reflected across all action files.

amondnet:
- Updated Claude Code version in base-action/action.yml
- Updated Claude Code version in root action.yml

fix #349

### DIFF
--- a/.github/workflows/bump-claude-code-version.yml
+++ b/.github/workflows/bump-claude-code-version.yml
@@ -75,15 +75,15 @@ jobs:
           echo "Updating base-action/action.yml..."
           ACTION_CONTENT=$(gh api repos/${GITHUB_REPOSITORY}/contents/base-action/action.yml?ref=$DEFAULT_BRANCH --jq '.content' | base64 -d)
           UPDATED_CONTENT=$(echo "$ACTION_CONTENT" | sed -E "s/(npm install -g @anthropic-ai\/claude-code@)[0-9]+\.[0-9]+\.[0-9]+/\1${{ env.NEW_VERSION }}/")
-          
+
           if ! echo "$UPDATED_CONTENT" | grep -q "@anthropic-ai/claude-code@${{ env.NEW_VERSION }}"; then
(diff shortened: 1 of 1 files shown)


ashwin-ant (dismissed):
Thanks! We're actually not using this workflow at the moment but we'll probably switch to it soon.

ashwin-ant:
@amondnet mind fixing the Prettier errors? `bun run format`

amondnet:
@ashwin-ant done

PR #353, opened 2025-07-28
add schedule & workflow dispatch paths. Also make prepare logic conditional

km-anthropic:
This PR enables the `workflow_dispatch` and `schedule` functionality for agent mode. Here's an issue where it's tested out: https://github.com/km-anthropic/claude-code-action/actions/runs/16577715694


src/create-prompt/index.ts
ashwin-ant, line 1: If I'm understanding this correctly, we shouldn't need any changes to `createPrompt` in this PR, right? Or is this in the case that someone's using `workflow_dispatch` and `schedule` with tag mode? Should we just throw an error since that combination doesn't make sense?

km-anthropic: Honestly we should just throw an error, I updated the code to do that.

ashwin-ant: Can we undo the changes to this file then? Just to keep this PR small. Or are they needed?

src/create-prompt/types.ts
```diff
@@ -1,6 +1,6 @@
 export type CommonFields = {
   repository: string;
-  claudeCommentId: string;
+  claudeCommentId?: string;
```
ashwin-ant: Ideally we can bring this back since it's only for tag mode where we'd assume there is a claude comment?

ashwin-ant (changes requested):

ashwin-ant (approved):
Let's get this is in and iterate!

PR #360, opened 2025-07-29
refactor: implement discriminated unions for GitHub contexts

km-anthropic:
## Summary
- Implements discriminated union types to eliminate ~20 null checks throughout the codebase
- Splits context types for better type safety between entity and automation events
- Ensures entity-specific code paths always have required fields

### New Type Structure
- **`ParsedGitHubContext`**: For entity events (issues/PRs) with required `entityNumber` and `isPR`
- **`AutomationContext`**: For workflow_dispatch/schedule events without entity fields  
- **`GitHubContext`**: Union type for all contexts
- Added type guards: `isEntityContext()` and `isAutomationContext()`

### Key Benefits
1. **Type Safety**: Entity-specific functions are guaranteed to have `entityNumber` and `isPR`
2. **No Null Checks**: Removed ~20 null checks from entity-specific code paths
3. **Clear Separation**: Agent mode explicitly handles automation events, tag mode handles entities
4. **Better DX**: TypeScript knows exactly which fields are available for each event type

### Example Usage
```typescript
// Mode interfaces accept union type
shouldTrigger(context: GitHubContext): boolean

// Entity-specific functions get guaranteed fields
createInitialComment(octokit: Octokit, context: ParsedGitHubContext)
// No need to check if entityNumber exists - it's guaranteed\!
```

## Test Plan
- [x] All existing tests pass
- [x] TypeScript compilation succeeds
- [x] Formatting checks pass


ashwin-ant:
This looks awesome! I think some of Claude's comments about redundant checks might be right, can we simplify those a bit?

src/github/context.ts
```diff
@@ -265,21 +276,12 @@ export function isIssuesAssignedEvent(
...
   context: GitHubContext,
 ): context is ParsedGitHubContext {
-  return (
-    context.eventName === "issues" ||
-    context.eventName === "issue_comment" ||
-    context.eventName === "pull_request" ||
-    context.eventName === "pull_request_review" ||
-    context.eventName === "pull_request_review_comment"
-  );
+  return ENTITY_EVENT_NAMES.includes(context.eventName as any);
```
ashwin-ant: nit: let's change the arrays to not be `as const` so that they stay as `string[]`, then we won't need the `any` here

src/github/context.ts
```diff
@@ -265,21 +276,12 @@ export function isIssuesAssignedEvent(
...
 
 // Type guard to check if context is an automation context
 export function isAutomationContext(
   context: GitHubContext,
 ): context is AutomationContext {
-  return (
-    context.eventName === "workflow_dispatch" ||
-    context.eventName === "schedule"
-  );
+  return AUTOMATION_EVENT_NAMES.includes(context.eventName as any);
```
ashwin-ant: here too

ashwin-ant (changes requested):

src/github/context.ts
km-anthropic: as const let's us then interpret it as a tuple of literal strings (so we can check if the given event is an entity or automation event). If we set it as a string[] then we'd also have to manually define a separate type union like type EntityEventName = "issues" | "issue_comment" | ... and keep both in sync. Instead I removed the any and added the type of event it was!

src/github/context.ts
km-anthropic: same as above

ashwin-ant (approved):

PR #343, opened 2025-07-25
feat: add GITHUB_HOST to github-mcp-server for GitHub Enterprise Server

uc4w6c:
I'd like to use `github-mcp-server` with GitHub Enterprise Server.
To support that, I added support for the `GITHUB_HOST` environment variable when running the Docker container.

I've tested the changes using GitHub.com (Cloud).

This change is intended to support the setup described here:
https://github.com/github/github-mcp-server?tab=readme-ov-file#github-enterprise-server-and-enterprise-cloud-with-data-residency-ghecom

PR #361, opened 2025-07-29
fix: move env var before image name in docker run for github-mcp-server

uc4w6c:
Fixes #343.

When I rebased the previous change (https://github.com/anthropics/claude-code-action/pull/343#issuecomment-3130194832), I mistakenly placed the GITHUB_HOST variable *after* the image name in the Docker run command, which caused a runtime error.

This PR fixes the variable placement so that it is correctly passed into the container.

Here are the test results from the GitHub Actions runs:

スクリーンショット 2025-07-30 7 52 06

スクリーンショット 2025-07-30 7 52 47

スクリーンショット 2025-07-30 7 52 59


ashwin-ant (approved):
Thanks for fixing this!

PR #365, opened 2025-07-30
feat: Skip initial comment creation when override_prompt is set

yukukotani:
### Summary

This PR improves the behavior when `override_prompt` is set by making the initial Claude comment creation conditional. When using `override_prompt`, the standard prompt instructions that enforce comment updates using `mcp__github_comment__update_claude_comment` are not included, so creating an initial comment becomes unnecessary.

### Changes

- **Conditional comment creation**: Skip initial comment creation when `override_prompt` is set
- **Optional comment ID handling**: Update tag mode to handle cases where comment ID may be null
- **Conditional MCP comment server**: Only include the GitHub comment MCP server when a Claude comment ID is provided
- **Improved error message**: Clarify that comment ID is required for tag mode "unless you're using override_prompt"
- **Git config refactoring**: Update `configureGitAuth` to fetch user data directly via Octokit instead of passing it as a parameter

### Test Updates

- Add test coverage for the conditional inclusion of the comment server based on whether `claudeCommentId` is provided
- Update existing tests to provide `claudeCommentId` where needed

### Benefits

1. **Cleaner execution**: Avoids creating unnecessary comments when using `override_prompt`
2. **Better resource usage**: Only includes MCP servers that are actually needed
3. **Improved flexibility**: Allows more streamlined workflows when using custom prompts

### Breaking Changes

None - this change is backward compatible and only affects behavior when `override_prompt` is explicitly set.

### DIFF
--- a/src/mcp/install-mcp-server.ts
+++ b/src/mcp/install-mcp-server.ts
@@ -72,22 +72,24 @@ export async function prepareMcpConfig(
       mcpServers: {},
     };
 
-    // Always include comment server for updating Claude comments
-    baseMcpConfig.mcpServers.github_comment = {
-      command: "bun",
-      args: [
-        "run",
-        `${process.env.GITHUB_ACTION_PATH}/src/mcp/github-comment-server.ts`,
-      ],
-      env: {
-        GITHUB_TOKEN: githubToken,
-        REPO_OWNER: owner,
-        REPO_NAME: repo,
-        ...(claudeCommentId && { CLAUDE_COMMENT_ID: claudeCommentId }),
-        GITHUB_EVENT_NAME: process.env.GITHUB_EVENT_NAME || "",
-        GITHUB_API_URL: GITHUB_API_URL,
-      },
-    };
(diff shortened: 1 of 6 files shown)


src/mcp/install-mcp-server.ts
```diff
@@ -72,22 +72,24 @@ export async function prepareMcpConfig(
...
+        env: {
+          GITHUB_TOKEN: githubToken,
+          REPO_OWNER: owner,
+          REPO_NAME: repo,
+          CLAUDE_COMMENT_ID: claudeCommentId,
+          GITHUB_EVENT_NAME: process.env.GITHUB_EVENT_NAME || "",
+          GITHUB_API_URL: GITHUB_API_URL,
+        },
+      };
+    }
```
yukukotani, line 92: github-comment-server won't work without `CLAUDE_COMMENT_ID`, so disabling it if it's empty.

yukukotani:
@km-anthropic I think this might go against your concept of Execution Modes. How do you think?

yukukotani:
let's wait v1

PR #381, opened 2025-08-02
feat: ship slash commands with GitHub Action

km-anthropic:
## Summary
- Adds infrastructure to bundle and install slash commands with the GitHub Action
- Includes a `/code-review` slash command for automated PR reviews
- Commands are automatically copied to `~/.claude/` during action setup

## Changes
1. Created `/slash-commands/` directory to store bundled commands
2. Added `/code-review` command with multi-agent review logic
3. Modified `setup-claude-code-settings.ts` to copy slash commands after settings setup
4. Added test coverage for the installation process

## Benefits
- Users automatically get curated slash commands without manual installation
- We control and can update the commands that ship with our action
- Simple implementation - just copy files during setup
- Enables workflows to use `override_prompt: '/code-review ...'`

## Next Steps
- [ ] Add configuration option to disable slash command installation
- [ ] Consider prefix to avoid conflicts (e.g., `/gh-code-review`)
- [ ] Add more slash commands for common GitHub workflows


base-action/src/setup-claude-code-settings.ts
```diff
@@ -65,4 +67,18 @@ export async function setupClaudeCodeSettings(
 
   await $`echo ${JSON.stringify(settings, null, 2)} > ${settingsPath}`.quiet();
   console.log(`Settings saved successfully`);
+
+  const __filename = fileURLToPath(import.meta.url);
+  const __dirname = dirname(__filename);
+  const slashCommandsDir = join(__dirname, "..", "..", "slash-commands");
```
ashwin-ant: Technically this will break in the base-action repo (https://github.com/anthropics/claude-code-base-action/). I think what I'd actually do here is add a slash commands dir to the `run-claude` script (and also as an input to the base-action/action.yml file.

km-anthropic: Fixed by adding slash_commands_dir as an optional input to base-action/action.yml. Only copies if directory is provided, so standalone usage won't break.

base-action/action.yml
```diff
@@ -61,6 +61,9 @@ inputs:
     description: "Timeout in minutes for Claude Code execution"
     required: false
     default: "10"
+  slash_commands_dir:
```
ashwin-ant: `experimental` for now?

km-anthropic: ah yeah nice catch

ashwin-ant (approved):

PR #384, opened 2025-08-03
feat: merge prepare and run-claude steps into unified entrypoint

ashwin-ant:
- Create run-claude-core.ts as the central Claude execution function
- Update action.yml to use single "Run Claude" step
- Create src/entrypoints/run.ts that combines prepare + run-claude logic
- Pass environment variables as object to Claude instead of setting process.env
- Update create-prompt to return tools (still exports env vars for compatibility)
- Add TODO comments for future refactoring to remove env var dependency
- Both action.yml and base-action continue to work correctly

This simplifies the action flow by removing intermediate steps and environment variable passing between prepare and run-claude phases.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -247,7 +247,7 @@ runs:
       shell: bash
       run: |
         # Try to format the turns, but if it fails, dump the raw JSON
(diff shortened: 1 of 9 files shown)


ashwin-ant:
@claude review this again. look specifically for edge cases that may have been missed in the migration. don't focus on minor things that can be fixed easily later

ashwin-ant:
@claude review again. concisely list things that may be missed in this refactor. don't bring up issues that existed previously.

PR #392, opened 2025-08-04
Fix/trigger validation redundant conditional

uc4w6c:
This is a refactor with no functional changes.

### DIFF
--- a/src/github/validation/trigger.ts
+++ b/src/github/validation/trigger.ts
@@ -1,6 +1,5 @@
 #!/usr/bin/env bun
 
-import * as core from "@actions/core";
 import {
   isIssuesEvent,
   isIssuesAssignedEvent,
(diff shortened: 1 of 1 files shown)


src/github/validation/trigger.ts
```diff
@@ -140,9 +137,3 @@ export function checkContainsTrigger(context: ParsedGitHubContext): boolean {
 export function escapeRegExp(string: string) {
   return string.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
 }
-
-export async function checkTriggerAction(context: ParsedGitHubContext) {
```
uc4w6c, line 144: This is no longer called after [this change](https://github.com/anthropics/claude-code-action/pull/333/files#diff-984420baa74ce018c448c860b3773bef370e894fc6e1cf1c73e3281f9b1e02d5L43).

PR #410, opened 2025-08-05
fix: remove git config user.name and user.email from allowed tools

ashwin-ant:
These git config commands are no longer needed as allowed tools since Claude should not be modifying git configuration settings. Updated the corresponding test to reflect this intentional change.


km-anthropic (approved):

noamta-gloat:
@ashwin-ant with this change, what would be the user and email used by the action's commits? Do we have to set them before calling the action?

Issue #207, opened 2025-06-26
Reuse existing Claude Code comment instead of creating a new comment when commits are pushed to PRs


shuo-alma:
Currently, when we configure claude code action using the auto review example, a new comment will be created by the claude code action when new commits are pushed to a PR. We will see many comments made by Claude Code instead of always updating the same comment when we push changes to a PR. It would be super helpful to add an option of always updating the same comment instead of creating new ones.

michaeljherrmann:
I was just coming here to create the same request! 

What I think is happening is that claude will always create an [initial tracking comment](https://github.com/anthropics/claude-code-action/blob/1e006bf2d00ba53d35d0abf2196a09249788b1cc/src/entrypoints/prepare.ts#L54). Then for a review, it creates a github review which will be posted as a new comment. 

I found that the tracking comment will then later get updated and repeat a lot of information from the review as well.  As a workaround I added:
```
disallowed_tools: "mcp__github_file_ops__update_claude_comment"
```
that way claude isn't allowed to update the tracking comment, but what would be ideal for me is that it wouldn't post it in the first place or perhaps so that it deletes it after. I can already see the status of the github action at the bottom of the PR page and can click there for more info if I need.

One idea is to expose this as an [input option](https://github.com/anthropics/claude-code-action?tab=readme-ov-file#inputs), so it's configurable?

phclark:
I had initially thought this is what the `use_sticky_comment` parameter is for, but notice that enabling the parameter doesn't effect behavior. I don't actually see where this parameter (which gets passed as an env variable to the entrypoint scripts) gets used directly. 
I agree though, this is something that is desperately needed, as comments tend to be very long and quickly clutter up a PR.

ashwin-ant:
This is indeed what `use_sticky_comment` is for. Are you not seeing that behavior?

ashwin-ant:
Ah it looks like there was a bug in the previous PR (#211). I have a fix out here: https://github.com/anthropics/claude-code-action/pull/226

phclark:
No, I'm not seeing it. Looking through the [PR](https://github.com/anthropics/claude-code-action/pull/211/files#diff-d2e0fde23f62c841678459c5cd0a4a2de38a542f35481cd1b7394bb20af869c7R66) that added the parameter a few days ago, I think I see the issue:


```
useStickyComment: process.env.STICKY_COMMENT === "true",
```

should be:
```
[...]

PR #411, opened 2025-08-05
Adding optional `sticky_comment_app_bot_id` and `sticky_comment_app_bot_name` options

piotrb:
This is to address the issues in #207 around this being quite hardcoded to Claude's own github app, adding two options to override the default behavior and have more flexible sticky comment behaviour.

`sticky_comment_app_bot_id` can just be set to your custom GH App id
`sticky_comment_app_bot_name` if for some reason you can't just use the ID, you can also override the name being matched, this allows you to match for any bot user with its name containing this string

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -110,6 +110,14 @@ inputs:
     description: "Use just one comment to deliver issue/PR comments"
     required: false
     default: "false"
+  sticky_comment_app_bot_id:
+    description: "The ID of the app bot that will be used to create the sticky comment. Defaults to the Claude app bot ID."
+    required: false
+    default: "209825114"
+  sticky_comment_app_bot_name:
+    description: "The name of the app bot that will be used to create the sticky comment. Defaults to 'claude'."
+    required: false
+    default: "claude"
   use_commit_signing:
     description: "Enable commit signing using GitHub's commit signature verification. When false, Claude uses standard git commands"
     required: false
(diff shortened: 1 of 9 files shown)


km-anthropic:
@ashwin-ant this looks reasonable to me. Do you see any security concerns of users potentially masking Claude's bot ID with another?

aulneau:
could i make one slight request for an addition? it'd be great if we could include a string value to match against, that is injected as a hidden html comment within the comment. this would be helpful if we used sticky comments in the experimental review mode which uses the default github token.

piotrb:
> could i make one slight request for an addition? it'd be great if we could include a string value to match against, that is injected as a hidden html comment within the comment. this would be helpful if we used sticky comments in the experimental review mode which uses the default github token.

That seems like a great potential enhancement, not sure if it should be scope creeping this PR though :)

This was a gentle un-hardcoding of a few constants .. adding new logic would probably require adding some tests or something .. which I'm not really up for contributing right now :)

aulneau:
> > could i make one slight request for an addition? it'd be great if we could include a string value to match against, that is injected as a hidden html comment within the comment. this would be helpful if we used sticky comments in the experimental review mode which uses the default github token.
> 
> That seems like a great potential enhancement, not sure if it should be scope creeping this PR though :)
> 
> This was a gentle un-hardcoding of a few constants .. adding new logic would probably require adding some tests or something .. which I'm not really up for contributing right now :)

Understandable! I'll probably make a pr for this!

piotrb:
> @ashwin-ant this looks reasonable to me. Do you see any security concerns of users potentially masking Claude's bot ID with another?

As far as I could tell, this bot user ID is ONLY ever used to create the initial comment, and in there its only used to determine if an existing comment came from the bot, so there isn't really anything its guarding from some kind of masking ..

piotrb:
Rebased with current main since there was conflicts

piotrb:
Hi @km-anthropic, would it be possible to have this change merged?

ranyhb:
@km-anthropic do you know when this feature will be merged ?

ranyhb:
@ashwin-ant 
can we add this implement ? 
small change that will add the enhancement of sticky-comment when using a user rather than claude-code..

FreCap:
Updated the PR so it has no conflicts anymore: https://github.com/anthropics/claude-code-action/pull/735

also added the following flag:

| `sticky_comment_matching_strategy` | Strategy for matching existing sticky comments: 'id_or_name' (match by ID OR name OR body), 'id_and_name' (match by ID AND name, allowing multiple comments per bot by changing name) | No       | `id_or_name`  |

PR #438, opened 2025-08-11
refactor: remove redundant step comments

uc4w6c:
There were multiple STEP 2 comments in the prepare script.
Since the `STEP X` comments are unnecessary, they have been removed entirely.

https://github.com/anthropics/claude-code-action/blob/main/src/entrypoints/prepare.ts#L28

https://github.com/anthropics/claude-code-action/blob/main/src/entrypoints/prepare.ts#L46

No functional changes.

### DIFF
--- a/src/entrypoints/prepare.ts
+++ b/src/entrypoints/prepare.ts
@@ -28,7 +28,7 @@ async function run() {
     const githubToken = await setupGitHubToken();
     const octokit = createOctokit(githubToken);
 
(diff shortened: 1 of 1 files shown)


uc4w6c:
In version 1.0.0, prepare.ts was changed, so I’ve merged those changes into this PR.
The issue of multiple STEP 2 comments was resolved, but there was still a problem where STEP 1 and 2 were missing and it started from STEP 3, so I’ve fixed that.

PR #439, opened 2025-08-11
Fix/remove unneeded tool in review mode

uc4w6c:
In review mode, the expected way to post comments is via `Bash(gh issue comment:*)` or `mcp__github_inline_comment__create_inline_comment`. 
However, `mcp__github_comment__update_claude_comment` is included in the available tools even when no Claude comment existed. 
https://github.com/anthropics/claude-code-action/blob/main/src/mcp/install-mcp-server.ts#L75

Updated `prepareMcpConfig` to include this tool only if `claudeCommentId` is set.

When `claudeCommentId` is not set, it causes an error on the MCP side, so this change has no impact.
https://github.com/anthropics/claude-code-action/blob/main/src/mcp/github-comment-server.ts#L41

### DIFF
--- a/test/install-mcp-server.test.ts
+++ b/test/install-mcp-server.test.ts
@@ -92,6 +92,7 @@ describe("prepareMcpConfig", () => {
       repo: "test-repo",
       branch: "test-branch",
       baseBranch: "main",
+      claudeCommentId: "1",
       allowedTools: [],
       context: mockContext,
     });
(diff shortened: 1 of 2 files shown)


test/install-mcp-server.test.ts
```diff
@@ -680,4 +720,85 @@ describe("prepareMcpConfig", () => {
 
     process.env.DEFAULT_WORKFLOW_TOKEN = oldTokenEnv;
   });
+
+  test("should include github_inline_comment server when mode is experimental-review and isPR is true", async () => {
```
uc4w6c, line 724: Added tests for the `github_inline_comment` server, which were previously missing.

art-shen:
Would it properly attach comments to the parent review with this?

uc4w6c:
@amal 
No. This fix only prevents the error, so it doesn’t add comments to the parent review.

uc4w6c:
Version 1.0.0 has been released and the review mode has been removed.
The new Agent mode resolves this issue, so I’m closing this PR.

PR #457, opened 2025-08-18
make github server url customizable when using github enterprise

YanyChoi:
- It was impossible to override environment variables `GITHUB_SERVER_API` and `GITHUB_API_URL` inside the prepare step.
- This made it impossible to use claude code with github enterprise.
- Make both environment variables inserted when calling the step, and use the default value when not declared.


ashwin-ant:
Can you share why this is needed? These should be included by default in the GHE workflow environment.

YanyChoi:
I thought so too, but the environment variables were only available on the job, but it didn't get inserted to the prepare step unless i inserted it through the step configuration's env settings.

PR #462, opened 2025-08-18
fix: provides github token for claude code action

ChrisJBurns:
Currently when running the `gh` command in the action, there is an error in the action logs that suggests that the `GH_TOKEN` isn't being set. We've solved this internally in our company by providing the `GH_TOKEN` in the action.


ashwin-ant (approved):
Thank you!

PR #492, opened 2025-08-27
feat: expose file download feature outside of tag mode

atsushi-ishibashi:
This PR implements the feature requested in #426 to make GitHub asset download functionality available outside of Tag Mode.

**Changes:**

- **New input parameter**: Added `download_github_assets` (default: `false`) to `action.yml`
- **Agent Mode enhancement**: Extended Agent Mode to download GitHub images when enabled and running in entity context (issues/PRs)
- **Environment variable**: Downloaded file paths are exposed via `CLAUDE_ASSET_FILES` as comma-separated values
- **Context-aware**: Only downloads assets for entity events (`issue_comment`, `pull_request`, etc.), not automation events
- **Error handling**: Failures don't break the entire action execution

**Usage:**

```yaml
  with:
    download_github_assets: true
    prompt: |
      Analyze images from this issue/PR.
      Files available in CLAUDE_ASSET_FILES environment variable.
```

**Files changed:**
- `action.yml` - Added new input parameter
- `src/modes/agent/index.ts` - Implemented download functionality
- `examples/agent-with-assets.yml` - Added usage example with shell expansion
- `test/modes/agent.test.ts` - Added comprehensive tests

This enables Agent Mode workflows to access and analyze GitHub attachments (screenshots, diagrams, etc.) that were previously only available in Tag Mode.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -27,6 +27,10 @@ inputs:
     description: "Comma-separated list of allowed bot usernames, or '*' to allow all bots. Empty string (default) allows no bots."
     required: false
     default: ""
+  download_github_assets:
+    description: "Download GitHub assets (images, attachments) for access via environment variables"
+    required: false
+    default: "false"
 
   # Claude Code configuration
   prompt:
@@ -143,6 +147,7 @@ runs:
         ADDITIONAL_PERMISSIONS: ${{ inputs.additional_permissions }}
         CLAUDE_ARGS: ${{ inputs.claude_args }}
         ALL_INPUTS: ${{ toJson(inputs) }}
+        DOWNLOAD_GITHUB_ASSETS: ${{ inputs.download_github_assets }}
 
     - name: Install Base Action Dependencies
       if: steps.prepare.outputs.contains_trigger == 'true'
--- a/action.yml
+++ b/action.yml
@@ -27,10 +27,6 @@ inputs:
     description: "Comma-separated list of allowed bot usernames, or '*' to allow all bots. Empty string (default) allows no bots."
     required: false
     default: ""
-  download_github_assets:
-    description: "Download GitHub assets (images, attachments) for access via environment variables"
-    required: false
-    default: "false"
 
   # Claude Code configuration
   prompt:
@@ -147,7 +143,6 @@ runs:
         ADDITIONAL_PERMISSIONS: ${{ inputs.additional_permissions }}
         CLAUDE_ARGS: ${{ inputs.claude_args }}
         ALL_INPUTS: ${{ toJson(inputs) }}
-        DOWNLOAD_GITHUB_ASSETS: ${{ inputs.download_github_assets }}
 
     - name: Install Base Action Dependencies
       if: steps.prepare.outputs.contains_trigger == 'true'
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -47,28 +47,27 @@ jobs:
 
 ## Inputs
 
-| Input                          | Description                                                                                                                                                         | Required | Default   |
-| ------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- | --------- |
-| `anthropic_api_key`            | Anthropic API key (required for direct API, not needed for Bedrock/Vertex)                                                                                          | No\*     | -         |
-| `claude_code_oauth_token`      | Claude Code OAuth token (alternative to anthropic_api_key)                                                                                                          | No\*     | -         |
-| `prompt`                       | Instructions for Claude. Can be a direct prompt or custom template for automation workflows                                                                         | No       | -         |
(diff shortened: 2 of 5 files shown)


km-anthropic:
## Found some issue in the example workflow file

### The Issue
The implementation sets `CLAUDE_ASSET_FILES` using `process.env` which only exists within the current Node.js process and **doesn't persist to subsequent GitHub Actions steps**. This makes the feature unusable as documented in the example workflow.

### How I Tested
Created a test workflow to verify environment variable persistence between steps:
1. Step 1: Runs the action with `download_github_assets: true`
2. Step 2: Checks if `CLAUDE_ASSET_FILES` is available
3. Result: **❌ Variable is empty in the next step**

**Test run:** https://github.com/km-anthropic/claude-code-action/actions/runs/17262284603

### Test Output
```
=== Checking CLAUDE_ASSET_FILES availability ===
CLAUDE_ASSET_FILES value: ''
❌ BUG CONFIRMED: CLAUDE_ASSET_FILES is empty!
The environment variable set by process.env doesn't persist between steps
```

### The Problem Code
In `src/modes/agent/index.ts` line 144:
```typescript
process.env.CLAUDE_ASSET_FILES = downloadedPaths.join(",");
```

This only sets the variable for the current process, not for subsequent workflow steps.

### The Fix
Replace line 144 with one of these options:

**Option 1: Use GitHub Actions core (recommended)**
```typescript
core.exportVariable('CLAUDE_ASSET_FILES', downloadedPaths.join(','));
```

**Option 2: Write to GITHUB_ENV file**
```typescript
const githubEnvFile = process.env.GITHUB_ENV;
if (githubEnvFile) {
  require('fs').appendFileSync(
    githubEnvFile,
    `CLAUDE_ASSET_FILES=${downloadedPaths.join(",")}\n`
  );
}
```

### Impact
Without this fix:
- The example workflow in `examples/agent-with-assets.yml` won't work
- Step 2 can't access the downloaded asset paths from Step 1
- The feature is effectively broken for multi-step workflows

The asset downloading itself works correctly - the issue is only with making the paths available to subsequent steps.

### Test Workflow Used
You can see the test workflow here: https://github.com/km-anthropic/claude-code-action/blob/main/.github/workflows/test-asset-env.yml

cc @atsushi-ishibashi - this needs to be fixed before merging, as the feature won't work as documented without this change.

atsushi-ishibashi:
@km-anthropic Fixed with Option1👍️

km-anthropic:
Thanks! Can you update usage.md too! then we should be good to go

atsushi-ishibashi:
@km-anthropic 
I wrote the example yml just to give you an idea, but should I leave it as is? Since its level of detail is different from the other examples, I was thinking of deleting it.

km-anthropic:
> @km-anthropic I wrote the example yml just to give you an idea, but should I leave it as is? Since its level of detail is different from the other examples, I was thinking of deleting it.

Yeah I think that's fine, ideally a simpler and cleaner example would be great if you can do it.

atsushi-ishibashi:
@km-anthropic 
Update usage.md and simplify the example GitHub Actions workflow.

km-anthropic (dismissed):

action.yml
```diff
@@ -27,6 +27,10 @@ inputs:
     description: "Comma-separated list of allowed bot usernames, or '*' to allow all bots. Empty string (default) allows no bots."
     required: false
     default: ""
+  download_github_assets:
+    description: "Download GitHub assets (images, attachments) for access via environment variables"
+    required: false
+    default: "false"
```
ashwin-ant: Does this need to be an input? Feels like we can do it by default without much issue.

atsushi-ishibashi: @ashwin-ant 
Since file download overhead could affect GitHub Actions runtime, I added an input to disable file downloads by default in agent mode. That said, I think it would also be fine to perform downloads by default. (If it seems possible to speed up the download process, I might try that in a separate PR.)

ashwin-ant: I'd prefer to just make it happen automatically given that most issues and PRs don't have images anyway. We can add an opt-out later if it becomes necessary.

atsushi-ishibashi: While writing usage.md, I also felt that it’s somewhat confusing that this input only works in agent mode, whereas in tag mode downloads occur regardless of its value. Moreover, since the mode is now auto-detected, this distinction seems even less clear.

atsushi-ishibashi: > I'd prefer to just make it happen automatically given that most issues and PRs don't have images anyway. We can add an opt-out later if it becomes necessary.

Sure👍️ I'll remove the input

atsushi-ishibashi: @ashwin-ant resolved!

km-anthropic (approved):
Generally looks good to me! Can you please update the docs, `usage.md` in specific?

atsushi-ishibashi:
> Generally looks good to me! Can you please update the docs, `usage.md` in specific?

@km-anthropic 
Are you suggesting that we should delete examples/agent-with-assets.yml and move its contents into usage.md?

xFlaviews:
Any update?

PR #80, opened 2025-05-28
Support ANTHROPIC_BASE_URL override from env

gbmarc1:
We would like to override the ANTHROPIC_BASE_URL when using the claude-code github action.
Related [PR](https://github.com/anthropics/claude-code-base-action/pull/20).

PR #561, opened 2025-09-10
feat: Add ANTHROPIC_CUSTOM_HEADERS environment variable support

jimutt:
Hi! 
We were happy to find that support for the `ANTHROPIC_BASE_URL` env var was added in these PRs: 
https://github.com/anthropics/claude-code-base-action/pull/20
https://github.com/anthropics/claude-code-action/pull/80

We're running an internal LLM gateway for additional control and hence need to set the base URL, however we are also dependent on the `ANTHROPIC_CUSTOM_HEADERS` env var that Claude Code supports: https://github.com/anthropics/claude-code/issues/321#issuecomment-2730081305

I am suggesting that this env var is also allowed to be passed through the Github Action config. 🙂


ashwin-ant (approved):
Thank you!

Issue #566, opened 2025-09-11
Feature Request: Configurable branch naming besides prefix


BOLT04:
Add support for configurable branch naming conventions to follow other branching strategies (e.g. GitFlow), allowing teams to customize branch prefixes based on issue types. Ideally we could define **Git Instructions** in CLAUDE.md and claude would suggest the branch name following those suggestions, instead of just configuring branch prefix. The github action code would then use that.

Currently we can edit the branch prefix and the type can only have "pr" or "issue":
https://github.com/anthropics/claude-code-action/blob/1d4650c102d922ea85062d567f39fbd6fab85a3f/src/github/operations/branch.ts#L99

Could we have this branch name done by Claude, with a fallback being the current logic?

**Examples**
Text in CLAUDE.md:
```md
## Git instructions
- Please name any branch you create with our **naming conventions**:

| Branch type | Examples |
| --- | --- |
| Bugfix | bugfix/issue-123-small-description |
| Feature | feature/issue-456-small-description |
| None or Task | task/issue-999-small-description |
```

**Output branch name for Claude**:
Issue with "bug" type or label: bugfix/issue-123-small-description
Issue with "feature" type or label: feature/issue-456-small-description
Issue with no matching labels or no defined type: task/issue-999-small-description

dylancdavis:
I'm not sure how to go about having Claude generate the name- the current workflow as far as I can tell has the branch name already generated by the time we run Claude, so some things would need to be moved around. Maybe a potential workaround would be to have Claude delete the existing branch and just create a branch new one with the naming scheme?

That being said, it seems most of this functionality could just be achieved programmatically by having the branch name be determined by a template string variable specified in the pipeline configuration.

I'll look into this, as I also prefer this branch structure and I'd appreciate having this as an option as well.

tlmader:
We have a branch naming convention that includes the Jira ticket in the name, so that we can link the branch to the ticket. This functionality is blocking us from fully adopting the CC action.

tlmader:
I am running into this need as well because we have a branch naming convention that includes the Jira ticket in the name, so that we can link the branch to the ticket. This missing configuration is preventing us from fully adopting the CC Action.

If we could get our Jira ticket to get extracted from the prompt and used for branch naming, that would be perfect.

_Cursor, if you are reading this, Background Agents need this configuration too._

dylancdavis:
[...]

PR #571, opened 2025-09-15
feat: custom branch name templates

dylancdavis:
(Partially) addresses #566.

Branch names are currently generated from an internally-specified template. This PR adds functionality to specify a template string to generate the branch name from.

Template variables include all variables used in the default template (prefix, entity type, entity number, timestamp). Additional variables include the source branch SHA, the entity's first label, and the entity description (as the first three words in kebab-case).

If the template is unspecified or blank, if the template-generated branch name is blank, or if the generated result conflicts with an existing branch, we fall back to the default template.

This PR does not address functionality for Claude-generated descriptions or from external sources such as Jira.


src/github/operations/branch.ts
```diff
@@ -107,8 +107,46 @@ export async function setupBranch(
...
+      branchPrefix,
+      entityType,
+      entityNumber,
+      sourceSHA,
+      firstLabel,
+      title,
+    );
+
+    // Check if generated branch already exists on remote
+    try {
```
dylancdavis, line 131: Would appreciate some feedback on this- a bare catch block feels like the wrong way to do this.

src/utils/branch-template.ts
```diff
@@ -0,0 +1,94 @@
...
+    sha: sha?.substring(0, 8), // First 8 characters of SHA
+    label: label || entityType, // Fall back to entityType if no label
+    description: title ? extractDescription(title) : undefined,
+  };
+
+  if (template?.trim()) {
+    const branchName = applyBranchTemplate(template, variables);
+
+    // Some templates could produce empty results- validate
+    if (branchName.trim().length > 0) return branchName;
```
dylancdavis, line 86: We don't enforce Kubernetes compatibility for the template case- I figured users that needed it could make their template compatible, and that this way we allow users to specify underscores or uppercase characters in their template if desired.

action.yml
```diff
@@ -23,6 +23,10 @@ inputs:
     description: "The prefix to use for Claude branches (defaults to 'claude/', use 'claude-' for dash format)"
     required: false
     default: "claude/"
+  branch_name_template:
+    description: "Template for branch naming. Available variables: {{prefix}}, {{entityType}}, {{entityNumber}}, {{timestamp}}, {{sha}}, {{label}}, {{description}}. {{label}} will be first label from the issue/PR, or {{entityType}} as a fallback. {{description}} will be the first 3 words of the issue/PR title in kebab-case. Default: '{{prefix}}{{entityType}}-{{entityNumber}}-{{timestamp}}'"
```
dylancdavis: The only new template variables here are label and description, which are admittedly the ones needed for my purposes. I'm avoiding Claude-generated descriptions or variables from external sources (e.g. Jira) for now from simplicity; I'd recommend a separate PR for additional template variables.

benkingcode: Why only first 3 words? Would be great to have the full title as the branch name eg `claude/571-add-custom-branch-name-templates`

BOLT04:
is there a way to get claude to review this PR @dylancdavis , wdyt?

action.yml
ashwin-ant: Seems like going up to 5 words could be okay?

ashwin-ant (dismissed):

ashwin-ant:
Really cool feature, thanks for submitting this! What do you think about making it the description 5 words?

action.yml
dylancdavis: No solid reason for 3 words. Just personal preference and trying to keep to shorter branch names. 5 words is fine with me

dylancdavis:
Updated to 5 words and made it an optional parameter in case we want more flexibility in the future

tlmader:
@ashwin-ant any chance we could get a follow-up review? We really need this functionality!

gurish165:
@ashwin-ant Can we get this reviewed?

ashwin-ant (approved):
Looks great!

gurish165:
Let's go!

PR #580, opened 2025-09-22
Add show_full_output option to control output verbosity

ddworken:
## Summary
- Adds a new `show_full_output` input option (defaults to `false`) to control output verbosity
- When disabled, shows only essential information like errors, completion status, and tool names
- Automatically enables full output when GitHub Actions debug mode is active

## Changes
- Added `show_full_output` input to action.yml
- Implemented output sanitization in run-claude.ts to filter non-essential details when disabled
- Always shows permission denials, errors, and final result messages regardless of setting
- Updated documentation to explain the new option and its behavior with debug mode

## Why
This change reduces log noise in GitHub Actions by default, making it easier to see important information. Full output can still be enabled for debugging when needed.

## Test plan
- [ ] Test with `show_full_output: false` (default) - verify minimal output
- [ ] Test with `show_full_output: true` - verify full JSON output
- [ ] Test with `ACTIONS_STEP_DEBUG=true` - verify debug mode enables full output
- [ ] Verify permission denials and errors are always visible
- [ ] Verify final result message is always shown

### DIFF
--- a/base-action/action.yml
+++ b/base-action/action.yml
@@ -55,6 +55,10 @@ inputs:
     description: "Optional path to a custom Bun executable. If provided, skips automatic Bun installation and uses this executable instead. WARNING: Using an incompatible version may cause problems if the action requires specific Bun features. This input is typically not needed unless you're debugging something specific or have unique needs in your environment."
     required: false
     default: ""
+  show_full_output:
+    description: "Show full JSON output from Claude Code. WARNING: This outputs ALL Claude messages including tool execution results which may contain secrets, API keys, or other sensitive information. These logs are publicly visible in GitHub Actions. Only enable for debugging in non-sensitive environments."
+    required: false
+    default: "false"
 
 outputs:
   conclusion:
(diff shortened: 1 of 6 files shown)


base-action/src/run-claude.ts
```diff
@@ -138,30 +192,49 @@ export async function runClaude(promptPath: string, options: ClaudeOptions) {
...
+  // Determine if full output should be shown
+  // Show full output if explicitly set to "true" OR if GitHub Actions debug mode is enabled
+  const isDebugMode = process.env.ACTIONS_STEP_DEBUG === "true";
+  let showFullOutput = options.showFullOutput === "true" || isDebugMode;
+
+  if (isDebugMode && options.showFullOutput !== "false") {
+    console.log("Debug mode detected - showing full output");
+    showFullOutput = true;
+  } else if (!showFullOutput) {
+    console.log("Running Claude Code (full output hidden for security)...");
```
ashwin-ant, line 204:
```suggestion
    console.log("Running Claude Code (full output hidden for security)...");
    console.log("Rerun in debug mode or enable `show_full_output: true` in your workflow file for full output.");
```

ashwin-ant (dismissed):

base-action/action.yml
```diff
@@ -55,6 +55,10 @@ inputs:
     description: "Optional path to a custom Bun executable. If provided, skips automatic Bun installation and uses this executable instead. WARNING: Using an incompatible version may cause problems if the action requires specific Bun features. This input is typically not needed unless you're debugging something specific or have unique needs in your environment."
     required: false
     default: ""
+  show_full_output:
```
ashwin-ant, line 58: Actually I think you want this in the `action.yml` at repo root since that's what we run from in claude-code-action. It may also be good to move the security note into SECURITY.md rather than README

ashwin-ant (dismissed):

base-action/action.yml
ddworken: Done!

ashwin-ant (approved):

PR #632, opened 2025-10-19
chore: upgrade actions/checkout from v4 to v5

DaleSeo:
I noticed that `actions/checkout` is outdated in the workflow files generated when setting up this action by running `/install-github-app` through Claude Code. 
This PR upgrades `actions/checkout` from v4 to v5 in all workflows, examples, and documentation, so users won't have to upgrade it manually.


ashwin-ant (approved):
Thank you!

PR #651, opened 2025-10-30
fix: The `use_sticky_comment`option does not reference the input `bot_id`.

huuyafwww:
Hello, thank you for developing such an excellent library.
I am currently using this action in my project.
I'm utilizing the `use_sticky_comment` option, but
a new comment is created with every commit.

Upon investigating the cause, I found that the bot user commenting in my project was not Claude.
Therefore, the `use_sticky_comment` option is not functioning properly.
When Claude runs, it first comments “Claude Code is working…”, right?
It reflects the execution results by updating this comment.
However, when a new commit is added, it cannot find an `existingComment` , so it creates a new comment again.

In my project, the GitHub Actions official user comments,
but the current code only considers comments made by the official Claude App user.
Therefore, I modified it so that if the botId in the actions input matches, it is also recognized as an `existingComment` .
This allows `existingComments` to be found even when users other than the official Claude App comment.

Once this change is reflected, my project will specify `botId` as `41898282` .
Below is the information for the GitHub Actions bot.
https://api.github.com/users/github-actions[bot]

### DIFF
--- a/src/github/operations/comments/create-initial.ts
+++ b/src/github/operations/comments/create-initial.ts
@@ -39,7 +39,7 @@ export async function createInitialComment(
         issue_number: context.entityNumber,
       });
       const existingComment = comments.data.find((comment) => {
-        const idMatch = comment.user?.id === CLAUDE_APP_BOT_ID;
+        const idMatch = comment.user?.id === CLAUDE_APP_BOT_ID || comment.user?.id === parseInt(context.inputs.botId);
         const botNameMatch =
           comment.user?.type === "Bot" &&
           comment.user?.login.toLowerCase().includes("claude");
--- a/src/github/operations/comments/create-initial.ts
+++ b/src/github/operations/comments/create-initial.ts
@@ -14,8 +14,6 @@ import {
 } from "../../context";
 import type { Octokit } from "@octokit/rest";
 
-const CLAUDE_APP_BOT_ID = 209825114;
-
 export async function createInitialComment(
   octokit: Octokit,
   context: ParsedGitHubContext,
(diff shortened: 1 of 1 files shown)


src/github/operations/comments/create-initial.ts
```diff
@@ -39,7 +39,7 @@ export async function createInitialComment(
         issue_number: context.entityNumber,
       });
       const existingComment = comments.data.find((comment) => {
-        const idMatch = comment.user?.id === CLAUDE_APP_BOT_ID;
+        const idMatch = comment.user?.id === CLAUDE_APP_BOT_ID || comment.user?.id === parseInt(context.inputs.botId);
```
ad-m-ss:
```suggestion
        const idMatch = comment.user?.id === parseInt(context.inputs.botId);
```

Why is there a need for a hardcoded constant if that value is also defined in https://github.com/huuyafwww/claude-code-action/blob/87cf21a4f276c5c71a16d388a981a6f16ba7af13/action.yml#L83 ? If there are no hardcoded constants, there won't be such mistakes, and there will only be one way for it to work.

ad-m-ss: BTW. That constant is duplicated: https://github.com/anthropics/claude-code-action/blob/037b85d0d2140e92af387c40ee8bbd9f0fc4d7a5/src/github/constants.ts#L8 , https://github.com/anthropics/claude-code-action/blob/037b85d0d2140e92af387c40ee8bbd9f0fc4d7a5/src/github/operations/comments/create-initial.ts#L17

ad-m-ss: There are two different numbers, so never mind.

huuyafwww: @ad-m-ss 
Thank you!
I've fixed it in the following commit.

https://github.com/anthropics/claude-code-action/pull/651/commits/1b026cfeae037dd07246e7856f117f5b4d2f0933

PR #656, opened 2025-11-04
fix: move /health endpoint check before 404 response

### LINKED ISSUE #10: File not found error causes entire action to fail
Hey team! Excited for the potential here. Quick bug report:

- https://github.com/gentrace/gentrace-node/pull/559#discussion_r2097554314
- https://github.com/gentrace/gentrace-node/actions/runs/15134621259/job/42543398832

## Description
When using the Claude Code Action, the workflow fails completely if it attempts to access a file that doesn't exist in the repository. This occurred in our repository when the action tried to access `tests/lib/test-dataset.test.ts` which wasn't present, resulting in a fatal Git error with exit code 128.

What's strange is that file most definitely exists. Not sure if I have to explicitly check out the repository in before invoking the Claude Code action? Doesn't seem to be the case in the examples thoug.h

## Current Behavior
- The action attempts to compute SHA for referenced files
- If a file doesn't exist, it produces a Git error: `fatal: could not open '[filename]' for reading: No such file or directory`
- The entire workflow fails with `ShellError: Failed with exit code 128`
- No partial results or analysis is provided for other valid files

## Expected Behavior
- Gracefully handle missing files without failing the entire workflow

nathanclevenger:
## Problem

The  endpoint was unreachable because it was checked AFTER the 404 response for non- URLs. This caused all requests to  to return 404.

This bug prevented the proxy from working correctly and caused automated Claude Code reviews to fail with 'API Error: 404 Not Found'.

## Root Cause

In `base-action/src/proxy-server.ts`, the code flow was:

1. Check if URL is NOT `/v1/messages` → return 404
2. Check if URL IS `/health` → return health stats (UNREACHABLE)

The health check could never be reached because it failed the first check.

## Solution

Move the `/health` endpoint check BEFORE the 404 response:

1. Check if URL IS `/health` → return health stats 
2. Check if URL is NOT `/v1/messages` → return 404

## Testing

- Manual test: Start proxy and curl  (should return stats JSON)
- Manual test: curl  with POST (should proxy to claude-lb)
- Manual test: curl  (should return 404)

## Impact

This fix resolves automated Claude Code reviews failing in the platform repository. Once merged and the SHA is updated in the platform workflows, automated reviews will start working again.

Closes #10

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -246,7 +246,7 @@ runs:
         VERTEX_REGION_CLAUDE_3_7_SONNET: ${{ env.VERTEX_REGION_CLAUDE_3_7_SONNET }}
 
     - name: Update comment with job link
-      if: steps.prepare.outputs.contains_trigger == 'true' && steps.prepare.outputs.claude_comment_id && always()
+      if: steps.prepare.outputs.contains_trigger == 'true' && steps.prepare.outputs.claude_comment_id && !cancelled()
       shell: bash
       run: |
         bun run ${GITHUB_ACTION_PATH}/src/entrypoints/update-comment-link.ts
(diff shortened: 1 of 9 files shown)


nathanclevenger:
Closing - this was created against the wrong repository (anthropics/claude-code-action instead of dot-do fork). The fix is already in our fork's main branch.

PR #658, opened 2025-11-04
fix: add idleTimeout and bedrock-only proxy mode

nathanclevenger:
## Problem

1. **Timeout Issue**: Default 10s Bun.serve() `idleTimeout` was causing `[Bun.serve]: request timed out after 10 seconds` errors during long Claude API requests
2. **Proxy Bypass Bug**: Proxy was not being used when only Bedrock credentials were available, bypassing Cloudflare AI Gateway monitoring

## Root Cause

The `getProxyMode()` function only returned `'bedrock-anthropic'`, `'anthropic-only'`, or `'direct'` modes. When only `AWS_BEARER_TOKEN_BEDROCK` was available (without `ANTHROPIC_API_KEY`), it returned `'direct'` mode, which caused `shouldUseProxy()` to return `false`, preventing `ANTHROPIC_BASE_URL` from being set.

## Solution

1. **Added `idleTimeout: 300`** (5 minutes) to prevent proxy timeout
2. **Added `'bedrock-only'` mode** to ensure proxy is used when Bedrock credentials are available (for AI Gateway monitoring)
3. **Updated logging** to reflect new bedrock-only mode

## Impact

- All requests will now route through Cloudflare AI Gateway (when credentials are available)
- Bedrock usage will be tracked in AI Gateway logs
- Long-running Claude requests won't timeout
- Better observability and monitoring across all credential configurations

## Testing

Test in GitHub Actions workflows to verify:
- Bedrock requests appear in AI Gateway logs
- No more 10-second timeouts
- Proxy correctly detects and uses available credentials

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -246,7 +246,7 @@ runs:
         VERTEX_REGION_CLAUDE_3_7_SONNET: ${{ env.VERTEX_REGION_CLAUDE_3_7_SONNET }}
 
     - name: Update comment with job link
-      if: steps.prepare.outputs.contains_trigger == 'true' && steps.prepare.outputs.claude_comment_id && always()
+      if: steps.prepare.outputs.contains_trigger == 'true' && steps.prepare.outputs.claude_comment_id && !cancelled()
       shell: bash
       run: |
         bun run ${GITHUB_ACTION_PATH}/src/entrypoints/update-comment-link.ts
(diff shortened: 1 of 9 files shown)


nathanclevenger:
Wrong repository - closing. PR belongs in dot-do/claude-code-action fork.

Issue #12, opened 2025-05-21
Claude GitHub Action requires both `CLOUD_ML_REGION` and model-specific region variables for Vertex AI


moesy:
## Summary

When using the Claude GitHub Action with `use_vertex: true`, the action fails unless **both** `CLOUD_ML_REGION` and the corresponding model-specific `VERTEX_REGION_CLAUDE_*` environment variable are set.

I assumed `CLOUD_ML_REGION` to act as a global default for all Vertex AI models, but that does not appear to be the case.

If `CLOUD_ML_REGION` is omitted, the action fails with:

```
API Error: No region was given. The client should be instantiated with the `region` option or the `CLOUD_ML_REGION` environment variable should be set.
```

If the model-specific variable (`VERTEX_REGION_CLAUDE_3_7_SONNET` in this case) is omitted the run also fails with:

```
API Error: No region was given. The client should be instantiated with the `region` option or the `CLOUD_ML_REGION` environment variable should be set.
```

## Expected Behavior

Setting `CLOUD_ML_REGION` should be sufficient to route all models through Vertex AI unless explicitly overridden by a model-specific variable.

## Request

- Support `CLOUD_ML_REGION` as a global default across all models
- If model-specific region variables are required, document the full list and precedence behavior

moesy:
This is no longer an issue with the latest versions of the action.

PR #659, opened 2025-11-04
fix: set idleTimeout to 255 (Bun max)

nathanclevenger:
## Problem
The proxy was failing with:
```
TypeError [ERR_INVALID_ARG_TYPE]: Bun.serve expects idleTimeout to be 255 or less
```

PR #12 set `idleTimeout: 300` (5 minutes), but Bun requires it to be ≤255 seconds.

## Solution
Changed `idleTimeout` from 300 to 255 seconds (4.25 minutes).

This is still sufficient for long-running Claude API requests while staying within Bun's limit.

## Testing
- Confirms idleTimeout is now within Bun's allowed range
- Maintains adequate timeout for Claude API requests

### DIFF
--- a/base-action/src/proxy-server.ts
+++ b/base-action/src/proxy-server.ts
@@ -128,7 +128,7 @@ export async function startProxyServer(): Promise<number> {
   const server = Bun.serve({
     port: PROXY_PORT,
     hostname: '127.0.0.1',
-    idleTimeout: 300, // 5 minutes for long-running Claude API requests
+    idleTimeout: 255, // Max allowed by Bun (4.25 minutes) for long-running Claude API requests
 
     async fetch(req: Request): Promise<Response> {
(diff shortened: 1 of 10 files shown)


base-action/src/proxy-server.ts
```diff
@@ -0,0 +1,217 @@
...
+}
+
+export async function startProxyServer(): Promise<number> {
+  // Validate credentials at startup - fail fast if misconfigured
+  validateCredentials();
+
+  const server = Bun.serve({
+    port: PROXY_PORT,
+    hostname: '127.0.0.1',
+    idleTimeout: 255, // Max allowed by Bun (4.25 minutes) for long-running Claude API requests
```
Jarred-Sumner, line 131: `idleTimeout: 0`

will make it unlimited

PR #667, opened 2025-11-07
Add preserve_checkout_credentials option

kusumotoa:
## Summary

Add `preserve_checkout_credentials` option to allow workflows to preserve git credentials from `actions/checkout` for subsequent git operations after Claude Code Action execution.

## Problem

When workflows need to perform git push operations after Claude Code Action runs, they encounter authentication failures. This is because Claude Code Action removes the `http.https://github.com/.extraheader` configuration set by `actions/checkout` and replaces it with its own authentication token.

Example workflow that fails without this option:
```yaml
- uses: actions/checkout@v4
  with:
    prompt: "Analyze changes..."
- name: Push changes
  run: git push  # ❌ Fails: authentication error
```

## Solution

Added `preserve_checkout_credentials` input option (default: `false` for backward compatibility):

- When `true`: Preserves the extraheader configuration from `actions/checkout`
- When `false`: Uses Claude Code Action's own authentication (existing behavior)

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -77,6 +77,10 @@ inputs:
     description: "Enable commit signing using GitHub's commit signature verification. When false, Claude uses standard git commands"
     required: false
     default: "false"
+  preserve_checkout_credentials:
+    description: "Preserve git credentials from actions/checkout for subsequent git operations. When true, Claude Code will not modify the git config extraheader set by actions/checkout. This is useful for workflows that need to perform git push operations after Claude Code execution."
+    required: false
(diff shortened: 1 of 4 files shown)


kusumotoa:
Closing to address critical issues found in code review:

1. Non-Signing Mode authentication: Need to update remote URL even in preserve mode
2. Token separation weakening: Need to add security warnings and documentation
3. Commit history: Squashing into single commit

Will reopen with fixes shortly.

PR #672, opened 2025-11-11
fix: prevent indefinite hang when OAuth token expires

ShahadIshraq:
## Problem

When OAuth tokens expire (e.g., subscription ends), the action hangs indefinitely at "Setting up Claude Code..." without failing or providing feedback to users.

## Root Cause

The `runClaude()` function in `base-action/src/run-claude.ts` has no timeout mechanism. When Claude Code CLI receives an expired OAuth token, it hangs during authentication without timing out or returning an error code.

## Solution

Add a configurable timeout mechanism with real-time authentication error detection:

- **Timeout Protection**: Default 15-minute timeout prevents indefinite hangs
- **Error Detection**: Monitors output for auth-related errors (401, 403, "expired", "unauthorized", etc.)
- **Clear Guidance**: Provides actionable error messages when auth failures are detected
- **Configurable**: New `claude_code_timeout_ms` input allows customization

## Changes

- Add timeout mechanism to `base-action/src/run-claude.ts`
- Extract helper functions for readability (`containsAuthenticationError`, `logAuthenticationErrorGuidance`, `createTimeoutPromise`)
- Add `claude_code_timeout_ms` input to both `action.yml` and `base-action/action.yml` (default: 900000ms = 15 min)
- Update error messages in `base-action/src/validate-env.ts` to explain timeout behavior
- Add comprehensive test suite for authentication error detection

## Testing

- ✅ Unit tests added for `containsAuthenticationError()` function
- ✅ Tests verify case-insensitive matching and false positive prevention
- ✅ All existing tests continue to pass

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -113,6 +113,10 @@ inputs:
     description: "Newline-separated list of Claude Code plugin marketplace Git URLs to install from (e.g., 'https://github.com/user/marketplace1.git\nhttps://github.com/user/marketplace2.git')"
     required: false
     default: ""
+  claude_code_timeout_ms:
+    description: "Maximum time in milliseconds to wait for Claude Code execution before timing out (default: 900000 = 15 minutes). Prevents indefinite hangs when authentication fails or tokens expire."
+    required: false
+    default: "900000"
 
 outputs:
   execution_file:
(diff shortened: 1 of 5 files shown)


ar03:
Reviewed this PR. The implementation is solid:

  - Timeout mechanism with Promise.race prevents indefinite hangs
  - Auth error detection provides actionable feedback to users
  - Graceful shutdown (SIGTERM then SIGKILL) handles stuck processes correctly
  - Unit tests cover the error detection logic well

  The configurable claude_code_timeout_ms input is a good design choice for users with longer-running workflows.

  This addresses a real pain point where expired OAuth tokens cause the action to hang forever at "Setting up Claude Code..." with no feedback. Ready for merge.

PR #706, opened 2025-11-28
fix: `commentBody` may be `null`

plaflamme:
This handles the cases where `pull_request_review` events have no comments (`commentBody` field is `null`). In those cases, the `null` value is converted to the empty string.

The issue was testing `!commentBody` which was triggerring on empty strings as well. This guard was removed (which is the fix), but for clarity, the `commentBody` field was also made optional to make it clear that the comment may be missing.

### DIFF
--- a/src/create-prompt/index.ts
+++ b/src/create-prompt/index.ts
@@ -192,11 +192,6 @@ export function prepareContext(
       if (!isPR) {
         throw new Error("IS_PR must be true for pull_request_review event");
       }
-      if (!commentBody) {
-        throw new Error(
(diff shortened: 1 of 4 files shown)


ashwin-ant (dismissed):

ashwin-ant:
@plaflamme mind running `bun run format`?

plaflamme:
@ashwin-ant done in https://github.com/anthropics/claude-code-action/pull/706/commits/7842718a91bc0e8a1cfa0d388381550f9e046687

ashwin-ant (approved):

PR #734, opened 2025-12-11
fix: Use frozen lockfile for reproducible builds

shortstacked:
This ensures the action installs using the lockfile so we get reproducible builds. Given the increase in supply chain attacks recently this is good practice to have.


shortstacked:
Bumping for visibility.

algomaster99:
Fully agreed.

Issue #764, opened 2025-12-23
Ability to run GitHub action on local machine


yhdesai:
Is there a way where I create the issue on GitHub, and it runs the Claude code on my personal machine, then pushes the changes back to GitHub

ModeEric:
This is already possible using [GitHub's self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners).

You would just connect this to your repo, and modify your workflow file to use `runs-on: self-hosted`

There is room to do this with a new feature from the CLI, but as far as I can tell, nothing currently exists

PR #775, opened 2025-12-28
feat(examples): add workflow_dispatch example for on-demand Claude tasks

ModeEric:
Adds a new example showing how to trigger Claude manually via GitHub UI or API without creating an issue. This addresses the use case described in #764  where users want to run Claude on-demand.

Features:
- Trigger via GitHub Actions UI with custom task input
- Trigger via GitHub API for external integrations
- Optional issue number input for additional context
- Includes curl example for API usage

### DIFF
--- a/examples/workflow-dispatch.yml
+++ b/examples/workflow-dispatch.yml
@@ -52,3 +52,6 @@ jobs:
             ${{ github.event.inputs.issue_number && format('Related Issue: #{0}', github.event.inputs.issue_number) || '' }}
 
             Task: ${{ github.event.inputs.task }}
+          # Basic file operations - add more tools as needed (e.g., Bash commands)
+          claude_args: |
+            --allowedTools "Read,Write,Edit,MultiEdit,LS"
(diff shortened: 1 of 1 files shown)


examples/workflow-dispatch.yml
```diff
@@ -0,0 +1,54 @@
...
+        uses: anthropics/claude-code-action@v1
+        with:
+          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          prompt: |
+            Repository: ${{ github.repository }}
+            Branch: ${{ github.ref_name }}
+            ${{ github.event.inputs.issue_number && format('Related Issue: #{0}', github.event.inputs.issue_number) || '' }}
+
+            Task: ${{ github.event.inputs.task }}
```
ashwin-ant, line 54: This is cool, should we also include some allowed tools to make this work out of the box a bit better?

ashwin-ant (dismissed):

examples/workflow-dispatch.yml
ModeEric: Yeah good idea! I ran into similar issues with testing, so I updated the PR to allow some basic file editing tools so it works out of the box.

PR #789, opened 2026-01-04
feat: isolate sticky comments by job ID to prevent workflow conflicts

### LINKED ISSUE #419: Sticky comment for review mode
Hi there!

We are currently using a [forked version](https://github.com/Uniswap/claude-code-action/tree/feat/sticky-comment) of the action to enable sticky comments in review mode. I'm happy to create a PR for this, but wanted to confirm/explore what it could look like. (we would prefer to not maintain a fork)

We want to continue to use the sticky comment in review mode for a more thorough analysis/overview of the PR, and we only want one comment because we use stacked diffs, and often rebase/force push to a given pr many times. If the bot were to use the comment functionality each time and put the whole thing in the body of the review, it will clog up the PR significantly.

the changes I've done at a high level:
- sticky comment works only with a custom override prompt
- make sure we persist content of comment when the action runs again (so the action can learn from what was already posted)
- it injects a (configurable) hidden comment so we can find it between runs

cc @ashwin-ant

TosinAF:
## Summary

When multiple workflows use `use_sticky_comment: true` on the same PR, they previously could co-opt each other's comments due to loose bot ID/name matching. This change uses `github.job` to automatically isolate each workflow's sticky comment.

**Problem:** PR #780 added `bot_name` for isolation, but the fallback matching logic still had issues where comments from other workflows could be co-opted.

**Solution:** Use the GitHub Actions job ID (`github.job`) as an automatic, unique identifier for each workflow's sticky comment.

### Changes

- Add hidden header `<!-- sticky-job: {jobId} -->` to sticky comments
- Match comments **strictly** by this job ID header instead of bot ID/name
- Each workflow job automatically gets its own isolated comment
- **Zero configuration needed** - works automatically

### How It Works

```yaml
# .github/workflows/claude-docs-review.yml
jobs:
  claude-docs-review:  # <-- This job ID becomes the sticky comment identifier
    steps:
      - uses: anthropics/claude-code-action@main
        with:
          use_sticky_comment: true
          # No bot_name needed! Uses job ID automatically

# .github/workflows/claude-security-review.yml
jobs:
  claude-security-review:  # <-- Different job ID = different comment
    steps:
      - uses: anthropics/claude-code-action@main
        with:
          use_sticky_comment: true
```

### Mental Model

- **One job = one sticky comment**
- Same job ID = same comment gets updated (intended)
- Different job IDs = separate comments (no conflicts)

## Test Plan

- [ ] Run `bun test` to verify new tests pass
- [ ] Run `bun run typecheck` to verify types
- [ ] Manual test: Create PR with two workflows using `use_sticky_comment: true` and verify they get separate comments
- [ ] Manual test: Re-run same workflow and verify it updates the same comment

Closes #419
Related to #780

### DIFF
--- a/src/github/operations/branch.ts
+++ b/src/github/operations/branch.ts
@@ -151,10 +151,7 @@ export async function setupBranch(
       // Handle open PR: Checkout the PR branch
       console.log("This is an open PR, checking out PR branch...");
 
-      const originalBranchName = prData.headRefName;
-      // Use unique local branch name to avoid conflicts with existing branches
-      // Format: pr-{number} ensures uniqueness since PR numbers are unique per repo
-      const localBranchName = `pr-${entityNumber}`;
+      const branchName = prData.headRefName;
 
       // Determine optimal fetch depth based on PR commit count, with a minimum of 20
       const commitCount = prData.commits.totalCount;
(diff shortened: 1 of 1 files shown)


TosinAF:
@claude can you rebase over latest main?

PR #847, opened 2026-01-20
Fix multiline claude_args parsing issue with spaces in Bash tools (#844)

### LINKED ISSUE #844: [BUG] Multiline claude_args with Bash tools containing spaces parsed incorrectly
## Bug Description

When using multiline YAML for `claude_args` with Bash tool patterns that contain spaces (like `Bash(git diff --name-only HEAD~1)`), the SDK incorrectly parses the `--allowedTools` argument by splitting on spaces.

## Reproduction

```yaml
claude_args: |
  --max-turns 15
  --allowedTools Bash(git diff --name-only HEAD~1),Bash(git diff HEAD~1),Read,Glob,Grep
```

## Expected Behavior

SDK should parse `allowedTools` as:
```json
["Bash(git diff --name-only HEAD~1)", "Bash(git diff HEAD~1)", "Read", "Glob", "Grep"]
```

## Actual Behavior

SDK parses `allowedTools` as:
```json
["Bash", "git", "diff"]
```

From the logs:
```
SDK options: {
  "allowedTools": [
    "Bash",
    "git",
    "diff"
  ],
  ...
  "extraArgs": {
    "max-turns": "15",
    "name-only": "HEAD~1"
  },
  ...
}
```

Note that `--name-only HEAD~1` is also incorrectly parsed as a separate extraArg.

## Evidence

Full error:
```
SDK execution error: error: Claude Code process exited with code 1
```

## Workaround

Put everything on a single line with quotes:
```yaml
claude_args: --max-turns 15 --allowedTools "Bash(git diff --name-only HEAD~1),Bash(git diff HEAD~1),Read,Glob,Grep"
```

## Environment

- Action version: `anthropics/claude-code-action@v1`
- Runner: `ubuntu-latest`
- Event: `pull_request`

## Notes

The [migration guide](https://github.com/anthropics/claude-code-action/blob/main/docs/migration-guide.md) shows multiline `claude_args` examples, but they don't include Bash tools with spaces:
```yaml
claude_args: |
  --max-turns 15
  --allowedTools Edit,Read,Write,Bash
```

The issue appears to be that `shell-quote` (or whatever parser is used) doesn't handle the multiline + spaces-in-arguments combination correctly.

jose-compu:
## Summary

Fixes issue #844 where multiline `claude_args` with Bash tools containing spaces were incorrectly parsed by the shell-quote parser, causing arguments to be split at spaces.

## Changes

1. **Added detection logic** to identify incorrectly parsed arguments (e.g., `Bash(git` without closing paren, or standalone command words after tool names)
2. **Added warning** when incorrect parsing is detected, guiding users to quote values with spaces
3. **Updated documentation** in migration guide to clarify quoting requirement for Bash tools with spaces
4. **Added tests** to verify the fix works correctly and warnings are triggered appropriately

## Testing

All existing tests pass, and new tests verify:
- Quoted values with spaces parse correctly
- Unquoted values without spaces work as before (no false warnings)
- Warnings are triggered when incorrect parsing is detected

## Solution

The fix detects common patterns of incorrect parsing and warns users. The proper solution (as mentioned in the issue) is to quote values with spaces, which now works correctly in both single-line and multiline YAML formats.

Fixes #844

### DIFF
--- a/base-action/package-lock.json
+++ b/base-action/package-lock.json
@@ -9,6 +9,7 @@
       "version": "1.0.0",
       "dependencies": {
         "@actions/core": "^1.10.1",
+        "@anthropic-ai/claude-agent-sdk": "^0.2.11",
         "shell-quote": "^1.8.3"
       },
       "devDependencies": {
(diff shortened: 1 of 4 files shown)


jose-compu:
please review @chrislloyd @dougbarth @bhuga if worth it.

ashwin-ant (approved):

base-action/package-lock.json
ashwin-ant, line 1: I think this change isn't needed, we shouldn't have a package lock file at all since we have bun.lock

PR #901, opened 2026-02-04
feat: add ability to provide GitHub.com token to setup-bun to avoid rate limits

peloyeje:
## Summary

Pass `github_dotcom_token` input (or fallback `github.token`) to the `setup-bun` action, which prevents 403 rate limit errors when fetching Bun release tags from GitHub API on shared runners

Follow-up of https://github.com/anthropics/claude-code-action/pull/861

## Problem

The `setup-bun` action fetches Bun release tags from the GitHub API. Without an authenticated token, it hits rate limits (403 error) when running from shared GitHub Actions runners:

```
Error: API rate limit exceeded for <IP>. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
```


ashwin-ant:
An alternative here might be to just pass a path to pass `path_to_bun_executable` instead. That way you can install Bun however you like. That might be cleaner than adding another input that's only used for one thing. 

https://github.com/anthropics/claude-code-action/blob/main/action.yml#L120C3-L123

peloyeje:
> An alternative here might be to just pass a path to pass `path_to_bun_executable` instead. That way you can install Bun however you like. That might be cleaner than adding another input that's only used for one thing.
> 
> [`main`/action.yml#L120C3-L123](https://github.com/anthropics/claude-code-action/blob/main/action.yml?rgh-link-date=2026-02-06T01%3A17%3A35Z#L120C3-L123)

Sounds reasonable indeed!
Just opened https://github.com/anthropics/claude-code-action/pull/913 to supersede this one

peloyeje:
Replaced by https://github.com/anthropics/claude-code-action/pull/913

PR #940, opened 2026-02-13
feat: isolate sticky comments by job ID to prevent workflow conflicts

TosinAF:
## Summary

- Isolate sticky comments per workflow job using `<!-- bot: {jobId} -->` HTML comment headers
- Each workflow job (e.g. `claude-code-review`, `claude-docs-review`) automatically gets its own isolated comment via `GITHUB_JOB`
- Preserve bot headers through the MCP comment server update flow (fetch current comment, extract header, re-prepend after sanitization)
- Use regex-based comment matching with `hasAnyHeader` check for proper multi-bot isolation and legacy fallback
- Paginate comment fetching to handle PRs with 30+ comments

### Key design decisions

- **`GITHUB_JOB` for auto-isolation**: No user configuration needed — each workflow job naturally gets a unique identifier
- **MCP server header preservation**: Instead of modifying `sanitizer.ts` (security code), the MCP server fetches the current comment to extract and re-prepend the bot header after sanitization
- **`<!-- bot: -->` format**: Aligns with #780's header format for future compatibility
- **`context.inputs.botId` for matching**: Uses the dynamic bot ID input instead of hardcoded `CLAUDE_APP_BOT_ID`

### Files changed

| File | Change |
|------|--------|
| `action.yml` | Pass `GITHUB_JOB` env var |
| `src/github/context.ts` | Add `jobId` to inputs |
| `src/github/operations/comments/common.ts` | `createCommentBody()` with bot header, `extractBotHeader()` |
| `src/github/operations/comments/create-initial.ts` | Regex matching, pagination, `hasAnyHeader` check |
| `src/github/operations/comments/update-with-branch.ts` | Pass bot identifier |
| `src/github/operations/comment-logic.ts` | Preserve bot header through `updateCommentBody()` |
| `src/mcp/github-comment-server.ts` | Fetch current comment, extract + re-prepend header |
| `src/github/utils/sanitizer.ts` | Reverted — no modifications to security code |

## Test plan

- [x] All 658 tests pass
- [x] TypeScript typecheck clean
- [x] Verified bot header format: `<!-- bot: {jobId} -->`
- [x] Verified header preservation through comment update lifecycle
- [x] Verified sanitizer strips bot headers (MCP server handles preservation separately)
- [ ] Test with two concurrent workflow jobs on same PR

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -263,6 +263,7 @@ runs:
         INCLUDE_COMMENTS_BY_ACTOR: ${{ inputs.include_comments_by_actor }}
         EXCLUDE_COMMENTS_BY_ACTOR: ${{ inputs.exclude_comments_by_actor }}
         GITHUB_RUN_ID: ${{ github.run_id }}
+        GITHUB_JOB: ${{ github.job }}
         USE_STICKY_COMMENT: ${{ inputs.use_sticky_comment }}
         CLASSIFY_INLINE_COMMENTS: ${{ inputs.classify_inline_comments }}
         DEFAULT_WORKFLOW_TOKEN: ${{ github.token }}
--- a/src/github/context.ts
+++ b/src/github/context.ts
@@ -92,6 +92,7 @@ type BaseContext = {
     branchNameTemplate?: string;
     useStickyComment: boolean;
     classifyInlineComments: boolean;
+    jobId: string;
     useCommitSigning: boolean;
     sshSigningKey: string;
     botId: string;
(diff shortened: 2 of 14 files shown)


ashwin-ant:
Thanks for this! I tested this a bit and hit an issue.

With tag mode (`@claude` mentions on a PR) and `use_sticky_comment: true`, the bot header (`<!-- bot: claude -->`) is correctly added to the initial comment and preserved through the update lifecycle. `GITHUB_JOB` is correctly passed and used as the bot identifier. Non-sticky flow also still works fine. All good there.

Sticky comment reuse doesn't work for `issue_comment` events though. The condition on line 39 of `create-initial.ts` checks `isPullRequestEvent(context)`, which only returns `true` when `eventName === "pull_request"`. But `@claude` mentions on PRs fire as `issue_comment` events, so the lookup path is never reached and a new comment is always created. On my second `@claude` mention, a brand new comment was created instead of reusing the existing one, even though both had matching `<!-- bot: claude -->` headers. The fix is probably to check `context.isPR` instead of `isPullRequestEvent(context)`, since `context.isPR` is true for issue_comment events on PRs too. (This bug also existed before this PR, so it's not a regression you introduced.)

A couple of smaller things:
- The `"i"` flag on the header regex makes matching case-insensitive, but job IDs should probably match case-sensitively to avoid collisions between jobs like `build` and `Build`.
- `extractBotHeader` makes the trailing newline optional (`\n?`), but if the original header lacks one, the concatenation in `updateCommentBody` produces `<!-- bot: foo -->**Claude finished...` with no line break.

ashwin-ant (changes requested):
see comment

TosinAF:
@ashwin-ant i resolved the issues you described above! cheers

PR #988, opened 2026-03-01
fix: handle shell-quote treating # as comment in claude_args

### LINKED ISSUE #980: shell-quote treats # in claude_args values as comments, causing null values and AJV crash
**Describe the bug**

The action uses `shell-quote`'s `parse()` to process `claude_args`. When a flag value contains `#` (even if backslash-escaped), `shell-quote` treats it as a comment and returns a `{comment: "..."}` object instead of a string. The action's `.filter(arg => typeof arg === 'string')` then drops it, leaving the flag with no value, which the parser sets to `null`. This `null` propagates into the SDK config and triggers an AJV schema validation crash before any API calls are made.

**To Reproduce**

1. Create a workflow using `anthropics/claude-code-action@v1`
2. Set `claude_args` with a value containing `#` (note: `#` is preserved inside YAML literal blocks `|`):
   ```yaml
   claude_args: |
     --model claude-sonnet-4-5-20250929
     --append-system-prompt \#\ Use\ best\ practices
   ```
3. Run the workflow
4. The action crashes with an AJV validation error at $0 cost

You can verify the parsing behavior locally with Node.js:

```js
const { parse } = require('shell-quote');

// Backslash escaping collapses \# to #, which shell-quote treats as a comment
parse('--append-system-prompt \\#\\ Use\\ best\\ practices')
// → ["--append-system-prompt", {comment: " Use best practices"}]

// The action filters out non-string objects:
// → ["--append-system-prompt"]
// → parser sets: {"append-system-prompt": null}

// Unescaped # is also treated as a comment
parse('--append-system-prompt # Use best practices')
// → ["--append-system-prompt", {comment: " Use best practices"}]

// Double-quoted values work correctly
parse('--append-system-prompt "# Use best practices"')
// → ["--append-system-prompt", "# Use best practices"]
```

**Expected behavior**

The action should parse `#` in flag values as literal text, not as a comment. At minimum, `null` values should be rejected with a clear error instead of being passed to the SDK.

**Workflow yml file**

```yaml
name: Repro AJV Crash
on: workflow_dispatch

jobs:
  repro:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
        with:
          prompt: "Say hello"
          claude_args: |
            --model claude-sonnet-4-5-20250929
            --append-system-prompt \#\ Use\ best\ practices
```

**API Provider**

[ ] AWS Bedrock
[ ] GCP Vertex

**Additional context**

The root cause is in `base-action/src/parse-sdk-options.ts`, `parseClaudeArgsToExtraArgs()`:

```typescript
const args = parseShellArgs(claudeArgs).filter(
  (arg): arg is string => typeof arg === "string",  // drops {comment: "..."} objects
);
```

When the loop reaches a `--flag` whose value was dropped, `nextArg` is `undefined`, so the parser falls into the boolean flag branch and sets `result[flag] = null`. The resulting config:

```json
{
  "model": "claude-sonnet-4-5-20250929",
  "append-system-prompt": null
}
```

This `null` triggers the AJV crash:
```
SDK execution error: 14 | depsCount: ${X},
15 | deps: ${Y}}`};var Tj={keyword:"dependencies",type:"object",schemaType:"object",...
error: Claude Code process exited with code 1
```

**Workaround:** Double-quote values containing `#`:
```yaml
# BROKEN — backslash escaping still results in # after shell-quote parses it
claude_args: |
  --append-system-prompt \#\ Use\ best\ practices

# WORKS — double quotes prevent # from being treated as a comment
claude_args: |
  --append-system-prompt "# Use best practices"
```

May be related to #947, #892, #852 which report the same AJV error signature.

yoheidemachi:
`shell-quote`'s `parse()` treats `#` as the start of a shell comment and returns `{comment: "..."}` objects instead of strings. The existing `.filter(arg => typeof arg === 'string')` drops these objects, leaving the preceding flag with no value (`null`). This `null` propagates into the SDK config and triggers an AJV schema validation crash.

## Changes

- Convert `{comment: "..."}` objects back to strings (prepending `#`) instead of dropping them
- Add test cases for `#` in quoted and unquoted flag values

## Reproduction (before fix)

```yaml
claude_args: |
  --append-system-prompt \#\ Use\ best\ practices
```

Crashes with AJV validation error because `append-system-prompt` receives `null`.

Closes #980

### DIFF
--- a/base-action/src/parse-sdk-options.ts
+++ b/base-action/src/parse-sdk-options.ts
@@ -92,9 +92,17 @@ function parseClaudeArgsToExtraArgs(
   if (!claudeArgs?.trim()) return {};
 
   const result: Record<string, string | null> = {};
-  const args = parseShellArgs(claudeArgs).filter(
-    (arg): arg is string => typeof arg === "string",
-  );
+  const args = parseShellArgs(claudeArgs)
+    .map((arg) => {
+      if (typeof arg === "string") return arg;
+      // shell-quote treats # as a comment and returns {comment: "..."} objects.
+      // Convert these back to strings since # is valid in flag values.
+      if (typeof arg === "object" && arg !== null && "comment" in arg) {
+        return `#${(arg as { comment: string }).comment}`;
+      }
+      return undefined;
+    })
+    .filter((arg): arg is string => arg !== undefined);
 
   for (let i = 0; i < args.length; i++) {
     const arg = args[i];
--- a/base-action/test/parse-sdk-options.test.ts
+++ b/base-action/test/parse-sdk-options.test.ts
@@ -313,7 +313,8 @@ describe("parseSdkOptions", () => {
 
     test("should handle unquoted # by converting comment object back to string", () => {
       const options: ClaudeOptions = {
-        claudeArgs: "--model claude-sonnet-4-5-20250929 --append-system-prompt #Use",
+        claudeArgs:
+          "--model claude-sonnet-4-5-20250929 --append-system-prompt #Use",
       };
 
       const result = parseSdkOptions(options);
--- a/base-action/test/parse-sdk-options.test.ts
+++ b/base-action/test/parse-sdk-options.test.ts
@@ -298,6 +298,32 @@ describe("parseSdkOptions", () => {
     });
   });
 
+  describe("hash character handling", () => {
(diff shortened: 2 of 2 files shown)


ashwin-ant:
Closing in favor of #1055

PR #994, opened 2026-03-02
fix: add CLAUDE_WORKING_DIR support to main action

### LINKED ISSUE #315: Broken CLAUDE_WORKING_DIR support
**Describe the bug**
The claude-code-action supports setting `CLAUDE_WORKING_DIR` to switch to a different directory before running claude. This is important for monorepos or repositories with nested CLAUDE.md files, to ensure the CI job picks up all the relevant context.

This worked by explicitly cd'ing to `$CLAUDE_WORKING_DIR` before running claude. This code still exists in the base action: https://github.com/anthropics/claude-code-action/blob/main/base-action/action.yml#L124-L128

However, after the changes in https://github.com/anthropics/claude-code-action/pull/285 the working directory is onl set in the "inner" action, but not when running the "outer" action, resulting in claude being run from the top-level directory always.

I believe the cd to `$CLAUDE_WORKING_DIR` should be moved from the base action to the outer action to fix this.

**Expected behavior**
Claude should run in  `$CLAUDE_WORKING_DIR`

TomokiIchi:
## Summary

- Add `CLAUDE_WORKING_DIR` environment variable support to the main composite action (`action.yml`)
- The `base-action/action.yml` already has this feature (L156-160), but the main action was missing it after the refactor in #285
- This is a 4-line addition that checks for `CLAUDE_WORKING_DIR` and `cd`s to it before running the entrypoint

## Motivation

Monorepo users need to run Claude Code in a subdirectory (e.g., `systems/journey` within a larger repo). The `CLAUDE_WORKING_DIR` environment variable was designed for this, but currently only works with `claude-code-base-action`, not the main `claude-code-action`.

This causes Claude Code to:
- Start in the repository root instead of the intended subdirectory
- Miss project-specific `.claude/` configurations (CLAUDE.md, settings, skills)
- Require workarounds like prompt-based `cd` instructions

## Changes

Added the same `CLAUDE_WORKING_DIR` check from `base-action/action.yml` to the main `action.yml`:

```bash
if [ -n "$CLAUDE_WORKING_DIR" ]; then
  echo "Changing directory to CLAUDE_WORKING_DIR: $CLAUDE_WORKING_DIR"
  cd "$CLAUDE_WORKING_DIR"
fi
```

The guard (`if [ -n ... ]`) ensures no behavior change when the variable is not set.

## Usage

```yaml
  env:
    CLAUDE_WORKING_DIR: ${{ github.workspace }}/my-subdirectory
```

Fixes #315

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -186,6 +186,11 @@ runs:
       id: run
       shell: bash
       run: |
+        # Change to CLAUDE_WORKING_DIR if set (for running in custom directories)
+        if [ -n "$CLAUDE_WORKING_DIR" ]; then
+          echo "Changing directory to CLAUDE_WORKING_DIR: $CLAUDE_WORKING_DIR"
+          cd "$CLAUDE_WORKING_DIR"
+        fi
         bun run ${GITHUB_ACTION_PATH}/src/entrypoints/run.ts
       env:
         # Prepare inputs


TomokiIchi:
Hello! Thank you for maintaining this project.
I've added support for CLAUDE_WORKING_DIR to the main action to fix #315.
Please let me know if there is anything I need to change, fix, or if I missed any contribution guidelines. I'd be happy to make any necessary updates.

Thank you for your time and for this awesome project!

PR #995, opened 2026-03-02
fix: strip comment lines from claudeArgs before shell-quote parsing

### LINKED ISSUE #802: shell-quote treats # as comment, swallowing all subsequent flags in claude_args
## Bug Description

The `shell-quote` library used in `base-action/src/parse-sdk-options.ts` treats `#` as a shell comment character. When users include shell-style comments in their `claude_args`, **all content after the `#` is swallowed**, including subsequent flags on new lines.

## Reproduction

```yaml
# In workflow file
claude_args: |
  --model 'claude-haiku-4-5'
  # This is a comment
  --allowed-tools 'mcp__github_inline_comment__create_inline_comment'
```

The `--allowed-tools` flag is **never parsed** because `shell-quote` treats everything after `#` as a comment.

## Evidence

```typescript
import { parse as parseShellArgs } from "shell-quote";

const input = `--model 'claude-haiku'
# This is a comment
--allowed-tools 'Edit'`;

console.log(parseShellArgs(input));
// Output:
// [
//   "--model",
//   "claude-haiku", 
//   { comment: " This is a comment\n--allowed-tools 'Edit'" }
//   //         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//   //         The --allowed-tools flag is INSIDE the comment object!
// ]
```

## Impact

- Users cannot add documentation comments to their `claude_args`
- Any `#` character (even in comments explaining workarounds) breaks subsequent flag parsing
- This affects all flags after the comment, not just `--allowed-tools`

## Related

This bug was discovered while investigating why the workaround for #800 wasn't working. Users adding explanatory comments about the workaround inadvertently broke their configuration.

## Proposed Fix

Strip shell-style comment lines before parsing with `shell-quote`:

```typescript
function stripShellComments(input: string): string {
  return input
    .split("\n")
    .filter((line) => !line.trim().startsWith("#"))
    .join("\n");
}
```

I have a fix ready with tests and can submit a PR.

marcelsafin:
## Problem

`shell-quote`'s `parse()` treats `#` as a shell comment character, silently swallowing all content after it — including subsequent flags on new lines. This causes flags like `--allowed-tools` to be dropped when users include comments in their `claude_args` YAML block.

### Reproduction (from #802)

```yaml
claude_args: |
  --model 'claude-haiku-4-5'
  # This is a comment
  --allowed-tools 'mcp__github_inline_comment__create_inline_comment'
```

The `--allowed-tools` flag is silently dropped.

## Fix

Strip lines starting with `#` before passing to `shell-quote`'s `parse()` in `parseClaudeArgsToExtraArgs()`. This preserves the existing parsing behavior for everything else while preventing comment lines from swallowing subsequent flags.

## Tests

Added 3 test cases in `parse-sdk-options.test.ts`:
- Comment lines between flags are ignored
- Multiple interleaved comments with flags
- Flags after comments are preserved

All 656 existing tests pass.

Fixes #802

### DIFF
--- a/base-action/src/parse-sdk-options.ts
+++ b/base-action/src/parse-sdk-options.ts
@@ -92,7 +92,17 @@ function parseClaudeArgsToExtraArgs(
   if (!claudeArgs?.trim()) return {};
 
   const result: Record<string, string | null> = {};
-  const args = parseShellArgs(claudeArgs).filter(
+
+  // Strip shell-style comment lines before parsing.
+  // shell-quote treats '#' as a comment character and silently swallows
+  // everything after it, including subsequent flags on new lines.
+  // See: https://github.com/anthropics/claude-code-action/issues/802
+  const sanitizedArgs = claudeArgs
+    .split("\n")
+    .filter((line) => !line.trim().startsWith("#"))
+    .join("\n");
+
+  const args = parseShellArgs(sanitizedArgs).filter(
     (arg): arg is string => typeof arg === "string",
   );
 
--- a/base-action/test/parse-sdk-options.test.ts
+++ b/base-action/test/parse-sdk-options.test.ts
@@ -367,4 +367,51 @@ describe("parseSdkOptions", () => {
       );
     });
   });
+
+  describe("comment handling (issue #802)", () => {
+    test("should ignore shell-style comment lines in claudeArgs", () => {
+      const options: ClaudeOptions = {
+        claudeArgs: [
+          "--allowed-tools 'mcp__github__create_comment'",
+          "# This is a comment",
+          "--max-turns 5",
+        ].join("\n"),
+      };
+      const result = parseSdkOptions(options);
+
+      expect(result.sdkOptions.allowedTools).toContain(
+        "mcp__github__create_comment",
+      );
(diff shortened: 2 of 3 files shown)


marcelsafin:
Thanks for the review feedback @copilot!

**Re: shell-quote comment test** — Fixed in 7ba3507. The test now has a proper assertion documenting shell-quote's raw behavior.

**Re: quote-aware comment stripping** — Good catch. In theory, a multi-line `--system-prompt` containing Markdown headings (`# Heading`) could be affected. However, in practice:

1. `claude_args` comes from a YAML block scalar (`|`), where each line is typically a separate flag or a standalone comment
2. Multi-line quoted arguments spanning multiple YAML lines are extremely unusual in this context
3. shell-quote itself doesn't handle multi-line quoted strings well — it treats each line independently

That said, if maintainers prefer a more defensive approach, I can restrict stripping to lines that are *only* a comment (no other content), or add a note in the docs that `#` comment lines are stripped. Happy to adjust based on guidance.

ashwin-ant:
Closing in favor of #1055

PR #956, opened 2026-02-17
fix: allow # character in branch name validation

ovp87:
## Summary

- Adds `#` to the branch name validation whitelist regex in `validateBranchName()`
- The `#` character is valid per [git-check-ref-format](https://git-scm.com/docs/git-check-ref-format) and commonly used in branch naming conventions like `bugfix/#123-description`
- Since PR #736 switched git commands to `execFileSync` (no shell interpolation), `#` poses no injection risk — it's only meaningful in shell contexts (as a comment character)
- Adds test cases for branch names containing `#`

Fixes #751

PR #999, opened 2026-03-02
fix: allow @ character in branch name validation

### LINKED ISSUE #998: Branch name validation rejects @ character which is valid in git
## Description

The branch name validation in `src/github/operations/branch.ts` rejects branch names containing `@`, even though `@` is a valid character in git branch names. This causes the action to fail on PRs with `@` in the branch name.

## Steps to Reproduce

1. Create a PR from a branch with `@` in the name, e.g. `TICKET-123@add-feature`
2. The Claude PR review workflow triggers
3. The prepare step fails with:
   ```
   Prepare step failed with error: Invalid branch name: "TICKET-123@add-feature". 
   Branch names must start with an alphanumeric character and contain only alphanumeric characters, 
   forward slashes, hyphens, underscores, or periods.
   ```

## Root Cause

The allowlist regex in `src/github/operations/branch.ts`:

```js
const validPattern = /^[a-zA-Z0-9][a-zA-Z0-9/_.-]*$/;
```

Does not include `@` in the allowed character set. While `@{` is a special git reflog syntax (and is correctly rejected separately), a standalone `@` or `@` followed by non-`{` characters is perfectly valid in git branch names.

## Expected Behavior

Branch names containing `@` (but not `@{`) should be accepted. The regex could be updated to:

```js
const validPattern = /^[a-zA-Z0-9][a-zA-Z0-9/_.@-]*$/;
```

With the existing `@{` check still catching the unsafe git reflog syntax.

## Environment

- Action version: v1.0.27
- Event: `pull_request`

## Impact

Any PR with `@` in the branch name will fail the Claude review workflow. This is a common pattern in some teams (e.g. `TICKET@description` convention).

msyed-godaddy:
## Summary

- Updates the branch name validation regex to allow `@` in branch names
- The `@` character is valid in git branch names and commonly used in workflows (e.g., `TICKET-123@add-feature`)
- The dangerous `@{` sequence (git reflog syntax) is already caught by a separate dedicated check, so allowing `@` in the whitelist regex is safe
- Adds test cases for valid branch names containing `@`

Fixes #998

## Test plan

- [x] All existing tests pass (31 tests, 0 failures)
- [x] New test cases added for `@` in branch names (`TICKET-123@add-feature`, `user@feature-branch`, `feature@v2`)
- [x] Existing `@{` rejection test still passes (caught by the separate `@{` check)
- [ ] Manual verification: branch names like `TICKET-123@add-feature` no longer cause validation errors

### DIFF
--- a/src/github/operations/branch.ts
+++ b/src/github/operations/branch.ts
@@ -28,7 +28,7 @@ function extractFirstLabel(githubData: FetchDataResult): string | undefined {
  *
  * Valid branch names:
  * - Start with alphanumeric character (not dash, to prevent option injection)
- * - Contain only alphanumeric, forward slash, hyphen, underscore, or period
+ * - Contain only alphanumeric, forward slash, hyphen, underscore, period, or at sign
  * - Do not start or end with a period
  * - Do not end with a slash
  * - Do not contain '..' (path traversal)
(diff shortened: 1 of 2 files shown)


ovp87:
very similar to #956

Issue #896, opened 2026-02-03
[Regression in v1.0.42] setup-bun token usage broken on GHES


ef-trss:
**Describe the bug**
[This commit](https://github.com/anthropics/claude-code-action/commit/4ce5f178c22b5a6494fb3246ecd96bce6ac99c67) (merged in #861) breaks installs GitHub Enterprise Server (GHES) because it expects a token valid for github.com.

**To Reproduce**

1. Run GitHub Enterprise Server
2. Setup GitHub Actions workflow using anthropics/claude-code-action per official docs
3. Trigger action

**Expected behavior**
This action, along with the setup-bun action, should minimize use of GitHub API to avoid API limits rather than requiring token usage.  Alternatively, expose a separate input to be passed to setup-bun rather than defaulting to the token passed to claude-code-action as this one is needed for interacting with GHES repositories.

**Screenshots**

<img width="789" height="191" alt="Image" src="https://github.com/user-attachments/assets/7bd10f26-04c5-4d31-bb13-5d1b426ed1a9" />

**Workflow yml file**

<img width="380" height="249" alt="Image" src="https://github.com/user-attachments/assets/0e5b3b19-4acb-4e53-9316-c0592422326c" />

**API Provider**

[ ] Anthropic First-Party API (default)
[ ] AWS Bedrock
[ ] GCP Vertex

**Additional context**
N/A

peloyeje:
Apologies everyone, noticed this bug while testing https://github.com/anthropics/claude-code-action/pull/861 internally but forgot to push latest changes to the PR before it was merged...
Actual fix is in https://github.com/anthropics/claude-code-action/pull/901

uda832:
Can confirm that this exact error also occurs on GitHub Enterprise (GHE) as well

What `actions/create-github-app-token@v1` shows in the logs:
```
Run actions/create-github-app-token@v1
  with:
    app-id: ***
    private-key: ***
  
    github-api-url: https://api.{name of our domain}.ghe.com/
```
---

Error:
```txt
Error: Error: Failed to fetch url https://api.github.com/repos/oven-sh/bun/git/refs/tags. (status code: 401, status text: Unauthorized)
{
  "message": "Bad credentials",
  "documentation_url": "https://docs.github.com/rest",
  "status": "401"
}
```

luka5:
I was running into the same issue today and created https://github.com/anthropics/claude-code-action/pull/1001 to resolve it :)

dr-goo:
[...]

PR #913, opened 2026-02-09
docs: add FAQ item about bun install on GHES

peloyeje:
## Summary

Relying on the built-in `setup-bun` action to install bun in GHES setups often which fails with 403 rate limit errors as the calls to github.com are not authenticated.

This PR adds a doc section to offer guidance to install bun in such environments and pass the executable as an action parameter

It supersedes https://github.com/anthropics/claude-code-action/pull/861 and https://github.com/anthropics/claude-code-action/pull/901

PR #1001, opened 2026-03-03
Use default github token only on public github instance

luka5:
This fixes https://github.com/anthropics/claude-code-action/issues/896 and replaces https://github.com/anthropics/claude-code-action/pull/913 the same way as it is done in oven-sh/setup-bun itself: https://github.com/oven-sh/setup-bun/pull/157

Tested and works with GitHub Enterprise Server


luka5:
Update/second commit: Turns out, if you are on GHES and use bedrock, you must still provide the input `github_token`, otherwise we run into https://github.com/anthropics/claude-code-action/blob/main/src/github/token.ts#L77-L80 - even if we would not expect it.

jordevo (approved):

Issue #1020, opened 2026-03-06
validateBranchName rejects valid non-ASCII (e.g. Japanese) branch names


HayashidaShun:
## Description

`validateBranchName` in `src/github/operations/branch.ts` uses a strict ASCII-only regex pattern:

```typescript
const validPattern = /^[a-zA-Z0-9][a-zA-Z0-9/_.-]*$/;
```

This rejects branch names containing non-ASCII characters (e.g. Japanese), even though Git and GitHub fully support them.

## Steps to Reproduce

1. Create a PR from a branch with a non-ASCII name (e.g. `feat/add-機能追加`)
2. Trigger any workflow using `claude-code-action@v1` on that PR
3. Action fails with:
   ```
   Invalid branch name: "feat/add-機能追加". Branch names must start with an alphanumeric character and contain only alphanumeric characters, forward slashes, hyphens, underscores, or periods.
   ```

## Expected Behavior

The action should accept valid Git branch names containing non-ASCII characters (Unicode letters).

## Suggested Fix

Update the regex to allow Unicode letters while still blocking dangerous characters:

```typescript
const validPattern = /^[\p{L}\p{N}][\p{L}\p{N}/_.-]*$/u;
```

This maintains the security intent (blocking control characters, shell metacharacters, etc.) while supporting internationalized branch names.

Lubrsy706:
I'm working on this together with #998 (same regex). Will submit a PR that updates the whitelist pattern to support Unicode letters (`\p{L}`) and the `@` character.

HayashidaShun:
Hi, just a gentle follow-up on this issue.
Any updates or feedback would be appreciated. Thanks!

mrudelle:
hitting the same issue with parentheses in base branch names like `fix(scope)/the-branch-name`
note, there are various dupes of that issue: #751 #1024 #998

PR #1022, opened 2026-03-06
fix: allow @, +, and Unicode branch names

### LINKED ISSUE #998: Branch name validation rejects @ character which is valid in git

Lubrsy706:
## Summary

Fixes #998 and #1020

The whitelist regex in `validateBranchName` (`src/github/operations/branch.ts`) previously rejected valid git branch names containing `@`, `+`, or non-ASCII Unicode characters.

### Changes

- update `src/github/operations/branch.ts` to allow `@`, `+`, and Unicode letters/numbers in branch names
- keep rejecting Git-reserved patterns like `@{`, `.lock`, `..`, consecutive `/`, and invalid ref characters
- extend `test/validate-branch-name.test.ts` with coverage for `+` alongside the existing `@` and Unicode cases

### Security

The `@{` reflog syntax is still rejected explicitly. Control characters, shell metacharacters, and Git-special characters like `~`, `^`, `:`, `?`, `*`, `[`, `]`, and `\` remain blocked. The validator still forbids names that could trigger option injection or path traversal.

## Test plan

- [x] `bun test test/validate-branch-name.test.ts`
- [x] `bun run typecheck`

### DIFF
--- a/src/github/operations/branch.ts
+++ b/src/github/operations/branch.ts
@@ -28,7 +28,7 @@ function extractFirstLabel(githubData: FetchDataResult): string | undefined {
  *
  * Valid branch names:
  * - Start with alphanumeric character (not dash, to prevent option injection)
- * - Contain only alphanumeric, forward slash, hyphen, underscore, period, or hash (#)
+ * - Contain only alphanumeric, forward slash, hyphen, underscore, period, hash (#), @, or plus
  * - Do not start or end with a period
  * - Do not end with a slash
  * - Do not contain '..' (path traversal)
@@ -59,9 +59,10 @@ export function validateBranchName(branchName: string): void {
   }
 
   // Strict whitelist pattern: alphanumeric or Unicode letter start, then those plus
-  // slash/hyphen/underscore/period/#/@/+. `#`, `@`, and `+` are valid in git branch
-  // names; only `@{` is reserved (checked separately below). Unicode letters (\p{L})
-  // support non-ASCII branch names (e.g. Japanese, Chinese characters).
+  // slash/hyphen/underscore/period/@/+. `@` and `+` are valid in git branch names;
+  // only `@{` is reserved (checked separately below). Unicode letters (\p{L}) support
+  // non-ASCII branch names (e.g. Japanese, Chinese characters).
+  // `#` is also valid per git-check-ref-format and commonly used in issue-linked names.
   const validPattern = /^[\p{L}\p{N}][\p{L}\p{N}/_.#\-@+]*$/u;
 
   if (!validPattern.test(branchName)) {
--- a/src/github/operations/branch.ts
+++ b/src/github/operations/branch.ts
@@ -58,14 +58,15 @@ export function validateBranchName(branchName: string): void {
     );
   }
 
-  // Strict whitelist pattern: alphanumeric start, then alphanumeric/slash/hyphen/underscore/period/hash.
-  // # is valid per git-check-ref-format and commonly used in branch names like "fix/#123-description".
-  // All git calls use execFileSync (not shell interpolation), so # carries no injection risk.
-  const validPattern = /^[a-zA-Z0-9][a-zA-Z0-9/_.#-]*$/;
+  // Strict whitelist pattern: alphanumeric or Unicode letter start, then those plus
+  // slash/hyphen/underscore/period/#/@/+. `#`, `@`, and `+` are valid in git branch
+  // names; only `@{` is reserved (checked separately below). Unicode letters (\p{L})
+  // support non-ASCII branch names (e.g. Japanese, Chinese characters).
+  const validPattern = /^[\p{L}\p{N}][\p{L}\p{N}/_.#\-@+]*$/u;
 
   if (!validPattern.test(branchName)) {
     throw new Error(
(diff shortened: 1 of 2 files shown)


Lubrsy706:
Friendly ping: this PR has been ready for review since March 6, 2026. If there is anything blocking it or if you would prefer changes before merge, let me know and I can update it. Would appreciate a review when someone has time.

Lubrsy706:
Updated this PR to include `+` as well. Git accepts branch names like `feature+x`, so the validator now allows `@`, `+`, and Unicode branch names while still rejecting reserved/unsafe patterns like `@{`, `.lock`, `..`, and invalid ref characters. I also added test coverage for the `+` case and reran the branch-name tests plus `typecheck`.

Lubrsy706:
Following up on this PR. The branch-name handling change should be ready for review; please let me know if you'd like any updates or additional test coverage from my side.

Lubrsy706:
Rebased this branch onto the latest `main` and resolved the merge conflicts. The branch-name validation changes now include `#`, `@`, `+`, and Unicode support together, and I reran `bun test test/validate-branch-name.test.ts` after the rebase. PR should now be conflict-free and ready for review.

Lubrsy706:
Following up on this PR. The branch-name handling fix should be ready for review; if you would like any adjustments or broader coverage for special-character branch names, I can update it. Would appreciate a review when someone has time.

aki310sony:
Adding a data point from another affected repository, since this has been open since March.

Our branch names are Japanese by convention and this is not something we can change — they are derived from issue titles, e.g.:

```
feature/346-sensor_settings-isp_settings-への設定変更が値検証を通らず設定ファイルを壊せる
```

Every PR from such a branch fails the review job:

```
Error: Action failed with error: Invalid branch name: "feature/346-...".
Branch names must start with an alphanumeric character, underscore, or '@' and
contain only alphanumeric characters, forward slashes, hyphens, underscores,
periods, hashes (#), plus signs (+), commas (,), at signs (@), or parentheses.
```

Two observations that may help prioritise this:

**The `@` / `+` / `#` parts of this PR have already landed separately** (#1411, #1244, #956-adjacent work), so what remains unmerged here is only the Unicode part. The whitelist has been widened one ASCII character class at a time — `@`, `_`, `,`, `+`, parentheses — while non-ASCII, which is equally valid per `git-check-ref-format`, is the one case still rejected.

**The comment in `validateBranchName` itself states the constraint is no longer needed for safety:**

> All git calls use execFileSync (not shell interpolation), so none of these characters carry injection risk.

The reachable path for us is `setupBranch()` in tag mode, which validates the head branch name unconditionally for open PRs before any git call. Our workaround is to drop `track_progress` so the action runs in agent mode, which does not call `setupBranch()` — but that also removes the tracking comment, which is where the review output was being delivered.

Happy to test a fix against real Japanese branch names if that would help.

Lubrsy706:
Thanks @aki310sony for the concrete example and offer to test. I've merged the latest `main`, resolved the conflicts, and updated this PR to focus on the remaining Unicode support in e6b15635c756284e84ea9510cdc85675289ed485. The ASCII characters and leading `_` / `@` already supported upstream remain accepted.

Your exact Japanese branch name is now a regression case. I also covered decomposed characters (including Japanese dakuten) and invalid Unicode-containing names without removing the existing safety checks.

Validation:
- `bun test`: 983 passed, 0 failed
- `bun run typecheck` and `bun run format:check`: passed
- Local Git smoke checks through the actual `setupBranch()` path: your branch fetched and checked out successfully for both same-repository and fork PRs, with a Unicode base branch

Would you be able to test this revision in your repository with `track_progress` enabled? The local checks do not replace that end-to-end workflow validation. Maintainers: this is now a two-file diff against current `main`; I'd appreciate a review of the remaining Unicode change.

PR #1047, opened 2026-03-11
fix: truncate step summary to stay under GitHub's 1024k limit

### LINKED ISSUE #927: GITHUB_STEP_SUMMARY exceeds 1024k limit on long sessions
## Bug Description

When Claude runs a long session (1h+ with many tool calls), the step summary output exceeds GitHub's 1024k hard limit, causing the error:

```
$GITHUB_STEP_SUMMARY upload aborted, supports content up to a size of 1024k, got 1026k.
```

## Root Cause

In `src/entrypoints/run.ts`, `writeStepSummary()` formats the entire execution output into markdown via `formatTurnsFromData()` and appends it to `$GITHUB_STEP_SUMMARY` with no total size check.

While `formatResultContent()` in `format-turns.ts` truncates individual tool results to 3000 chars, the total accumulated markdown from many turns can easily exceed 1MB.

## Steps to Reproduce

1. Tag `@claude` on a complex issue that requires many tool calls (planning a multi-step feature, etc.)
2. Let the session run for 30+ minutes
3. The step summary upload fails with the 1024k error

## Expected Behavior

The step summary should be truncated to stay under GitHub's 1024k limit. Possible approaches:
- Truncate the formatted markdown to ~900k with a "... truncated ..." message
- Collapse older turns and only show recent activity in detail
- Add an input option to control max summary size (related: #206)

## Actual Behavior

The action errors with the `$GITHUB_STEP_SUMMARY upload aborted` message. The error is non-fatal (the Claude work itself still succeeds), but it's noisy and prevents viewing the execution summary.

## Environment

- `anthropics/claude-code-action@v1`
- Session duration: ~1h with many tool calls
- Generated summary size: 1026k (just over the 1024k limit)

sunnypatell:
long Claude sessions (30+ min, many tool calls) generate step summary markdown that exceeds GitHub's hard 1024k limit for `$GITHUB_STEP_SUMMARY`, causing:

```
$GITHUB_STEP_SUMMARY upload aborted, supports content up to a size of 1024k, got 1026k.
```

the error is non-fatal (Claude's work still succeeds), but it's noisy and prevents viewing the execution summary. users currently have to set `display_report: false` to avoid it entirely, losing the summary for all sessions.

fixes #927

### what changed

added `truncateStepSummary()` in `format-turns.ts` that:

1. checks the markdown's byte size against a 1000k budget (leaves 24k headroom for content from earlier steps in the same job)
2. if over budget, cuts at the last section boundary (`---`) to avoid breaking mid-content
3. appends a truncation notice explaining the limit and pointing to `display_report: false`
4. handles multi-byte characters correctly via `TextEncoder`/`TextDecoder` (no mid-codepoint cuts)

applied in both the formatted report path and the raw JSON fallback path in `writeStepSummary()`.

### tests

5 new tests in `format-turns.test.ts`:

- under-limit content passes through unchanged
- over-limit content is truncated with notice
- cuts at last section separator when one exists in the latter half
- multi-byte characters (emoji) survive without corruption
- default limit matches GitHub's constraint

all 35 tests pass.

### DIFF
--- a/src/entrypoints/format-turns.ts
+++ b/src/entrypoints/format-turns.ts
@@ -412,6 +412,46 @@ export function formatGroupedContent(groupedContent: GroupedContent[]): string {
   return markdown;
 }
 
+// GitHub's hard limit for $GITHUB_STEP_SUMMARY content.
+// Using 1000k instead of 1024k to leave headroom for any existing content
+// that may have been appended by earlier steps in the same job.
+const STEP_SUMMARY_MAX_BYTES = 1000 * 1024;
+
+/**
+ * Truncate markdown to fit within GitHub's step summary size limit (1024k).
+ * Cuts at the last complete section boundary (---) before the limit,
+ * then appends a truncation notice.
+ */
+export function truncateStepSummary(
+  markdown: string,
+  maxBytes: number = STEP_SUMMARY_MAX_BYTES,
+): string {
+  const encoded = new TextEncoder().encode(markdown);
+  if (encoded.byteLength <= maxBytes) {
+    return markdown;
+  }
+
+  const truncationNotice =
+    "\n\n---\n\n" +
+    "> **Note:** This report was truncated to fit within GitHub's step summary size limit (1024k).\n" +
+    "> To disable the report entirely, set `display_report: false` in your workflow.\n";
+
+  const noticeBytes = new TextEncoder().encode(truncationNotice).byteLength;
+  const budget = maxBytes - noticeBytes;
+
+  // Decode back to string at the byte budget boundary
+  const truncated = new TextDecoder().decode(encoded.slice(0, budget));
+
+  // Find the last section separator to cut at a clean boundary
+  const lastSeparator = truncated.lastIndexOf("\n---\n");
+  const cutPoint =
+    lastSeparator > truncated.length * 0.5
+      ? lastSeparator
+      : truncated.length;
+
+  return truncated.substring(0, cutPoint) + truncationNotice;
+}
+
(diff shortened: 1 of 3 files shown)


sunnypatell:
addressing copilot's 5 inline comments (3 fixed in d5b55ae, 2 intentionally kept as-is):

**fixed in d5b55ae:**

1. **budget going negative** (format-turns.ts:443) - added a guard: if `maxBytes <= noticeBytes`, returns just the notice trimmed to fit. handles the edge case where the limit is absurdly small.

2. **replacement char inflating encoded size** (format-turns.ts:453) - added a re-verification loop that trims one character at a time if the re-encoded result exceeds the budget after `TextDecoder` replaces incomplete multi-byte sequences with U+FFFD.

3. **test name mismatch** (format-turns.test.ts:469) - renamed to "uses default limit with headroom below GitHub's 1024k constraint" to reflect the intentional 1000k budget.

4. **tighter section boundary assertions** (format-turns.test.ts:450) - added `not.toContain("Section 3")` and `not.toContain("C".repeat(50))` to validate Section 3 content was excluded. also fixed the limit calculation so it actually triggers the boundary cut.

**intentionally not changed:**

5. **checking existing summary file size** (run.ts:108) - the 24k headroom (1000k vs 1024k) covers realistic earlier-step content. `stat`-ing the summary file on every write adds I/O and complexity for a scenario that effectively doesn't happen in practice (earlier steps would need to write 24k+ of summary, which is extremely unusual for steps that precede Claude). if this ever becomes a real problem, the fix would be to reduce the budget further, not to add file I/O.

PR #1083, opened 2026-03-18
fix: use repo default branch instead of hardcoded 'main' in agent mode

### LINKED ISSUE #1080: restoreConfigFromBase fails on repos with non-"main" default branch
## Bug

When `claude-code-action` runs on a repository whose default branch is `master` (or anything other than `main`), the config restore step fails with:

```
fatal: couldn't find remote ref main
Error: Action failed with error: Command failed: git fetch origin main --depth=1
```

This happens because `baseBranch` falls back to hardcoded `"main"` in several places when no explicit `base_branch` input is provided.

## Affected locations

| File | Line(s) | Code |
|------|---------|------|
| `src/modes/agent/index.ts` | ~89, ~96 | `process.env.BASE_BRANCH \|\| context.inputs.baseBranch \|\| "main"` |
| `src/entrypoints/run.ts` | ~315 | `baseBranch \|\| "main"` |
| `src/entrypoints/update-comment-link.ts` | ~256 | `process.env.BASE_BRANCH \|\| "main"` |

## Expected behavior

When `base_branch` is not provided, the action should detect the repository's actual default branch instead of assuming `"main"`.

## The fix already exists in the codebase

`src/github/operations/branch.ts` (~line 185) already implements the correct pattern:

```typescript
const repoResponse = await octokits.rest.repos.get({ owner, repo });
sourceBranch = repoResponse.data.default_branch;
```

Alternatively, `${{ github.event.repository.default_branch }}` is available in the Actions context and could be used as the fallback instead of hardcoding `"main"`.

## Workaround

Callers can pass `base_branch` explicitly:

```yaml
  with:
    base_branch: ${{ github.event.repository.default_branch }}
```

## Reproduction

1. Use `claude-code-action@v1` on a repo whose default branch is `master`
2. Trigger via PR comment
3. Action fails at the config restore step before Claude even runs

CervEdin:
\`restoreConfigFromBase\` (added in v1.0.74 via #1066) fetches config
files from the PR base branch as a security measure. In agent mode,
\`baseBranch\` falls back to \`"main"\` when neither \`BASE_BRANCH\` nor
the action's \`baseBranch\` input is set. On repos whose default branch
is \`master\` (or anything else), this causes \`git fetch origin main\`
to fail, which aborts the entire run.

Read the repo's actual default branch from the event payload
(\`repository.default_branch\`) instead of assuming \`"main"\`. The
payload includes this field on every event type GitHub fires.

Fixes #1080

### DIFF
--- a/src/modes/agent/index.ts
+++ b/src/modes/agent/index.ts
@@ -1,4 +1,5 @@
 import { mkdir, writeFile } from "fs/promises";
+import * as github from "@actions/github";
 import { prepareMcpConfig } from "../../mcp/install-mcp-server";
 import { parseAllowedTools } from "./parse-tools";
 import {
@@ -85,15 +86,18 @@ export async function prepareAgentMode({
 
   // Check for branch info from environment variables (useful for auto-fix workflows)
   const claudeBranch = process.env.CLAUDE_BRANCH || undefined;
+  const repoDefaultBranch =
+    (github.context.payload.repository as { default_branch?: string })
+      ?.default_branch || "main";
   const baseBranch =
-    process.env.BASE_BRANCH || context.inputs.baseBranch || "main";
+    process.env.BASE_BRANCH || context.inputs.baseBranch || repoDefaultBranch;
 
   // Detect current branch from GitHub environment
   const currentBranch =
     claudeBranch ||
     process.env.GITHUB_HEAD_REF ||
     process.env.GITHUB_REF_NAME ||
-    "main";
+    repoDefaultBranch;
 
   // Get our GitHub MCP servers config
   const ourMcpConfig = await prepareMcpConfig({


CervEdin:
NGL, I feel like there should be some credit compensation or whatever when I need to use CC to fix anthropics bugs 🤷

CervEdin:
Note that this may not be a complete fix, as noted in #1080 

> ## Affected locations
> File	Line(s)	Code
> `src/modes/agent/index.ts`	~89, ~96	`process.env.BASE_BRANCH || context.inputs.baseBranch || "main"`
> `src/entrypoints/run.ts`	~315	`baseBranch || "main"`
> `src/entrypoints/update-comment-link.ts`	~256	`process.env.BASE_BRANCH || "main"`
> ## Expected behavior
> When `base_branch` is not provided, the action should detect the repository's actual default branch instead of assuming `"main"`.

other locations might need similar handling

ping @ashwin-ant I think this regression might be related to your changes in 7057f3318b938a2dd095fd89f786c11772b08197

PR #1093, opened 2026-03-20
Auto-set subprocess env scrub when allowed_non_write_users is configured

OctavianGuzu:
Sets `CLAUDE_CODE_SUBPROCESS_ENV_SCRUB=1` automatically whenever `allowed_non_write_users` is non-empty. The CLI (v2.1.79+) reads this flag to do a best-effort scrub of Anthropic, cloud, and GitHub Actions secrets from the environment of subprocesses it spawns (Bash tool, hooks, MCP stdio, etc.). The parent process keeps the vars for its own API calls; only child subprocess environments are scrubbed.

Updates the input description and docs/security.md to note the behavior.

Downstream workflows that set `allowed_non_write_users` get this automatically — no workflow-file changes needed.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -32,7 +32,16 @@ inputs:
     required: false
     default: ""
   allowed_non_write_users:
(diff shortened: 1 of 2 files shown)


wolffiex (approved):
Please make sure this gets a beefy release note

PR #1106, opened 2026-03-23
Add `list_inline_comments` tool to inline comment MCP server

nurmukhametov:
Adds a new MCP tool to list PR review comments, with optional `user_login` filter and `limit` parameter for early exit. This lets Claude read prior inline comments without shelling out to `gh api`.


nurmukhametov:
Hi @ashwin-ant, following up on this PR. As a user, I have found this feature is currently missing and would like to get it merged. Is this something you are open to including? Thanks!

Issue #759, opened 2025-12-20
[BUG] CLAUDE_APP_BOT_ID constant uses github-actions[bot] ID (41898282) instead of claude[bot] ID (209825114)


HipHoff:
## Describe the bug

The `CLAUDE_APP_BOT_ID` constant in `src/github/constants.ts` is set to `41898282`, which is actually the ID for `github-actions[bot]`, not `claude[bot]`.

The correct ID for `claude[bot]` is `209825114`.

## Evidence

Verified via GitHub API:
```json
{"id": 41898282, "login": "github-actions[bot]", "type": "Bot"}
{"id": 209825114, "login": "claude[bot]", "type": "Bot"}
```

Looking at `action.yml`, the defaults are:
- `bot_id: "41898282"` ← WRONG (this is github-actions[bot])
- `bot_name: "claude[bot]"` ← Correct name, but mismatched with ID

## Impact

When commits are made via git CLI (fallback when API commit signing fails), the commits have:
- Author: `41898282+claude[bot]@users.noreply.github.com`
- This creates an identity mismatch and results in **unsigned commits**

Compare to correctly-signed commits made via GitHub API:
- Author: `209825114+claude[bot]@users.noreply.github.com`  
- Committer: `GitHub <noreply@github.com>`
- These are **verified/signed** by GitHub

## Reproduction

1. Enable `use_commit_signing: true` in a workflow
2. Trigger Claude to make a commit
3. In some cases, commits fall back to git CLI instead of API
4. These commits show as unsigned in GitHub

## Workaround

Explicitly set the correct `bot_id` in workflow:
```yaml
- uses: anthropics/claude-code-action@v1
  with:
    use_commit_signing: true
    bot_id: "209825114"  # ← Workaround: correct claude[bot] ID
```

## Suggested Fix

Update `src/github/constants.ts`:
```typescript
// Before (incorrect)
export const CLAUDE_APP_BOT_ID = 41898282;

// After (correct)
export const CLAUDE_APP_BOT_ID = 209825114;
```

And update the default in `action.yml`:
```yaml
bot_id:
  description: "GitHub user ID to use for git operations"
  required: false
  default: "209825114"  # claude[bot] ID
```

## Related

This may be related to #281 / #282 (commits not verified), though those issues focus on when git CLI is used rather than the incorrect bot ID.

nsheaps:
~~It looks like this also breaks `use_sticky_comment` as it's trying to find comments from the wrong user ID.~~
[...]

PR #1123, opened 2026-03-25
fix: use configurable bot ID for sticky comment search

sxlehman-godaddy:
## Problem

`use_sticky_comment: "true"` never finds existing comments to update, causing duplicate comments on every workflow run. Two independent bugs cause this:

### 1. Hardcoded bot ID mismatch

`create-initial.ts` defines a local `CLAUDE_APP_BOT_ID = 209825114` (`claude[bot]`) that shadows the shared constant in `constants.ts` (`41898282`, `github-actions[bot]`). The comment search uses this value to match `comment.user?.id`, but the action's default `bot_id` is `"41898282"`. Under the default configuration, the search never matches.

**Fix**: Use `context.inputs.botId` (configurable via the `bot_id` input) instead of the hardcoded constant.

### 2. Event type guard too restrictive

The sticky comment guard requires `isPullRequestEvent(context)`, which only returns `true` for `eventName === "pull_request"`. Comment-triggered workflows (`issue_comment` events on PRs) are excluded even though `context.isPR` is `true`.

**Fix**: Widen the guard to also match `issue_comment` events on PRs via `isIssueCommentEvent(context) && context.isPR`. Review-thread reply behavior (`pull_request_review_comment` events) is unchanged — they continue to use `createReplyForReviewComment` as before.

## Changes

- `src/github/operations/comments/create-initial.ts`: Removed hardcoded bot ID, use `context.inputs.botId`; widened sticky comment guard to include PR issue comments
- `test/create-initial-comment.test.ts`: New test file with 5 regression tests covering sticky comment routing across event types

## Verification

- `bun run format:check` — passes
- `bun test` — 657/657 tests pass (5 new)

## Impact

Affects every user running with the default `github_token` who enables `use_sticky_comment` or `track_progress`. Previously, duplicate comments were created on every workflow run.

Related: #759 (same root cause — bot ID mismatch — different symptom: unsigned commits)

### DIFF
--- a/src/github/operations/comments/create-initial.ts
+++ b/src/github/operations/comments/create-initial.ts
@@ -8,13 +8,12 @@
 import { appendFileSync } from "fs";
 import { createJobRunLink, createCommentBody } from "./common";
 import {
-  isPullRequestReviewCommentEvent,
+  isIssueCommentEvent,
   isPullRequestEvent,
+  isPullRequestReviewCommentEvent,
   type ParsedGitHubContext,
 } from "../../context";
 import type { Octokit } from "@octokit/rest";
-
-const CLAUDE_APP_BOT_ID = 209825114;
 
 export async function createInitialComment(
   octokit: Octokit,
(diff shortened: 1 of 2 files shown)


sxlehman-godaddy:
@ashwin-ant Hi, any chance you could take a look? This has been open since March. We are inelegantly deleting comments downstream to get the behavior, this seems like it would be quite helpful to a lot of folks to do it on the right level, here. Happy to rebase, split, or address feedback.

Issue #1133, opened 2026-03-30
allowed_bots bypasses actor check but not permission check — bot actors still fail with 404


bhill-godaddy:
## Summary

`allowed_bots` bypasses the human actor check (`checkHumanActor` in `actor.ts`) but does **not** bypass the write permission check (`checkWritePermissions` in `permissions.ts`). When a bot like GitHub Copilot triggers a `pull_request_review` event, the action fails with a 404 from the collaborators API because bot accounts are not repository collaborators.

## Reproduction

1. Configure a workflow triggered by `pull_request_review` with `allowed_bots: "Copilot"`
2. Request a Copilot review on a PR
3. Copilot submits its review, triggering the workflow with `actor: Copilot`

**Expected:** The action runs successfully, with `allowed_bots` bypassing both the actor check and the permission check.

**Actual:** The action fails:
```
Checking permissions for actor: Copilot
GET /repos/{owner}/{repo}/collaborators/Copilot/permission - 404
##[error] Failed to check permissions: HttpError: Copilot is not a user
##[error] Action failed with error: Actor does not have write permissions to the repository
```

## Root cause

In `src/github/validation/permissions.ts`, `checkWritePermissions` has three bypass paths:

1. `allowedNonWriteUsers` match — but this uses a separate input (`allowed_non_write_users`), not `allowed_bots`
2. `actor.endsWith("[bot]")` — but Copilot's login is `Copilot`, not `copilot-pull-request-reviewer[bot]`
3. Collaborators API returns `admin` or `write` — but bot accounts are not collaborators, so the API returns 404

None of these paths handle a bot that is listed in `allowed_bots` but doesn't have the `[bot]` suffix.

Meanwhile, `checkHumanActor` in `actor.ts` correctly handles `allowed_bots` — it fetches the user type via `octokit.users.getByUsername`, sees `type: "Bot"`, and checks the `allowed_bots` list. But execution never reaches that point because `checkWritePermissions` throws first.

The call order in `src/entrypoints/run.ts` (line ~175):
```typescript
// checkWritePermissions runs first and throws for bot actors
const hasWritePermissions = await checkWritePermissions(
  octokit.rest, context, context.inputs.allowedNonWriteUsers, ...
);

// checkHumanActor runs later (in prepareAgentMode) and correctly checks allowed_bots
await checkHumanActor(octokit.rest, context);
```

## Suggested fix

[...]

Issue #1018, opened 2026-03-05
`checkWritePermissions` fails with 404 for non-user actors like `Copilot`


goodtune:
## Describe the bug

`checkWritePermissions()` in `src/github/validation/permissions.ts` calls `octokit.repos.getCollaboratorPermissionLevel()` with `github.actor` as the username. When the actor is `Copilot` (from a Copilot-initiated `pull_request_review`), this API call returns 404 because `Copilot` is not a regular GitHub user.

The function has a bypass for actors ending in `[bot]` (line 47), but `Copilot` doesn't match that pattern.

## To Reproduce

1. Configure `anthropics/claude-code-action@v1` on a `pull_request_review` workflow
2. Pass a `github_token` input (to bypass OIDC exchange — see #1017)
3. Have GitHub Copilot submit a review, triggering the workflow with `github.actor = "Copilot"`
4. The `checkWritePermissions` call fails with 404

## Error Log

```
Using provided GITHUB_TOKEN for authentication
Checking permissions for actor: Copilot
GET /repos/goodtune/ghp/collaborators/Copilot/permission - 404 with id AC45:74D71:1143342:4A95233:69A9F64B in 157ms
##[error]Failed to check permissions: HttpError: Copilot is not a user - https://docs.github.com/rest/collaborators/collaborators#get-repository-permissions-for-a-user
##[error]Action failed with error: Failed to check permissions for Copilot: HttpError: Copilot is not a user - https://docs.github.com/rest/collaborators/collaborators#get-repository-permissions-for-a-user
```

- Failing run: https://github.com/goodtune/ghp/actions/runs/22737527586/job/65943266000

## Root Cause

In `src/github/validation/permissions.ts`, the bot bypass only checks for `[bot]` suffix:

```typescript
// Line 47-50
if (actor.endsWith("[bot]")) {
  core.info(`Actor is a GitHub App: ${actor}`);
  return true;
}
```

`Copilot` is a non-user actor that doesn't follow the `[bot]` naming convention, so it falls through to the API call which 404s.

## Suggested Fix

Catch 404 errors from the collaborator permission API and check whether the actor is a known non-user entity, or fall back to checking `triggering_actor` permissions instead.

A simpler approach: wrap the API call in error handling that recognises "not a user" 404s and consults `allowed_bots` before throwing.

## Related Issues

- #1017 — Server-side OIDC token exchange also fails for Copilot actor (precedes this check)
[...]

Issue #903, opened 2026-02-05
Title: Support for GitHub Copilot-created Pull Requests


Krapp1:
Description

  When GitHub Copilot creates a pull request, the claude-code-action fails with a 404 error because it tries to look up the PR author as a regular
  GitHub user.

  Steps to Reproduce

  1. Use GitHub Copilot to create a pull request
  2. Have claude-code-action configured to run on pull_request events
  3. The action fails during permission checking

  Error Log

  Checking permissions for actor: Copilot
  GET /users/Copilot - 404 Not Found

  Expected Behavior

  The action should recognize that Copilot is a GitHub app/bot, not a regular user, and handle it appropriately (similar to how copilot[bot] or
  github-actions[bot] are handled).

  Workaround

  Currently skipping Copilot PRs with:
  if: github.event.pull_request.user.login != 'Copilot'

  Environment

  - Action version: anthropics/claude-code-action@v1
  - Configuration includes allowed_bots: 'copilot[bot],github-actions[bot]'

Issue #900, opened 2026-02-04
checkHumanActor fails with 404 for bot actors before checking allowed_bots configuration


julesmaurel:
## Describe the bug

The `checkHumanActor` function in agent mode fails with a 404 "Not Found" error for scheduled workflows before the `allowed_bots` configuration is checked. This makes the `allowed_bots` input ineffective for scheduled/cron-triggered workflows.

## Root Cause

In `src/github/validation/actor.ts`, the function makes an API call **before** checking the `allowed_bots` list:

```typescript
export async function checkHumanActor(octokit, githubContext) {
  // This API call fails with 404 for bot accounts
  const { data: userData } = await octokit.users.getByUsername({
    username: githubContext.actor, // e.g., "github-merge-queue[bot]"
  });

  const actorType = userData.type;

  // allowed_bots check happens AFTER the API call
  if (actorType !== "User") {
    const allowedBots = githubContext.inputs.allowedBots;
    // ... check logic never reached if API call fails
  }
}
```

For bot actors like `github-merge-queue[bot]`, `octokit.users.getByUsername()` returns 404 because bot accounts are not regular GitHub users accessible via the users API.

## To Reproduce

1. Create a scheduled GitHub Actions workflow using Claude Code action
2. Configure `allowed_bots: 'github-merge-queue'` (or any bot name)
3. Wait for the scheduled workflow to run
4. Observe 404 error in the "Run test generation skill" step:
   ```
   ##[error]Prepare step failed with error: Not Found - https://docs.github.com/rest
   ```

## Expected behavior

The `allowed_bots` check should happen **before** making the API call. If the actor matches the allowed bots pattern (including `*` for all bots), the function should return early without attempting the API call.

## Proposed Fix

```typescript
export async function checkHumanActor(octokit, githubContext) {
  const allowedBots = githubContext.inputs.allowedBots;

  // Check if all bots are allowed first (before any API call)
  if (allowedBots.trim() === "*") {
    console.log(`All bots are allowed, skipping human actor check`);
    return;
  }

  // Check if specific bot is in allowed list (before API call)
  const allowedBotsList = allowedBots
    .split(",")
    .map((bot) => bot.trim().toLowerCase().replace(/\[bot\]$/, ""))
    .filter((bot) => bot.length > 0);

  const actorName = githubContext.actor.toLowerCase().replace(/\[bot\]$/, "");
  
  if (allowedBotsList.includes(actorName)) {
[...]

PR #1144, opened 2026-04-02
fix: handle non-user actors (e.g. Copilot) in permission and actor checks

krislavten:
## Summary

GitHub Apps like **Copilot SWE Agent** set `GITHUB_ACTOR` to a value (e.g. `"Copilot"`) that is neither a valid GitHub user nor ends with `[bot]`. This causes two independent crashes in the action:

1. **`checkWritePermissions`** (`permissions.ts`): calls the collaborator permission API → 404 `"Copilot is not a user"`
2. **`checkHumanActor`** (`actor.ts`): calls `octokit.users.getByUsername()` → 404 *before* reaching the `allowed_bots` check

### Root Cause

Both functions assume the actor is either a real GitHub user or a `[bot]`-suffixed app. Some GitHub Apps (Copilot, and potentially future ones) use a plain name as `GITHUB_ACTOR` that doesn't match either pattern.

### Fix

**`permissions.ts`**:
- Check `allowed_bots` **before** calling the collaborator API, so known bots skip the API entirely
- Catch `"is not a user"` 404 errors and fall back to `allowed_bots` check instead of crashing

**`actor.ts`**:
- Move `allowed_bots` check **before** the `octokit.users.getByUsername()` call
- Catch 404 errors and provide a clear error message telling the user to add the bot to `allowed_bots`

### Tests

Added 11 new test cases covering:
- Non-`[bot]` actor in `allowed_bots` list (both pre-API and 404 fallback paths)
- Wildcard `*` for non-`[bot]` actors
- Non-`[bot]` actor NOT in `allowed_bots` (returns `false` / clear error message)
- Case-insensitive matching
- Non-404 errors still propagate correctly

All 664 existing tests continue to pass.

## Closes

- #900 — `checkHumanActor` fails with 404 before checking `allowed_bots`
- #903 — Support for GitHub Copilot-created Pull Requests
- #1018 — `checkWritePermissions` fails with 404 for non-user actors
- #1133 — `allowed_bots` bypasses actor check but not permission check

## Related

- Supersedes #1030 (which only fixed `permissions.ts` but not `actor.ts`)

### DIFF
--- a/src/github/validation/actor.ts
+++ b/src/github/validation/actor.ts
@@ -8,53 +8,75 @@
 import type { Octokit } from "@octokit/rest";
 import type { GitHubContext } from "../context";
 
+function isAllowedBot(actor: string, allowedBots: string): boolean {
+  const trimmed = allowedBots.trim();
+  if (trimmed === "*") return true;
+  if (!trimmed) return false;
+
+  const allowedList = trimmed
+    .split(",")
+    .map((bot) =>
+      bot
+        .trim()
+        .toLowerCase()
+        .replace(/\[bot\]$/, ""),
+    )
+    .filter((bot) => bot.length > 0);
+
+  const normalizedActor = actor.toLowerCase().replace(/\[bot\]$/, "");
+  return allowedList.includes(normalizedActor);
+}
+
 export async function checkHumanActor(
   octokit: Octokit,
   githubContext: GitHubContext,
 ) {
-  // Fetch user information from GitHub API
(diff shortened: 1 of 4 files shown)


fallemand:
We are facing this issue. Thanks!

ashwin-ant (approved):

ashwin-ant:
@krislavten mind running `bun run format` and pushing again?

crocodele:
Would be lovely to get these fixes merged. :pray:

PR #1161, opened 2026-04-04
fix: reject control characters in PATH_TO_CLAUDE_CODE_EXECUTABLE

### LINKED ISSUE #1160: security: PATH_TO_CLAUDE_CODE_EXECUTABLE newline injection into GITHUB_PATH
## Summary

`installClaudeCode()` in `src/entrypoints/run.ts` appends `PATH_TO_CLAUDE_CODE_EXECUTABLE`'s parent directory into `GITHUB_PATH` without sanitizing the value first:

```typescript
const customExecutable = process.env.PATH_TO_CLAUDE_CODE_EXECUTABLE;
if (customExecutable) {
  const claudeDir = dirname(customExecutable);
  if (githubPath) {
    await appendFile(githubPath, `${claudeDir}\n`);  // line 55
  }
```

`GITHUB_PATH` is a runner-managed file where each line is a directory to add to `PATH` for subsequent steps. If `PATH_TO_CLAUDE_CODE_EXECUTABLE` contains embedded newlines (e.g. `/usr/local/bin/claude\n/attacker/path`), `dirname()` passes through the embedded newline, and `appendFile` writes two lines — injecting an attacker-controlled directory into the runner's PATH.

## Impact

Any workflow that sets `PATH_TO_CLAUDE_CODE_EXECUTABLE` from an untrusted source (e.g. a workflow input, an environment variable populated from PR metadata) could have additional directories injected into the runner PATH for all subsequent steps in that job. This could be used to shadow system binaries.

In practice, `PATH_TO_CLAUDE_CODE_EXECUTABLE` is typically set by the workflow author, so the immediate risk is low. However, defense-in-depth sanitization is appropriate for anything written to runner environment files.

## Proposed fix

Strip control characters (including `\n`, `\r`, null bytes) from `claudeDir` before writing to `GITHUB_PATH`:

```typescript
const safePath = claudeDir.replace(/[\x00-\x1f\x7f]/g, "");
await appendFile(githubPath, `${safePath}\n`);
```

Or throw if the sanitized value differs from the original (fail-closed rather than silently stripping).

## References

- GitHub Actions security hardening: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions
- Related: `run.ts` lines 48-59

qozle:
## Problem

`PATH_TO_CLAUDE_CODE_EXECUTABLE` is passed through `dirname()` and written directly to the `GITHUB_PATH` file without sanitizing control characters:

```typescript
const claudeDir = dirname(customExecutable);
await appendFile(githubPath, `${claudeDir}\n`);
```

`GITHUB_PATH` is newline-delimited — each line becomes a directory added to `PATH` for subsequent steps. If `customExecutable` contains an embedded newline (e.g. set from an untrusted workflow input), `dirname()` passes it through and `appendFile` writes two lines, injecting an attacker-controlled directory into the runner PATH for all subsequent job steps.

## Fix

Add a fail-closed guard before the `appendFile` call that throws if `claudeDir` contains any control characters (`\x00–\x1f`, `\x7f`):

```typescript
if (/[\x00-\x1f\x7f]/.test(claudeDir)) {
  throw new Error(
    `PATH_TO_CLAUDE_CODE_EXECUTABLE contains invalid control characters: ${JSON.stringify(claudeDir)}`,
  );
}
```

**Fail-closed rather than silently stripping:** a path with embedded control characters is almost certainly a misconfiguration or injection attempt. Throwing loudly is better than quietly sanitizing and continuing — the latter hides the problem.

All 653 existing tests continue to pass.

Fixes #1160

### DIFF
--- a/src/entrypoints/run.ts
+++ b/src/entrypoints/run.ts
@@ -49,6 +49,13 @@ async function installClaudeCode(): Promise<void> {
   if (customExecutable) {
     console.log(`Using custom Claude Code executable: ${customExecutable}`);
     const claudeDir = dirname(customExecutable);
+    // Reject paths containing control characters — an embedded newline in claudeDir
+    // would inject an additional directory into GITHUB_PATH for all subsequent steps.
+    if (/[\x00-\x1f\x7f]/.test(claudeDir)) {
+      throw new Error(
+        `PATH_TO_CLAUDE_CODE_EXECUTABLE contains invalid control characters: ${JSON.stringify(claudeDir)}`,
+      );
+    }
     // Add to PATH by appending to GITHUB_PATH
     const githubPath = process.env.GITHUB_PATH;
     if (githubPath) {


qozle:
Closing — duplicate of #1185 which was already merged.

PR #1164, opened 2026-04-04
fix: bypass write permission check for actors in allowed_bots list

### LINKED ISSUE #1133: allowed_bots bypasses actor check but not permission check — bot actors still fail with 404

qozle:
## Problem

When a bot actor like `Copilot` (GitHub user type: `Bot`, but no `[bot]` suffix in its login) triggers a workflow with `allowed_bots: "Copilot"`, the action fails before the allowed list is ever consulted:

```
Checking permissions for actor: Copilot
GET /repos/{owner}/{repo}/collaborators/Copilot/permission — 404
##[error] Failed to check permissions: HttpError: Copilot is not a user
##[error] Action failed with error: Actor does not have write permissions to the repository
```

Every approach fails for the same reason — `checkWritePermissions` runs first and throws before `checkHumanActor` gets a chance to apply the `allowed_bots` allowlist.

## Root cause

`checkWritePermissions` only bypasses the collaborators API for actors whose login ends in `[bot]` (line 47). Bots without that suffix — `Copilot`, `renovate`, `dependabot` (when listed without suffix) — fall through to the collaborators API. That API returns 404 for accounts that are not repository collaborators, and the error propagates as a permission failure.

Meanwhile `checkHumanActor` (called later in `prepareAgentMode`/`prepareTagMode`) correctly handles these actors: it calls `octokit.users.getByUsername`, sees `type: "Bot"`, and checks the `allowed_bots` list. But execution never reaches it.

## Fix

Add an optional `allowedBots` parameter to `checkWritePermissions`. When the actor's normalized name (case-insensitive, `[bot]`-suffix-stripped) matches an entry in the list, the collaborators API call is skipped. This is the same trust model as the existing `[bot]`-suffix bypass — the repo owner explicitly configured the list.

```typescript
// src/github/validation/permissions.ts (new export)
export function isActorInAllowedBots(actor: string, allowedBots: string): boolean

// checkWritePermissions — new optional 5th parameter
export async function checkWritePermissions(
  octokit, context, allowedNonWriteUsers?, githubTokenProvided?, allowedBots?
)
```

`run.ts` passes `context.inputs.allowedBots` as the new argument.

Additional details:
- `allowed_bots: '*'` emits `⚠️ SECURITY WARNING` matching the `allowed_non_write_users='*'` precedent
- `isActorInAllowedBots` is exported as a standalone utility for independent testing
- 17 new tests added: specific match, wildcard, case-insensitive, `[bot]` normalization, and negative cases
- All 670 existing tests continue to pass

## After this fix

```yaml
# Both now work:
allowed_bots: "Copilot"           # exact match, case-insensitive
allowed_bots: "Copilot,renovate"  # multiple bots
allowed_bots: "*"                 # all bots (emits security warning)
```

Fixes #1133

### DIFF
--- a/src/entrypoints/run.ts
+++ b/src/entrypoints/run.ts
@@ -177,6 +177,7 @@ async function run() {
         context,
         context.inputs.allowedNonWriteUsers,
         !!process.env.OVERRIDE_GITHUB_TOKEN,
+        context.inputs.allowedBots,
       );
       if (!hasWritePermissions) {
         throw new Error(
--- a/src/github/validation/permissions.ts
+++ b/src/github/validation/permissions.ts
@@ -3,11 +3,40 @@ import type { ParsedGitHubContext } from "../context";
 import type { Octokit } from "@octokit/rest";
 
 /**
+ * Normalize a bot name for comparison: lowercase and strip the "[bot]" suffix.
+ * Matches the normalization used in checkHumanActor.
+ */
+function normalizeBotName(name: string): string {
+  return name.toLowerCase().replace(/\[bot\]$/, "");
+}
+
+/**
+ * Check if the actor is in the allowed_bots list.
+ * Handles wildcard "*" and comma-separated bot name lists.
+ * Names are normalized: case-insensitive and "[bot]" suffix is ignored.
+ */
+export function isActorInAllowedBots(
+  actor: string,
+  allowedBots: string,
+): boolean {
+  const trimmed = allowedBots.trim();
+  if (!trimmed) return false;
+  if (trimmed === "*") return true;
+  const actorName = normalizeBotName(actor);
+  const allowedList = trimmed
+    .split(",")
+    .map((b) => normalizeBotName(b.trim()))
+    .filter((b) => b.length > 0);
+  return allowedList.includes(actorName);
+}
+
+/**
  * Check if the actor has write permissions to the repository
  * @param octokit - The Octokit REST client
  * @param context - The GitHub context
  * @param allowedNonWriteUsers - Comma-separated list of users allowed without write permissions, or '*' for all
  * @param githubTokenProvided - Whether github_token was provided as input (not from app)
+ * @param allowedBots - Comma-separated list of bot usernames allowed to bypass the permission check, or '*' for all bots
  * @returns true if the actor has write permissions, false otherwise
  */
 export async function checkWritePermissions(
@@ -15,6 +44,7 @@ export async function checkWritePermissions(
   context: ParsedGitHubContext,
   allowedNonWriteUsers?: string,
   githubTokenProvided?: boolean,
+  allowedBots?: string,
 ): Promise<boolean> {
   const { repository, actor } = context;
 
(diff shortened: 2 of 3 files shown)


qozle:
Closing — whether `allowed_bots` should bypass the write permission check is a security design decision that should be made by the Anthropic team. The current behavior (two independent gates) may be intentional.

PR #1165, opened 2026-04-04
fix: surface pull-requests permission failure instead of silently exiting 0

### LINKED ISSUE #1121: claude-code-review workflow template generates insufficient permissions; action silently swallows write failure
## Description

Two related bugs that compound each other, resulting in the code review action running successfully but silently posting no output anywhere.

## Bug 1: Generated workflow template has wrong permissions

The workflow template generated by the `/install-github-app` command sets `pull-requests: read`:

```yaml
permissions:
  contents: read
  pull-requests: read
  issues: read
  id-token: write
```

Posting review comments requires `pull-requests: write`. The template should include write permission since that is the entire purpose of the code review workflow.

## Bug 2: Action silently swallows the permission error

When the action attempts to post a review comment and receives a 403 from the GitHub API due to insufficient permissions, it exits 0. The job reports success with no indication that the review was never written anywhere.

## Expected behaviour

- The generated template should include `pull-requests: write`
- The action should fail the step (non-zero exit) when it receives a permission error trying to write a review, so the misconfiguration is surfaced immediately

## Impact

Without either fix, there is no feedback loop: the template is wrong, the action doesn't complain, the job shows green, and reviews silently go nowhere.

qozle:
## Problem

Two compounding bugs that result in the code review action running successfully but silently posting no output anywhere — green check, zero comments, no indication anything went wrong.

## Bug 1: Example workflows have wrong permissions

`examples/pr-review-filtered-authors.yml` and `examples/pr-review-filtered-paths.yml` both specify:

```yaml
permissions:
  pull-requests: read   # ← wrong
```

Posting review comments requires `pull-requests: write`. Users copying these templates hit silent failures with no diagnostic.

## Bug 2: Action exits 0 when Claude can't post comments (403)

When Claude calls `create_inline_comment` and receives a 403, the MCP server already returns `{ isError: true }`. But Claude still concludes `subtype: "success"` because from its perspective it "attempted the task." The action trusts `subtype` as the sole success signal:

```typescript
const isSuccess = resultMessage.subtype === "success";
result.conclusion = isSuccess ? "success" : "failure";
```

Result: green check, zero review comments, no feedback.

## Fix

**Bug 1:** Change `pull-requests: read` → `write` in both example files.

**Bug 2 — two-part fix:**

*Part A* — add a 403-specific `helpMessage` to `github-inline-comment-server.ts` so Claude's response includes actionable guidance:

> This error means the workflow token lacks permission to post review comments. Add `pull-requests: write` to your workflow's permissions block.

*Part B* — after `runClaude()` returns "success", scan the execution file for `tool_result` items with `is_error: true` and the text `"Resource not accessible by integration"` (GitHub's canonical error for this permission class). If any are found, override the conclusion to failure with an actionable message:

```
Claude attempted to post 5 review comment(s) but received a GitHub permission error.
Add `pull-requests: write` to your workflow's permissions block:

  permissions:
    pull-requests: write
```

The string `"Resource not accessible by integration"` is unambiguous — it's the specific Octokit error for GitHub Apps integration permission failures, not a general 403. The scan is a no-op when the execution file is absent or contains no such errors.

Fixes #1121

### DIFF
--- a/examples/pr-review-filtered-authors.yml
+++ b/examples/pr-review-filtered-authors.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      pull-requests: read
+      pull-requests: write
       id-token: write
     steps:
       - name: Checkout repository
--- a/examples/pr-review-filtered-paths.yml
+++ b/examples/pr-review-filtered-paths.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      pull-requests: read
+      pull-requests: write
       id-token: write
     steps:
       - name: Checkout repository
--- a/src/entrypoints/run.ts
+++ b/src/entrypoints/run.ts
@@ -131,6 +131,56 @@ async function writeStepSummary(executionFile: string): Promise<void> {
   }
 }
 
+/**
+ * Scan the Claude execution file for inline comment tool calls that failed
+ * with a GitHub permission error. Returns an actionable failure message when
+ * found, or null when the file is absent or contains no such errors.
+ *
+ * "Resource not accessible by integration" is GitHub's canonical response
+ * when a GitHub App token lacks the required permission scope — in this case
+ * `pull-requests: write`.
+ */
+function detectCommentPermissionFailure(executionFile: string): string | null {
+  const PERMISSION_ERROR = "Resource not accessible by integration";
+  try {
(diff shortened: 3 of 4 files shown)


mathatomo57:
@ashwin-ant 
Hi @ashwin-ant, we ran into this exact issue — the review action completed successfully but posted nothing, and it took a while to figure out it was the read-only permission. This fix would really help. Any chance this could get a review?

PR #1173, opened 2026-04-05
fix: merge project permissions.allow into tag mode --allowedTools

### LINKED ISSUE #1063: Regression: Tag mode --allowedTools overrides project .claude/settings.json permissions.allow
## Summary

After PR #1002 ("Harden tag mode tool permissions against prompt injection", merged March 12, 2026), project-level `.claude/settings.json` `permissions.allow` entries are no longer respected in tag mode.

The hardcoded `--allowedTools` list in `src/modes/tag/index.ts` combined with headless `acceptEdits` mode means any tool not in that hardcoded list falls to "ask" → no prompt handler → denied. Project settings are loaded (`settingSources` confirms `["user", "project", "local"]`), but effectively ignored.

## Example

Given a project `.claude/settings.json`:

```json
{
  "permissions": {
    "allow": [
      "Bash(pnpm test:*)",
      "Bash(pnpm fix:*)",
      "Bash(node:*)"
    ]
  }
}
```

Any `@claude` task that requires running tests hits repeated permission denials, even though these tools are explicitly allowed in the project settings.

## Root Cause

PR #1002 changed `src/modes/tag/index.ts` to:

```typescript
claudeArgs += ` --permission-mode acceptEdits --allowedTools "${tagModeTools.join(",")}"`;
```

Where `tagModeTools` is a hardcoded list of read tools + git operations + GitHub MCP tools. In headless `acceptEdits` mode, anything not in `--allowedTools` is denied with no fallback.

## Previously Working

[Issue #189](https://github.com/anthropics/claude-code-action/issues/189) was closed in June 2025 with the reporter confirming that `.claude/settings.json` permissions were being reflected in the action without any special configuration. This confirms the behavior worked before the hardening changes.

## Docs Conflict

From [Claude Code Settings](https://code.claude.com/docs/en/settings):

> **Array settings merge across scopes.** When the same array-valued setting (such as `permissions.allow`) appears in multiple scopes, the arrays are **concatenated and deduplicated**, not replaced.

The current behavior contradicts this — CLI-level `--allowedTools` effectively replaces rather than merges with project settings.

## Related

- [Issue #695](https://github.com/anthropics/claude-code-action/issues/695) — Same pattern with MCP tools: tag mode filters `userAllowedMCPTools` to only `mcp__github_*`.

## Suggested Fix

The security hardening in PR #1002 is valid. But `tagModeTools` should merge project `permissions.allow` entries rather than ignoring them. `Bash(...)` allow rules from project settings don't bypass the `git-push.sh` wrapper or grant filesystem write access outside `$GITHUB_WORKSPACE`.

qozle:
Fixes #1063.

## Problem

PR #1002 (\"Harden tag mode tool permissions against prompt injection\") added an explicit `--allowedTools` list to tag mode. In `acceptEdits` mode with no prompt handler, anything not in that list hits \"ask\" → immediately denied. This regression broke project `.claude/settings.json` `permissions.allow` entries — tools like `Bash(pnpm test:*)` that maintainers explicitly approved are silently denied.

## Fix

`restoreConfigFromBase()` already runs in `run.ts` before the CLI starts, replacing `.claude/` with the trusted base-branch version. After that restore, read the project's trusted `permissions.allow` entries and append them to `claudeArgs` as an additional `--allowedTools` flag.

This restores the pre-#1002 behavior of respecting project settings while keeping the hardening intact — the entries come from the base branch (maintainer-reviewed), not the PR head.

## Security

No new trust surface. The project `permissions.allow` entries are read from the base-branch version of `.claude/settings.json` (already trusted by the existing restore logic), not from the PR-authored version.

### DIFF
--- a/src/entrypoints/run.ts
+++ b/src/entrypoints/run.ts
@@ -246,6 +246,34 @@ async function run() {
       }
       if (restoreBase) {
         restoreConfigFromBase(restoreBase);
+
+        // After restore, .claude/settings.json is the trusted base-branch version.
+        // For tag mode, merge its permissions.allow entries into claudeArgs so that
+        // project-approved tools (e.g. Bash(pnpm test:*)) are not silently denied.
+        // PR #1002 hardened tag mode with an explicit --allowedTools allowlist, which
+        // inadvertently stopped respecting project settings entirely.
+        if (modeName === "tag") {
+          try {
+            const settingsPath = ".claude/settings.json";
+            if (existsSync(settingsPath)) {
+              const settings = JSON.parse(readFileSync(settingsPath, "utf-8"));
+              const projectAllow: unknown = settings?.permissions?.allow;
+              if (Array.isArray(projectAllow) && projectAllow.length > 0) {
+                const projectTools = projectAllow.filter(
+                  (t): t is string => typeof t === "string" && t.length > 0,
+                );
+                if (projectTools.length > 0) {
+                  prepareResult.claudeArgs += ` --allowedTools "${projectTools.join(",")}"`;
+                  console.log(
+                    `Merged ${projectTools.length} project permission(s) from .claude/settings.json into tag mode tools`,
+                  );
+                }
+              }
+            }
+          } catch {
+            // Malformed settings.json — proceed with hardcoded tool list only.
+          }
+        }
       }
     }
 
--- a/src/entrypoints/run.ts
+++ b/src/entrypoints/run.ts
@@ -246,34 +246,38 @@ async function run() {
       }
       if (restoreBase) {
         restoreConfigFromBase(restoreBase);
+      }
+    }
 
-        // After restore, .claude/settings.json is the trusted base-branch version.
-        // For tag mode, merge its permissions.allow entries into claudeArgs so that
-        // project-approved tools (e.g. Bash(pnpm test:*)) are not silently denied.
-        // PR #1002 hardened tag mode with an explicit --allowedTools allowlist, which
-        // inadvertently stopped respecting project settings entirely.
-        if (modeName === "tag") {
-          try {
-            const settingsPath = ".claude/settings.json";
-            if (existsSync(settingsPath)) {
-              const settings = JSON.parse(readFileSync(settingsPath, "utf-8"));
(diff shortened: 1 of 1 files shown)


src/entrypoints/run.ts
```diff
@@ -246,6 +246,34 @@ async function run() {
       }
       if (restoreBase) {
         restoreConfigFromBase(restoreBase);
+
+        // After restore, .claude/settings.json is the trusted base-branch version.
+        // For tag mode, merge its permissions.allow entries into claudeArgs so that
+        // project-approved tools (e.g. Bash(pnpm test:*)) are not silently denied.
+        // PR #1002 hardened tag mode with an explicit --allowedTools allowlist, which
+        // inadvertently stopped respecting project settings entirely.
+        if (modeName === "tag") {
```
ashwin-ant: This only runs when `restoreBase` is set (PR-triggered tag mode). Issue-triggered tag mode also has the #1002 hardening but no restore, so project `permissions.allow` stays ignored there. The settings file is equally trusted in that case (default branch checkout). Suggest moving this block to just after the `if (restoreBase) { ... }` closes, still gated on `modeName === "tag"`.

src/entrypoints/run.ts
```diff
@@ -246,6 +246,34 @@ async function run() {
...
+            const settingsPath = ".claude/settings.json";
+            if (existsSync(settingsPath)) {
+              const settings = JSON.parse(readFileSync(settingsPath, "utf-8"));
+              const projectAllow: unknown = settings?.permissions?.allow;
+              if (Array.isArray(projectAllow) && projectAllow.length > 0) {
+                const projectTools = projectAllow.filter(
+                  (t): t is string => typeof t === "string" && t.length > 0,
+                );
+                if (projectTools.length > 0) {
+                  prepareResult.claudeArgs += ` --allowedTools "${projectTools.join(",")}"`;
```
ashwin-ant: String-concatenating into `claudeArgs` means this goes through shell-quote parsing in base-action. A tool entry containing `"` or `,` would break. Safer to thread the project tools through to `prepareTagMode`'s tool builder so they're added to the array before it's serialized. If that's too invasive for this PR, at minimum escape `"` in the joined string.

PR #1174, opened 2026-04-05
fix: pin bun runtime config and improve log hygiene

ashwin-ant:
- Pin `--config`, `--tsconfig-override`, and `--no-env-file` on the three `bun run` entrypoints so bun resolves runtime config from the action directory instead of the workspace
- Register the OIDC-exchanged app token with `core.setSecret`
- Omit `extraArgs` from the SDK options debug log alongside `env`
- Add `CLAUDE.md`, `CLAUDE.local.md`, and `.husky` to the config-restore list
- Prettier line-wrap in `base-action/test/parse-sdk-options.test.ts` (drift from #1055)

Tests, typecheck, and format all pass.

### DIFF
--- a/src/github/operations/restore-config.ts
+++ b/src/github/operations/restore-config.ts
@@ -15,6 +15,9 @@ const SENSITIVE_PATHS = [
   ".claude.json",
   ".gitmodules",
   ".ripgreprc",
+  "CLAUDE.md",
+  "CLAUDE.local.md",
+  ".husky",
 ];
 
 /**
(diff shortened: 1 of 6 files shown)


dicksontsai (approved):

src/github/operations/restore-config.ts
```diff
@@ -15,6 +15,9 @@ const SENSITIVE_PATHS = [
   ".claude.json",
   ".gitmodules",
   ".ripgreprc",
+  "CLAUDE.md",
+  "CLAUDE.local.md",
+  ".husky",
```
maxired, line 20: In our repo, we do commit the .husky files, but not the hook setup in themself.
with this setup, we need to rerun `npx husky` to reinstall the hooks, otherwise, our pre commit hook do not run.
this broke previous behavior

PR #1180, opened 2026-04-05
fix: allow track_progress with pull_request labeled action

### LINKED ISSUE #1095: track_progress should support pull_request 'labeled' action
## Problem

`track_progress: true` fails when the workflow is triggered by a `pull_request: [labeled]` event:

```
Error: Action failed with error: track_progress for pull_request events is only supported for actions: opened, synchronize, ready_for_review, reopened. Current action: labeled
```

## Use case

Using labels as a manual trigger for Claude reviews is a valid and common pattern. Instead of running the review on every PR event (`opened`, `synchronize`), the user adds a label (e.g., `claude`) to explicitly request a review when they want one. This avoids unnecessary API calls and gives the user control over when reviews happen.

Example workflow:

```yaml
on:
  pull_request:
    types: [labeled]

jobs:
  claude-review:
    if: ${{ github.event.label.name == 'claude' }}
    steps:
        with:
          track_progress: true
          # ...
```

## Why it should work

The `labeled` event payload contains the same `pull_request.number` and commit SHA as `opened` or `synchronize`. There's no technical reason `track_progress` can't function — it has all the information it needs to post progress updates on the PR.

## Current workaround

Setting `track_progress: false`, which loses the progress indicator.

## Expected behavior

`track_progress: true` should work with `pull_request` events of type `labeled` (and ideally `unlabeled` too, for completeness).

qozle:
## Problem

`track_progress: true` throws when the workflow fires on `pull_request: [labeled]`:

```
Error: track_progress for pull_request events is only supported for actions:
opened, synchronize, ready_for_review, reopened. Current action: labeled
```

Using labels as an explicit review trigger is a common pattern — users add a label to request a review rather than running on every push. The `labeled` event carries the same PR number and commit SHA as `opened`/`synchronize`, so `track_progress` has everything it needs.

## Changes

`src/modes/detector.ts`:
- Add `"labeled"` to `validActions` in `validateTrackProgressEvent` (removes the throw)
- Add `"labeled"` to `supportedActions` in `detectMode` (ensures tag mode is selected when `track_progress` + `prompt` are set)

`test/modes/detector.test.ts`:
- Add test: `track_progress: true` with `pull_request.labeled` returns `"tag"`

Fixes #1095

### DIFF
--- a/src/modes/detector.ts
+++ b/src/modes/detector.ts
@@ -67,6 +67,7 @@ export function detectMode(context: GitHubContext): AutoDetectedMode {
       "synchronize",
       "ready_for_review",
       "reopened",
+      "labeled",
     ];
     if (context.eventAction && supportedActions.includes(context.eventAction)) {
       // If prompt is provided, use agent mode (default for automation)
@@ -103,6 +104,7 @@ function validateTrackProgressEvent(context: GitHubContext): void {
       "synchronize",
       "ready_for_review",
       "reopened",
+      "labeled",
     ];
     if (!validActions.includes(context.eventAction)) {
       throw new Error(
--- a/test/modes/detector.test.ts
+++ b/test/modes/detector.test.ts
@@ -76,6 +76,20 @@ describe("detectMode with enhanced routing", () => {
       expect(detectMode(context)).toBe("agent");
     });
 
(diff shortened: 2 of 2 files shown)


qozle:
Closing — fix was too narrow (extends the allowlist rather than removing it). Will resubmit with the correct approach.

PR #1189, opened 2026-04-06
feat: expose token usage as step outputs

### LINKED ISSUE #59: Cost report
Is there a way to make claude report the cost after run?

### LINKED ISSUE #136: Mechanism for setting a cap on cost
I want to say "@claude spend up to $5 trying to do X"

adamhenson:
Closes #59, closes #136.

The action tracks token usage internally but never surfaces it as step outputs, so workflows can't gate on cost, log spend, or forward counts to external tools. This PR wires it up.

## Changes

**`base-action/src/run-claude-sdk.ts`** — Extended `ClaudeRunResult` with 5 new optional fields. Added accumulators in the SDK message loop summing `message.usage` from each assistant message, and set them on the result after the loop alongside `num_turns` from `resultMessage`.

**`src/entrypoints/run.ts`** — 5 new `core.setOutput()` calls after conclusion, using `?? ""` as fallback (same pattern as the rest of the codebase).

**`action.yml`** — 5 new entries in the `outputs:` section.

**`docs/usage.md`** — New `## Tracking Token Usage and Cost` section inserted between Structured Outputs and Ways to Tag @claude, with sub-sections covering logging, budget enforcement, and forwarding to a cost dashboard.

## Usage after this PR

```yaml
  id: claude
  with:
    prompt: "..."

- name: Log token usage
  run: |
    echo "Input tokens:  ${{ steps.claude.outputs.input_tokens }}"
    echo "Output tokens: ${{ steps.claude.outputs.output_tokens }}"
    echo "Turns:         ${{ steps.claude.outputs.turns }}"
```

Budget enforcement is now possible via a simple workflow step checking the outputs — partially addressing #136.


base-action/src/run-claude-sdk.ts
```diff
@@ -156,6 +162,10 @@ export async function runClaudeWithSdk(
 
   const messages: SDKMessage[] = [];
   let resultMessage: SDKResultMessage | undefined;
+  let totalInputTokens = 0;
```
andrevandal, line 164: what's your opinion about having just one const object and updating its props?

I feel it seems more structured but your approach is already fine

base-action/src/run-claude-sdk.ts
```diff
@@ -169,6 +179,14 @@ export async function runClaudeWithSdk(
       if (message.type === "result") {
         resultMessage = message as SDKResultMessage;
       }
+
+      if (message.type === "assistant") {
+        const usage = (message as SDKAssistantMessage).message.usage;
+        totalInputTokens += usage.input_tokens || 0;
```
andrevandal: I would prefer `??` instead of `||`. It feels safer.

base-action/src/run-claude-sdk.ts
```diff
@@ -169,6 +179,14 @@ export async function runClaudeWithSdk(
       if (message.type === "result") {
         resultMessage = message as SDKResultMessage;
       }
+
+      if (message.type === "assistant") {
+        const usage = (message as SDKAssistantMessage).message.usage;
```
andrevandal, line 193: this casting is used above this line, maybe moving out and using the same casting for both branches seems cleaner.

base-action/src/run-claude-sdk.ts
andrevandal: I'm suggesting something like

```
const budget = {
    tokens: {
      input: 0,
      output: 0,
    },
    cache: {
      read: 0,
      write: 0,
  }
} satisfies TYPE
```

src/entrypoints/run.ts
```diff
@@ -293,6 +293,24 @@ async function run() {
       core.setOutput("structured_output", claudeResult.structuredOutput);
     }
     core.setOutput("conclusion", claudeResult.conclusion);
+    if (claudeResult.inputTokens !== undefined) {
+      core.setOutput("input_tokens", String(claudeResult.inputTokens));
```
andrevandal, line 310: do we really need to transform intro string?

andrevandal (approved):
lgtm, thank you!

base-action/src/run-claude-sdk.ts
adamhenson: Good catch. Fixed in the latest commit -- changed all four to `??`. You're right that `||` would incorrectly treat `0` as falsy, and the cache fields are typed as `number | null` in `BetaUsage`, so `??` is the correct operator here.

src/entrypoints/run.ts
adamhenson: Yes - `core.setOutput` signature is `(name: string, value: string)` and the fields on `ClaudeRunResult` are `number | undefined`, so TypeScript requires the conversion. Passing the number directly would be a type error.

base-action/src/run-claude-sdk.ts
adamhenson: The two casts are for different types in different branches of the same discriminated union -- `SDKResultMessage` when `message.type === "result"` and `SDKAssistantMessage` when `message.type === "assistant"`. Moving them outside would mean casting `message` to two different types before the branches even run, which doesn't model the intent. Keeping each cast inside its own type-narrowed branch is the standard TypeScript pattern here.

base-action/src/run-claude-sdk.ts
adamhenson: Appreciate the suggestion! The flat variables feel more conventional here -- the rest of the codebase uses the same pattern (e.g. `resultMessage`, `commentId`, `claudeBranch` are all individual `let` declarations rather than grouped objects). A nested object would also require introducing a new type just for internal accumulation, without adding much clarity given these four values flow into separate fields on `ClaudeRunResult` right after.

adamhenson:
Gentle ping - PR has been open about 3 weeks with a community approval from @andrevandal. Happy to rebase, adjust the implementation, or address any feedback.

adamhenson:
Happy to close this if it does not fit the roadmap — just let me know either way.

albsala-backmarket:
Thank you for doing this @adamhenson ❤️ 

Any news on this topic?  It would be definitely useful to track this data.

cc: @ashwin-ant ?

docs/usage.md
```diff
@@ -257,6 +257,85 @@ See `examples/test-failure-analysis.yml` for a working example that:
...
+  with:
+    api_key: ${{ secrets.AGENTMETER_API_KEY }}
+    model: claude-sonnet-4-5
+    input_tokens: ${{ steps.claude.outputs.input_tokens }}
+    output_tokens: ${{ steps.claude.outputs.output_tokens }}
+    cache_read_tokens: ${{ steps.claude.outputs.cache_read_tokens }}
+    cache_write_tokens: ${{ steps.claude.outputs.cache_write_tokens }}
+    turns: ${{ steps.claude.outputs.turns }}
+    status: ${{ job.status }}
+```
```
ashwin-ant: The feature itself seems okay but please remove this.

nestorsgarzonc: @ashwin-ant I'm building a feature and in the pipeline I need to use the input_tokens and output_tokens to my metrics

nestorsgarzonc (approved):

docs/usage.md
adamhenson: @ashwin-ant Done. Removed the third-party service example and kept the logging and budget enforcement sections.

PR #1204, opened 2026-04-10
fix: remove .husky from SENSITIVE_PATHS

### LINKED ISSUE #1203: Adding .husky to SENSITIVE_PATHS breaks repos that commit .husky files but not the hook setup
## Problem

PR #1174 added `.husky` to the `SENSITIVE_PATHS` list in `src/github/operations/restore-config.ts`. This causes the `.husky` directory to be restored/reset during the action, which breaks repositories that commit their `.husky` files but rely on `npx husky` being run to install the hooks.

After this change, pre-commit hooks no longer run because the hook setup is wiped and would need to be reinstalled via `npx husky`.

This is a regression from the previous behavior where `.husky` was not treated as a sensitive path.

## Context

See the discussion on #1174: https://github.com/anthropics/claude-code-action/pull/1174#discussion_r3058932921

## Expected Behavior

Repositories that commit `.husky` directory contents should continue to have their git hooks work as expected after the action runs.

## Suggested Fix

Either:
- Remove `.husky` from `SENSITIVE_PATHS`, or
- Only restore specific files within `.husky` that are security-sensitive (e.g., prevent the action from *adding* new hooks), rather than blanket-restoring the entire directory
- ask the model tu run `npx husky` or similar to re install correctly the hooks

ashwin-ant:
Removes `.husky` from the `SENSITIVE_PATHS` list in `src/github/operations/restore-config.ts`. This fixes a regression from #1174 where repos that commit `.husky` files would have their git hook setup broken because the directory was being restored/reset.

Fixes #1203

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/github/operations/restore-config.ts
+++ b/src/github/operations/restore-config.ts
@@ -17,7 +17,6 @@ const SENSITIVE_PATHS = [
   ".ripgreprc",
   "CLAUDE.md",
   "CLAUDE.local.md",
-  ".husky",
 ];
 
 /**


maxired:
Hey, I am not sure to undersant the status here. Are we saying there is no way to have a safe behavior but working correctly ?

PR #1239, opened 2026-04-20
fix: resolve tsconfig.json via realpath to avoid bun fd mismatch

### LINKED ISSUE #1234: Bun: Internal error 'directory mismatch for directory' in v1 action on GitHub-hosted runners
## Summary

Every invocation of `anthropics/claude-code-action@v1` from our repo's `review` workflow fails with a Bun internal error during action setup:

```
Internal error: directory mismatch for directory "/home/runner/work/_actions/anthropics/claude-code-action/v1/tsconfig.json", fd 4
```

The error comes from Bun (the runtime the action uses); no user code runs. It reproduces on every PR in our repo.

## Reproduction

A reproducing run from PR #10526 in our repo (trajectory-labs-pbc/agent-c):
https://github.com/trajectory-labs-pbc/agent-c/actions/runs/24585295927/job/71893230564

The failure occurs in the `review` job, during action setup, before any user inputs are consumed.

## Impact

`review` jobs fail on every PR, blocking green CI for repos that require all checks to pass.

## Environment

- Runner: `ubuntu-latest` (GitHub-hosted)
- Action ref: `anthropics/claude-code-action@v1`
- Our repo: public, standard PR-triggered workflow

## Ask

- Pin Bun to a known-good version inside the action, or
- Surface a more actionable error if this is a repo- or path-specific issue, or
- Advise on a workaround we can apply from the caller side.

Happy to provide more logs on request.

richkuo:
## Summary

Fixes the Bun `Internal error: directory mismatch for directory ".../tsconfig.json", fd N` reported in #1234 by resolving the tsconfig path with `realpath` before passing it to `--tsconfig-override`.

## Root cause

GitHub-hosted runners place actions at paths like `/home/runner/work/_actions/<owner>/<action>/<ref>/`, where `<ref>` (e.g. `v1`) is typically a symlink to the actual SHA-named directory. When `${GITHUB_ACTION_PATH}/tsconfig.json` is passed to bun as-is, bun opens the file via the symlinked path but stores fd metadata keyed by the canonicalized path. On subsequent lookups bun sees a path mismatch and logs the internal error.

Pre-resolving with `realpath` gives bun a canonical path up front, so the fd's stored path and the lookup path agree.

## Changes

- `action.yml`: wrap each of the 3 occurrences of `--tsconfig-override="${GITHUB_ACTION_PATH}/tsconfig.json"` with `$(realpath ...)`.

`realpath` is part of coreutils and is always available on `ubuntu-latest` runners.

## Test plan

- [ ] Trigger `@claude` on a PR with the patched action ref and confirm the bun internal error no longer appears in logs.
- [ ] Verify the action still functions correctly (tsconfig resolution still works for bun's TypeScript execution).

Closes #1234

---
Generated with: Claude Opus 4.7 (1M) | Effort: medium

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -243,7 +243,7 @@ runs:
       run: |
         bun --no-env-file \
           --config="${GITHUB_ACTION_PATH}/bunfig.toml" \
-          --tsconfig-override="${GITHUB_ACTION_PATH}/tsconfig.json" \
+          --tsconfig-override="$(realpath "${GITHUB_ACTION_PATH}/tsconfig.json")" \
           run ${GITHUB_ACTION_PATH}/src/entrypoints/run.ts
       env:
         # Prepare inputs
@@ -380,7 +380,7 @@ runs:
         [ -x "$BUN_BIN" ] || BUN_BIN="bun"
         "$BUN_BIN" --no-env-file \
           --config="${GITHUB_ACTION_PATH}/bunfig.toml" \
-          --tsconfig-override="${GITHUB_ACTION_PATH}/tsconfig.json" \
+          --tsconfig-override="$(realpath "${GITHUB_ACTION_PATH}/tsconfig.json")" \
           run ${GITHUB_ACTION_PATH}/src/entrypoints/cleanup-ssh-signing.ts
 
     - name: Post buffered inline comments
@@ -397,7 +397,7 @@ runs:
         [ -x "$BUN_BIN" ] || BUN_BIN="bun"
         "$BUN_BIN" --no-env-file \
           --config="${GITHUB_ACTION_PATH}/bunfig.toml" \
-          --tsconfig-override="${GITHUB_ACTION_PATH}/tsconfig.json" \
+          --tsconfig-override="$(realpath "${GITHUB_ACTION_PATH}/tsconfig.json")" \
           run ${GITHUB_ACTION_PATH}/src/entrypoints/post-buffered-inline-comments.ts
 
     - name: Revoke app token
--- a/action.yml
+++ b/action.yml
@@ -241,10 +241,11 @@ runs:
       id: run
       shell: bash
       run: |
+        ACTION_DIR=$(realpath "${GITHUB_ACTION_PATH}")
         bun --no-env-file \
-          --config="${GITHUB_ACTION_PATH}/bunfig.toml" \
-          --tsconfig-override="$(realpath "${GITHUB_ACTION_PATH}/tsconfig.json")" \
-          run ${GITHUB_ACTION_PATH}/src/entrypoints/run.ts
+          --config="${ACTION_DIR}/bunfig.toml" \
+          --tsconfig-override="${ACTION_DIR}/tsconfig.json" \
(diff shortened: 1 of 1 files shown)


richkuo:
@ashwin-ant hi, can you review?

ashwin-ant:
Thanks for the PR. The change itself is safe and I'm fine taking it, but I want to flag a couple of things on the diagnosis.

The `_actions/.../v1` path isn't actually a symlink on GitHub-hosted runners. The runner extracts the tarball directly into a directory named after the ref. The `directory mismatch` message is a known bun bug in `--tsconfig-override` handling (see [oven-sh/bun#26793](https://github.com/oven-sh/bun/issues/26793)) and per that thread it's usually non-fatal log noise rather than a crash. We added `--tsconfig-override` in #1174, which lines up timing-wise, so the flag itself is the trigger.

Given that, I'd like to confirm this actually turns a failing run green rather than just suppressing the log line. Could you get a repro run against this branch before we merge?

One small ask while you're in here. Mind resolving the path once per step and reusing it for all three args? Keeps things consistent if bun grows similar quirks for `--config` or the script path:

```bash
ACTION_DIR=$(realpath "$GITHUB_ACTION_PATH")
bun --no-env-file \
  --config="$ACTION_DIR/bunfig.toml" \
  --tsconfig-override="$ACTION_DIR/tsconfig.json" \
  run "$ACTION_DIR/src/entrypoints/run.ts"
```

richkuo:
You're right on both counts — I re-ran this end-to-end and the "fix" is a no-op.

I exercised the branch via a one-shot `workflow_dispatch` workflow in a test repo (commenting `@claude` doesn't help since `issue_comment` always uses the default-branch workflow). The action executed from `richkuo/claude-code-action@fix/bun-tsconfig-realpath` with the `ACTION_DIR=$(realpath ...)` refactor applied. Logs:

```
ACTION_DIR=$(realpath "${GITHUB_ACTION_PATH}")
  --tsconfig-override="${ACTION_DIR}/tsconfig.json" \
  run "${ACTION_DIR}/src/entrypoints/run.ts"
...
Internal error: directory mismatch for directory
"/home/runner/work/_actions/richkuo/claude-code-action/fix/bun-tsconfig-realpath/tsconfig.json",
fd 4. You don't need to do anything, but this indicates a bug.
```

Two things fall out of that:

1. **The path isn't a symlink.** `realpath` returns the same string the runner already had, so `--tsconfig-override` sees the identical path before and after the fix.
2. **The fd-mismatch message is emitted regardless**, and the run was green — consistent with your read of [oven-sh/bun#26793](https://github.com/oven-sh/bun/issues/26793) as non-fatal noise.

So the original diagnosis on #1234 is wrong. There's nothing for `realpath` to canonicalize here, and this PR isn't changing behavior. Closing it. If the log noise is worth addressing, the right fixes are upstream (wait on the bun issue) or dropping `--tsconfig-override` in favor of a tsconfig at the action root — happy to open a separate PR for the latter if you want to explore it.

Thanks for pushing back on the diagnosis rather than just merging.

PR #1248, opened 2026-04-23
fix: allow + in branch names (generated by Claude Code EnterWorktree)

### LINKED ISSUE #1244: validateBranchName rejects '+' which EnterWorktree generates by converting '/' in worktree names
## Summary

`validateBranchName` in `src/github/operations/branch.ts` rejects branch names containing `+`, but Claude Code's `EnterWorktree` tool automatically converts `/` in worktree names to `+` when generating branch names. This creates a compatibility issue where branches created via the standard Claude Code workflow fail CI when using `claude-code-action`.

## Steps to Reproduce

1. Use Claude Code's `EnterWorktree` tool with a name containing `/`, e.g. `feat/skill-consolidation`
2. `EnterWorktree` generates the branch name `worktree-feat+skill-consolidation` (converts `/` → `+`)
3. Open a PR from that branch
4. `claude-code-action` fails with:
   ```
   Invalid branch name: "worktree-feat+skill-consolidation". Branch names must start with an
   alphanumeric character and contain only alphanumeric characters, forward slashes, hyphens,
   underscores, periods, or hashes (#).
   ```

## Root Cause

The whitelist pattern in `validateBranchName`:

```typescript
const validPattern = /^[a-zA-Z0-9][a-zA-Z0-9/_.#-]*$/;
```

does not include `+`, so branch names generated by `EnterWorktree` are rejected.

## Why This Should Be Fixed

As noted in the code comments, all git calls use `execFileSync` (not shell interpolation), so `+` carries no command injection risk — the same argument made for allowing `#`:

> All git calls use `execFileSync` (not shell interpolation), so `#` carries no injection risk.

The same logic applies to `+`. Git itself allows `+` in branch names, and since the execution is sandboxed via `execFileSync`, there is no security concern.

## Proposed Fix

Add `+` to the whitelist pattern:

```typescript
const validPattern = /^[a-zA-Z0-9][a-zA-Z0-9/_.#+\-]*$/;
```

Alternatively, fix `EnterWorktree` to use `-` instead of `+` when sanitizing worktree names — but since both tools are Anthropic products, fixing the validation in `claude-code-action` is the cleaner solution, as `+` is a valid and common character in branch names.

## Environment

- `claude-code-action`: v1 (latest)
- Claude Code CLI with `EnterWorktree` tool

awakia:
## Problem

\`validateBranchName\` rejects \`+\`, causing the action to fail on PRs from branches generated by Claude Code's \`EnterWorktree\` tool:

\`\`\`
##[error] Invalid branch name: "worktree-feat+skill-consolidation". Branch names must start with an
alphanumeric character and contain only alphanumeric characters, forward slashes, hyphens,
underscores, periods, or hashes (#).
\`\`\`

There is no workaround from the user side — the branch name is generated automatically by \`EnterWorktree\`, which converts \`/\` to \`+\` in worktree names (e.g. \`EnterWorktree("feat/skill-consolidation")\` creates branch \`worktree-feat+skill-consolidation\`). Fixes #1244.

## Root cause

The whitelist regex \`/^[a-zA-Z0-9][a-zA-Z0-9/_.#-]*$/\` excludes \`+\`. The validation exists to prevent command injection, but every git call in the action uses \`execFileSync\`, which bypasses the shell entirely (arguments go directly to the kernel via \`execve\`). There is no shell to interpret \`+\`, so the strict whitelist was blocking valid names with no security benefit.

\`+\` is explicitly permitted by \`git-check-ref-format\`. This is the same reasoning that was used to add \`#\` support in #1167.

## Fix

Add \`+\` to the whitelist: \`/^[a-zA-Z0-9][a-zA-Z0-9/_.#+-]*$/\`. Update the comment and error message to match.

## Testing

- 3 new test cases covering \`+\` in branch names generated by \`EnterWorktree\`
- All existing rejection cases still pass
- \`bun test\`: 665 pass, 0 fail


awakia:
Hey @ashwin-ant — this is essentially the same fix as #1167 (which you reviewed and merged), but for `+` instead of `#`. Claude Code's `EnterWorktree` tool generates branch names with `+` by converting `/` in worktree names, so users hitting this have no workaround. Would you be able to take a look?

ashwin-ant (approved):

awakia:
@ashwin-ant Thank you!!!

PR #1255, opened 2026-04-25
fix: write execution file when SDK throws

### LINKED ISSUE #1226: execution_file output not written when SDK throws on error_max_turns
## Bug Description

When Claude Code reaches the maximum number of turns (`error_max_turns`), the `execution_file` output is not written to disk. The execution data (messages, cost, turn count, duration) is accumulated in memory and the result object is printed to stdout, but the file at `$RUNNER_TEMP/claude-execution-output.json` is never created.

## Root Cause

In `base-action/src/run-claude-sdk.ts`, the file write happens **after** the `for await` loop's try/catch block:

```typescript
try {
  for await (const message of query(...)) {
    messages.push(message);
    // ...
  }
} catch (error) {
  console.error("SDK execution error:", error);
  throw new Error(`SDK execution error: ${error}`);
}

// This code is never reached when the SDK throws
try {
  await writeFile(EXECUTION_FILE, JSON.stringify(messages, null, 2));
  result.executionFile = EXECUTION_FILE;
} catch (error) {
  core.warning(`Failed to write execution file: ${error}`);
}
```

When the SDK throws (e.g., `Claude Code returned an error result: Reached maximum number of turns`), the catch block re-throws before the `writeFile` call is reached.

## Impact

- The `execution_file` output is empty, so downstream steps that depend on it (e.g., extracting cost, turn count, duration, tool error counts) get no data
- The result object **is** printed to stdout (with `num_turns`, `total_cost_usd`, `duration_ms`, `subtype: "error_max_turns"`), but the file is absent — consumers can't access the structured data
- This affects any workflow that uses `steps.<id>.outputs.execution_file` for post-run metrics or diagnostics

## Expected Behavior

The execution file should be written regardless of whether the SDK throws. The `messages` array is already populated with all conversation turns up to the point of failure — writing it to disk before re-throwing preserves diagnostic data.

## Suggested Fix

Move the `writeFile` call into a `finally` block, or write the file in the `catch` block before re-throwing:

```typescript
} catch (error) {
  console.error("SDK execution error:", error);
  // Write partial execution data before re-throwing
  try {
    await writeFile(EXECUTION_FILE, JSON.stringify(messages, null, 2));
    result.executionFile = EXECUTION_FILE;
  } catch (writeError) {
    core.warning(`Failed to write execution file: ${writeError}`);
  }
  throw new Error(`SDK execution error: ${error}`);
}
```

## Environment

- `claude-code-action@v1` (SHA: `905d4eb99ab3d43143d74fb0dcae537f29ac330a`)
- Bedrock mode with `--max-turns` set

Jerry2003826:
Fixes #1226

## Summary
- Persist the SDK message log before rethrowing SDK execution errors.
- Reuse the same execution-file writer on success and error paths.
- Add regression coverage for SDK errors after partial output.

## Root Cause
`runClaudeWithSdk` only wrote `EXECUTION_FILE` after the SDK async iterator completed. When `query()` threw, for example with `error_max_turns`, the catch path rethrew before writing the partial message log.

## Validation
- `bun test test/run-claude-sdk.test.ts`
- `bun test`
- `bun run typecheck`
- `bunx prettier --check src/run-claude-sdk.ts test/run-claude-sdk.test.ts`

Note: full `bun run format:check` on this Windows checkout reports pre-existing formatting/line-ending issues across unrelated files, so I verified the changed files with Prettier directly.

### DIFF
--- a/base-action/src/execution-file.ts
+++ b/base-action/src/execution-file.ts
@@ -1,0 +1,42 @@
+import * as core from "@actions/core";
+import { existsSync } from "fs";
+import { writeFile } from "fs/promises";
+import { join } from "path";
+
+const EXECUTION_FILENAME = "claude-execution-output.json";
+
+export function getExecutionFilePath(): string | undefined {
+  if (!process.env.RUNNER_TEMP) {
+    return undefined;
+  }
+  return join(process.env.RUNNER_TEMP, EXECUTION_FILENAME);
+}
+
+export async function writeExecutionFile(
+  messages: unknown[],
+): Promise<string | undefined> {
+  const executionFile = getExecutionFilePath();
+  if (!executionFile) {
+    core.warning("Failed to write execution file: RUNNER_TEMP is not set");
+    return undefined;
+  }
+
+  try {
+    await writeFile(executionFile, JSON.stringify(messages, null, 2));
+    console.log(`Log saved to ${executionFile}`);
+    return executionFile;
+  } catch (error) {
+    core.warning(`Failed to write execution file: ${error}`);
+    return undefined;
+  }
+}
+
+export function setExecutionFileOutputIfPresent(): string | undefined {
+  const executionFile = getExecutionFilePath();
+  if (!executionFile || !existsSync(executionFile)) {
+    return undefined;
+  }
+
+  core.setOutput("execution_file", executionFile);
+  return executionFile;
+}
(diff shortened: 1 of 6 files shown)


dhendry:
Woohoo, https://github.com/anthropics/claude-code-action/issues/1226 has been a real pain for me. Excited to see this getting fixed. Thanks @Jerry2003826 !

ashwin-ant (approved):

PR #1259, opened 2026-04-26
fix(cache): disable setup-bun cache to avoid 5-retry HTML-error burn

### LINKED ISSUE #1252: Cache save fails with 5-retry HTML-error pattern on successive runs against the same PR ref (key+ref collision)
**Describe the bug**
anthropics/claude-code-action@v1's post-job cache save fails consistently on successive runs against the same PR ref, burning ~20-30 seconds per failure on 5 retries of what  
is actually a permanent conflict (the GitHub cache API's "key already exists on this ref" response). The underlying @actions/cache save logic treats the HTML-formatted error
as transient and retries with exponential backoff when it should recognize the conflict as non-retryable and skip.

**To Reproduce**
1. Use anthropics/claude-code-action@v1 in a workflow triggered on pull_request with synchronize in the event list (default pattern).
  2. Open a PR that touches files the workflow processes. First run completes and successfully saves a cache entry (the 35 MB Bun binary tarball) against refs/pull/<N>/merge
  with a deterministic key like bun-swrvKXH4hkDYFKcrNV+Kct676IY=.
  3. Push a second commit to the same PR. A new workflow run starts.
  4. The post-step cache save fails with the 5-retry HTML error pattern:                                                                                                         
                                                                                                                                                                                 
  Post job cleanup.                                                                                                                                                              
  /usr/bin/tar --posix -cf cache.tzst --exclude cache.tzst -P -C /home/runner/work/<repo>/<repo> --files-from manifest.txt --use-compress-program zstdmt                         
  Sent 35242126 of 35242126 (100.0%), 58.0 MBs/sec                                                                                                                               
  Attempt 1 of 5 failed with error: Unexpected token '<', "<!DOCTYPE "... is not valid JSON. Retrying request in 3000 ms...                                                      
  Attempt 2 of 5 failed with error: Unexpected token '<', "<!DOCTYPE "... is not valid JSON. Retrying request in 5276 ms...                                                      
  Attempt 3 of 5 failed with error: Unexpected token '<', "<!DOCTYPE "... is not valid JSON. Retrying request in 9718 ms...                                                      
  Attempt 4 of 5 failed with error: Unexpected token '<', "<!DOCTYPE "... is not valid JSON. Retrying request in 10670 ms...                                                     
  Warning: Failed to save: Failed to FinalizeCacheEntryUpload: Failed to make request after 5 attempts: Unexpected token '<', "<!DOCTYPE "... is not valid JSON

**Expected behavior**
One of:                                                                                                                                                                        
                                                                                                                                                                                 
  1. The cache save recognizes that the key already exists on the ref, logs something like "cache already exists for key+ref, skipping save", and exits cleanly without retrying.
  2. The action uses a run-unique cache key with a restoreKeys prefix fallback so each run successfully saves its own entry (standard @actions/cache pattern for content that
  doesn't change across runs).                
  3. At minimum, the 409/HTML response is treated as a non-retryable error so the 5-retry wallclock burn doesn't happen.                                                         
                                              
  Either way, the warning should not appear on every second-and-subsequent push to a PR for a repo running the action as recommended.

**Screenshots**
 N/A. Log output and cache-entry inspection in Additional context.  

**Workflow yml file**
name: Adversarial Security Review
                                                                                                                                                                                 
  on:                                                                 
    pull_request:
      types: [opened, synchronize, reopened]  
      paths-ignore:                       
        - 'docs/**'
        - 'tasks/**'                                                                                                                                                             
        - 'PROJECT-PLAN.md'                   
                                                                                                                                                                                 
  permissions: {}                                                     
                                                                                                                                                                                 
  jobs:
    adversarial-review:                                                                                                                                                          
      if: >                                                           
        !github.event.pull_request.draft &&                                                                                                                                      
        github.event.pull_request.user.login != 'dependabot[bot]'     
      runs-on: ubuntu-latest                  
      permissions:                        
        contents: read                                                                                                                                                           
        pull-requests: write                                                                                                                                                     
        issues: read                                                                                                                                                             
        id-token: write                                                                                                                                                          
                                                                      
      steps:
        - name: Checkout repository                                                                                                                                              
          uses: actions/checkout@v4
          with:                                                                                                                                                                  
            fetch-depth: 0                                            

        - name: Run Adversarial Security Review
          with:
            prompt: |                         
              (~50-line adversarial reviewer prompt; full text at                                                                                                                
              .github/adversarial-reviewer-prompt.md in the repo.     
              Prompt text is not relevant to the bug; the cache                                                                                                                  
              failure fires regardless of prompt content.)

**API Provider**

[- ] Anthropic First-Party API (default)
[- ] AWS Bedrock
[- ] GCP Vertex

**Additional context**
Additional context                                                                                                                                                             
                    
  Cache-entry inspection confirms the key+ref collision pattern. Running gh api /repos/<owner>/<repo>/actions/caches on a repo exhibiting this bug shows one 35 MB Bun cache     
  entry per ref:                                                      
                                          
  35242126 bytes | ref=refs/pull/117/merge | key=bun-swrvKXH4hkDYFKcrNV+Kct676IY= | last=2026-04-23T14:25:02
  35242123 bytes | ref=refs/pull/109/merge | key=bun-swrvKXH4hkDYFKcrNV+Kct676IY= | last=2026-04-22T15:57:25                                                                     
  35242125 bytes | ref=refs/pull/91/merge  | key=bun-swrvKXH4hkDYFKcrNV+Kct676IY= | last=2026-04-21T13:52:22
  35242125 bytes | ref=refs/heads/main     | key=swrvKXH4hkDYFKcrNV+Kct676IY=     | last=2026-04-20T18:58:16                                                                     
                                                                                                                                                                                 
  Each ref has exactly one cache entry at the pinned 35 MB Bun binary size. The last_accessed_at stamp is the first successful save; all subsequent saves on the same ref fail   
  with the retry pattern because the key is deterministic (Bun version is pinned, so the hash is stable).                                                                        
                                                                                                                                                                                 
  Root cause. GitHub's cache API enforces "one save per key+ref" and returns a 409-shaped response (rendered as HTML at the client's level) when the key already exists. The     
  @actions/cache toolkit's default retry behavior treats the HTML parse failure as a transient error and burns 5 retry cycles before surfacing the warning. The bug surfaces on  
  every second-and-subsequent push to a PR because the bundled Bun binary cache key is stable across runs of the same action version.                                            
                                                                                                                                                                                 
  Impact. Roughly 20-30 seconds of wallclock time wasted per failed save. For a repo running the action on every PR push, this accumulates to tens of minutes per month of
  Actions quota burned on retry backoff. It also creates persistent warning noise in CI logs, which can mask real cache-related issues.                                          
                                                                      
  Proposed fixes (any of, or combination).                                                                                                                                       
                                          
  1. Use a key with a run-unique suffix plus restoreKeys fallback to the deterministic bun-version hash. Standard @actions/cache pattern. Example: key: bun-<pinned>-${{         
  github.run_id }}, restore-keys: bun-<pinned>-. Each run saves its own entry; restore falls back to the most recent matching prefix. Best long-term fix.                        
  2. Catch the 409/HTML response explicitly in the save step and log as "cache already exists for key+ref, skipping save" without retrying. Smallest-diff fix. Preserves current
  caching semantics; just stops the retry burn.                                                                                                                                  
  3. Expose an input to disable caching (save_cache: false or similar) for users who want to opt out entirely. Useful escape hatch independent of (1) and (2).
  4. Bundle a newer @actions/cache client if upstream already handles 409 better in a recent release. Worth checking actions/toolkit history.                                    
                                                                                                                                                                                 
  Environment.                                
                                                                                                                                                                                 
  - Action version: anthropics/claude-code-action@v1 (resolves to v1.0.104 as of 2026-04-23)                                                                                     
  - Runner: ubuntu-latest                                                                                                                                                        
  - Trigger: pull_request with types: [opened, synchronize, reopened]                                                                                                            
  - Repo visibility: private (should not be relevant)                                                                                                                            
                                                                      
  Happy to contribute a PR for fix (1) or (2) if helpful. Needed guidance on which approach maintainers prefer.

MukundaKatta:
## Why

Closes #1252.

`anthropics/claude-code-action@v1`'s post-job cache save fails consistently on every second-and-subsequent run against the same PR ref. From the issue:

```
Attempt 1 of 5 failed with error: Unexpected token '<', "<!DOCTYPE "... is not valid JSON. Retrying request in 3000 ms...
Attempt 2 of 5 failed with error: Unexpected token '<', "<!DOCTYPE "... is not valid JSON. Retrying request in 5276 ms...
...
Warning: Failed to save: Failed to FinalizeCacheEntryUpload: Failed to make request after 5 attempts
```

Root cause: `oven-sh/setup-bun@v2.2.0`'s post step calls `@actions/cache.saveCache` with a deterministic key derived from the Bun download URL (`bun-<hash>`). The key is identical across runs, so on every PR re-push the GitHub cache API rejects the duplicate key+ref with a 409 (HTML body). `@actions/cache@5.x` treats the unparsable response as transient and burns ~20-30s on 5 retries before surfacing a warning. The cache key isn't ref-aware and `setup-bun` doesn't expose a way to override it, so the issue is unavoidable as long as the cache is enabled.

This burns ~20-30s of wallclock per PR push after the first, plus persistent warning noise that masks real cache issues. See #1252 for the full repro and cache-entry inspection.

## What

Pass `no-cache: true` to `oven-sh/setup-bun` in both `action.yml` and `base-action/action.yml`. The 35 MB Bun binary downloads from `bun.sh` in 2-3s on GitHub-hosted runners (curl install also relies on this in PR #819), so disabling the cache is a net wallclock win and removes the warning entirely.

This is the smallest action-side fix — alternatives would require either patching `setup-bun` upstream (issue author's option 2) or replacing it with curl-based install (issue author's option 4 / open PR #819, which targets a separate GHES bug).

## Tested

Manual verification only — the change is YAML-only inside the composite step, and there are no unit tests for `action.yml` in `test/` (tests cover TypeScript logic).

To verify after merge, run the action twice on the same PR ref and confirm:
- The "Attempt 1 of 5 failed" warning no longer appears in the post-step.
- The Install Bun step prints `Downloading a new version of Bun` on every run instead of `Using a cached version of Bun`.
- Total Install Bun step time stays within a few seconds.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -177,6 +177,14 @@ runs:
       with:
         bun-version: 1.3.6
         token: ${{ inputs.github_token || github.token }}
+        # Disable setup-bun's cache. The upstream save step uses a deterministic
+        # key (Bun version) and isn't ref-aware: on every second-and-subsequent
+        # run against the same PR ref the GitHub cache API rejects the duplicate
+        # key+ref with a 409 (HTML body), and @actions/cache treats the unparsable
+        # response as transient and burns ~20-30s on 5 retries before warning.
+        # The 35 MB Bun binary downloads in 2-3s, so disabling the cache is a net
+        # wallclock win and removes the noisy warning. See issue #1252.
+        no-cache: true
 
     - name: Setup Custom Bun Path
       if: inputs.path_to_bun_executable != ''
--- a/base-action/action.yml
+++ b/base-action/action.yml
@@ -100,6 +100,8 @@ runs:
       uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # https://github.com/oven-sh/setup-bun/releases/tag/v2.2.0
       with:
         bun-version: 1.3.6
+        # Disable setup-bun's cache. See action.yml for details and issue #1252.
+        no-cache: true
 
     - name: Setup Custom Bun Path
       if: inputs.path_to_bun_executable != ''


ndrc81 (commented):
Approving this fix from a downstream-user perspective. We use claude-code-action@v1 in two workflows on a private repo (Claude Code Review and Adversarial Security Review), running on every PR per project policy. Each PR push burns approximately 5 minutes on the setup-bun post-step retries described in #1252, plus warning noise that masks real cache issues. At our current PR volume that pushes us to roughly 95% of the GitHub Team plan's 3000 min/month included allowance, with the burn entirely attributable to this bug.

The action-side fix here (no-cache: true on oven-sh/setup-bun) is the minimum-surface change and avoids the larger refactor of patching setup-bun upstream or migrating to a curl-based install. The 35 MB Bun binary downloads in 2-3s on GitHub-hosted runners, so the disabled-cache wallclock is a clear net win against the 20-30s of failing-retry burn on the current path. Confirming the YAML change is contained to the two action.yml files and does not change any TypeScript logic.

No additional regression risk we can identify. Strongly support merge.

necofuryai:
Reporting a second, independent failure mode that this PR also fixes.

We run `claude-code-action` from a Dependabot-triggered `workflow_run` job with
read-only permissions (untrusted PR content is analyzed there; writes are
isolated to a separate job). Per [GitHub's documented cache policy](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#cache-access-for-low-trust-workflow-triggers),
`workflow_run` is a low-trust trigger that only gets read-only access to the
default branch's cache scope — so unlike the key collision in #1252 (second run
onward), this hits the *first* run too. In all 25 executions of our analysis
job to date, setup-bun's post step logged:

```text
Cache reservation failed: cache write denied: token has no writable scopes
Failed to save: Unable to reserve cache with key bun-...
```

and restore missed in all 25 as well. The cache never provides value in this
configuration, so `no-cache: true` changes nothing for us except removing the
failed post-save and its warning. The only input that avoids it today is
`path_to_bun_executable`, which is documented as normally unnecessary and
version-sensitive.

Evidence: [workflow](https://github.com/necofuryai/necofuryai.dev/blob/main/.github/workflows/dependabot-advisory-review.yml) · [first run](https://github.com/necofuryai/necofuryai.dev/actions/runs/29703758615/job/88237178098) · [latest run](https://github.com/necofuryai/necofuryai.dev/actions/runs/30046802814/job/89339807193)
(observed on v1.0.177/178; the `setup-bun@v2.2.0` invocation is unchanged
through current `main`, so the behavior carries over).

One note: the branch currently conflicts with `main` (#1365 moved the setup-bun
pinning, and `base-action/action.yml` gets frequent version bumps), though the
fix itself remains the same two `no-cache: true` lines. @MukundaKatta if
rebasing is inconvenient, I'm happy to open a follow-up PR against current
`main` with credit and a link back here.

necofuryai:
I opened #1580 against current main as offered. It preserves @MukundaKatta's original commit authorship and links the additional downstream evidence from this thread. Happy to close it if #1259 is rebased instead.

Issue #533, opened 2025-09-04
v1 accessing (MCP) tools despite `allowedTools` being empty?


polis80cy:
Since migrating to v1, I am unable to stop Claude from accessing MCP (and other) tools that it didn't seem to need with the previous beta action. I don't understand why it's trying to load 15 different tools since these tools didn't seem to be part of the beta action and yet the code reviews were being carried out fine.

These tools are also erroring with "this command requires approval". They also seem to utilize a ton of tokens.

I tried to do this through the usage of `allowedTools` and `disallowedTools` but it doesn't seem to work. Probably my syntax is wrong too for these.

This is what I have currently (that doesn't work):

```
claude_args: |
            --system-prompt "Please refer to the project's CLAUDE.md file and .claude/agents/ documentation for:
            - Architecture patterns and conventions to follow
            - Testing patterns and requirements  
            - Code style guidelines and established practices
            - Technology stack and framework usage
            Pay special attention to violations of established patterns documented in these files."
            --allowedTools ""
            --disallowedTools "*"

          # Use sticky comments to make Claude reuse the same comment on subsequent pushes to the same PR
          use_sticky_comment: true
```

**Screenshots**
Here's a screenshot of a run:

<img width="1245" height="694" alt="Image" src="https://github.com/user-attachments/assets/7636763d-21d4-4856-9699-523625b23379" />

**Workflow yml file**
```
name: Claude Code Review

on:
  issue_comment:
    types: [created]
  pull_request_review_comment:
    types: [created]
  pull_request_review:
    types: [submitted]

jobs:
  claude-review:
    if: |
      (github.event_name == 'issue_comment' && github.event.issue.pull_request && contains(github.event.comment.body, '@claude')) ||
      (github.event_name == 'pull_request_review_comment' && contains(github.event.comment.body, '@claude')) ||
      (github.event_name == 'pull_request_review' && contains(github.event.review.body, '@claude'))
    
    runs-on: ubuntu-latest
    permissions:
      contents: read
      pull-requests: write
      issues: write
      id-token: write
    
    steps:
      - name: Checkout repository
        uses: actions/checkout@v4
        with:
          fetch-depth: 0

      - name: Run Claude Code Review
        uses: anthropics/claude-code-action@v1
        with:
          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
          
          prompt: |
            REPO: ${{ github.repository }}
[...]

Issue #264, opened 2025-07-14
Feature Request: Enable Task in tools


ad-m-ss:
Hi

Is there a reason why the `Task` (subagent) tool is not allowed by default ( https://github.com/anthropics/claude-code-action/blob/b3c6de94ea751d8f793c7bd19bfcfd6c067a0786/src/create-prompt/index.ts#L25-L33 )? Are they supported? Are there known limitations?

ad-m-ss:
It looks that Claude Code can execute the tool `Task` despite not having it explicitly allowed.

I am leaving that issue to look over regarding documentation updates.

Issue #181, opened 2025-06-18
Allowed tools from the example in the README do not work


kvz:
**Describe the bug**
Allowed tools from the example in the README do not work

**To Reproduce**
Add this:

```
      - name: Run Claude Code
        id: claude
        uses: anthropics/claude-code-action@beta
        with:
          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
          model: claude-opus-4-20250514
          allowed_tools: 
            - Bash(corepack yarn check)
            - Bash(corepack yarn)
            - Bash(corepack yarn add:*)
            - Bash(gh run:*)
```

Claude Code will complaint it does not have access to these tools when asked.

**Expected behavior**
It should happily run those.

**Additional context**

When I switched to this format, it did work: `allowed_tools: Bash(corepack yarn check),Bash(corepack yarn),Bash(corepack yarn add:*),Bash(gh run:*)`

ashwin-ant:
@kvz where in the readme do you see that syntax?

kvz:
Apologies, this also didn't work for me: 

![Image](https://github.com/user-attachments/assets/4b2add61-f3c8-4cc9-9829-7c1799b16598)

PR #1267, opened 2026-04-27
fix(agent): register built-in tools from --allowedTools

### LINKED ISSUE #690: `--allowedTools` in `claude_args` doesn't prevent default disabling of WebSearch and WebFetch
**Describe the bug**
When using `claude_args` with `--allowedTools` that includes `WebSearch` and `WebFetch`, these tools are still disabled by default through the `DISALLOWED_TOOLS` environment variable.

The `--allowedTools` parameter in `claude_args` is not reflected in the mode's allowed tools list during prompt creation, causing the default disabling behavior to take precedence. This prevents Claude from using web search and fetch capabilities even when explicitly allowed.

**To Reproduce**

1. Create a GitHub Actions workflow with `claude_args` containing `--allowedTools` that includes `WebFetch` and `WebSearch`
2. Trigger the action with a comment containing `@claude` and request it to search the web or fetch a URL
3. Check the execution logs
4. See error: Tools are rejected because `DISALLOWED_TOOLS: WebSearch,WebFetch` is set

**Expected behavior**
Since `WebFetch` and `WebSearch` are explicitly included in `--allowedTools`, they should be available for Claude to use during execution. The `--allowedTools` parameter should override the default disabling behavior for these tools.

**Screenshots**

Image

**Workflow yml file**

```
      - name: ***
        with:
          use_bedrock: "true"
          track_progress: true
          github_token: ${{ steps.app-token.outputs.token }}
          claude_args: |
            --model sonnet
            --mcp-config .mcp.json
            --allowedTools Task,Edit,Read,WebFetch,WebSearch,Glob,Grep,SlashCommand,mcp__github_comment__update_claude_comment,mcp__github_inline_comment__create_inline_comment,Bash(gh pr comment:*),Bash(gh pr diff:*),Bash(gh pr view:*)
            --disallowedTools ""
```

**API Provider**

[ ] Anthropic First-Party API (default)
[x] AWS Bedrock
[ ] GCP Vertex

**Additional context**

### Root Cause Analysis

The issue occurs in `src/create-prompt/index.ts` in the `createPrompt` function:

1. **Mode's getAllowedTools() returns empty array**:
   Both Tag mode (`src/modes/tag/index.ts`) and Agent mode (`src/modes/agent/index.ts`) implement `getAllowedTools()` to return an empty array:
   ```typescript
   getAllowedTools() {
     return [];
   }
   ```

2. **buildDisallowedToolsString applies default disabling**:
   In `src/create-prompt/index.ts`, the `buildDisallowedToolsString` function:
   ```typescript
   export function buildDisallowedToolsString(
     customDisallowedTools?: string[],
     allowedTools?: string[],
   ): string {
     // Tag mode: Disable WebSearch and WebFetch by default for security
     let disallowedTools = ["WebSearch", "WebFetch"];

     // If user has explicitly allowed some default disallowed tools, remove them
     if (allowedTools && allowedTools.length > 0) {
       disallowedTools = disallowedTools.filter(
         (tool) => !allowedTools.includes(tool),
       );
     }
     // ...
   }
   ```

3. **createPrompt uses mode.getAllowedTools()**:
   ```typescript
   const modeAllowedTools = mode.getAllowedTools();  // Returns []
   const modeDisallowedTools = mode.getDisallowedTools();

   const allDisallowedTools = buildDisallowedToolsString(
     modeDisallowedTools,
     modeAllowedTools,  // Empty array, so WebSearch/WebFetch remain disabled
   );

   core.exportVariable("DISALLOWED_TOOLS", allDisallowedTools);  // Sets "WebSearch,WebFetch"
   ```

4. **parseAllowedTools is not connected to mode.getAllowedTools()**:
   While Agent mode has `parseAllowedTools()` function in `src/modes/agent/parse-tools.ts` that can parse `--allowedTools` from `claude_args`, this parsed value is:
   - Used only in `prepareMcpConfig()`
   - **Not** returned by `mode.getAllowedTools()`
   - **Not** available during the `buildDisallowedToolsString()` call

josephfinlayson:
## Symptom

After this PR, `init.tools` reports the actual built-in tool set requested via `--allowedTools` instead of always defaulting to `[Bash, Read]` in agent mode. Today the model literally cannot call `Edit` / `Glob` / `Grep` / `Write` / `MultiEdit` / `LS` even when those names are passed via `claude_args: --allowedTools …`.

## Before / after

Run with `claude_args: --allowedTools 'Bash(git:*),Edit,Read,Glob,Grep,Write'`:

**Before**
```
init: { tools: ["Bash", "Read"], … }
```
The model has no way to call `Edit`, `Glob`, `Grep`, or `Write`.

**After**
```
init: { tools: ["Bash", "Edit", "Read", "Glob", "Grep", "Write"], … }
```
Registration matches what the user asked for.

## Root cause

The SDK's `Options.tools` field is the *registration* knob — it controls which built-in tools are wired into the spawned CLI session. `Options.allowedTools` is a separate, downstream auto-approve gate-list. `parseSdkOptions` constructed `sdkOptions` with `model`, `maxTurns`, `allowedTools`, `disallowedTools`, `systemPrompt`, `fallbackModel`, `pathToClaudeCodeExecutable`, `extraArgs`, `env`, `settingSources` — but never `tools`. With `tools` undefined, the SDK falls through to a reduced default in agent mode and the model only sees `[Bash, Read]`.

Issue #690 has this analysis (verbatim):

> While Agent mode has parseAllowedTools() function in src/modes/agent/parse-tools.ts that can parse --allowedTools from claude_args, this parsed value is: Used only in prepareMcpConfig(), Not returned by mode.getAllowedTools(), Not available during the buildDisallowedToolsString() call.

The fix belongs in the option-parsing layer because both agent and tag modes funnel through `parseSdkOptions`.

## Maintainer-blessed pattern that this PR makes work

`examples/ci-failure-auto-fix.yml` advertises:

```yaml
claude_args: "--allowedTools 'Edit,MultiEdit,Write,Read,Glob,Grep,LS,Bash(git:*),Bash(bun:*),Bash(npm:*),Bash(npx:*),Bash(gh:*)'"
```

Today that string only sets the auto-approve gate-list — the model never actually receives `Edit`/`MultiEdit`/`Write`/`Glob`/`Grep`/`LS`. After this PR, that example does what users (and the docs) expect.

## What changed

`base-action/src/parse-sdk-options.ts`:

1. New `BUILT_IN_TOOL_NAMES` constant: the SDK-known base tools (`Bash`, `Read`, `Write`, `Edit`, `MultiEdit`, `Glob`, `Grep`, `LS`, `NotebookRead`, `NotebookEdit`, `Task`, `TodoWrite`, `WebFetch`, `WebSearch`).
2. New `extractBuiltInTools(mergedAllowedTools)` helper that:
   - strips any `(…)` suffix (`Bash(git:*)` → `Bash`),
   - drops `mcp__*` entries (those are MCP tool names, not base tools),
   - keeps only entries whose stem is in `BUILT_IN_TOOL_NAMES`,
   - dedupes while preserving input order,
   - returns `undefined` if nothing matches (so we leave `Options.tools` unset and the SDK keeps its current default — preserving prior behaviour for that case).
3. Binds the result to `sdkOptions.tools` in the existing `sdkOptions` object literal. `sdkOptions.allowedTools` is unchanged — the auto-approve gate-list still carries the original entries (Bash patterns, MCP names, etc.) so MCP tool gating and `Bash(git:*)`-style patterns keep working.

## Backwards compatibility

- If `--allowedTools` only contains MCP/`Bash(...)` entries (after stem filtering produces no built-ins), `Options.tools` stays undefined and behaviour is unchanged.
- Users currently relying on the broken default will see `init.tools` *expand* to what they asked for, not contract — they always wanted those tools; the auto-approve gate matched the same names, the registration just never followed.
- No changes to `extraArgs`, MCP config merging, or tag/agent mode files.

## Tests

Added six cases to `base-action/test/parse-sdk-options.test.ts` under a new `built-in tool registration (sdkOptions.tools)` describe block:

1. `--allowedTools Edit,Read,Glob,Grep` → `tools = ["Edit","Read","Glob","Grep"]`.
2. `--allowedTools Bash(git:*),Edit,Read,mcp__github_comment__update_claude_comment` → `tools = ["Bash","Edit","Read"]`; `allowedTools` keeps the original entries.
3. No `--allowedTools` → `tools` undefined.
4. `--allowedTools` with only MCP entries → `tools` undefined; `allowedTools` keeps the MCP entries.
5. `--allowedTools Edit,Edit,Read` → `tools = ["Edit","Read"]` (deduped).
6. `--allowedTools Foobar,Read` → `tools = ["Read"]` (unknown stem dropped).
7. Plus a merge test combining `claudeArgs` `--allowedTools` and direct `options.allowedTools`.

`bun test` → 672 pass / 0 fail (34 pass / 0 fail in `parse-sdk-options.test.ts`).
`bun run typecheck` → clean.
`bun run format:check` → clean.

## Fixes / relates to

Fixes #690. Addresses the docs-vs-behaviour gap users hit in #181, #533, and #264.

### DIFF
--- a/base-action/src/parse-sdk-options.ts
+++ b/base-action/src/parse-sdk-options.ts
@@ -24,6 +24,68 @@ const ACCUMULATING_FLAGS = new Set([
 // Delimiter used to join accumulated flag values
 const ACCUMULATE_DELIMITER = "\x00";
 
+// Built-in tool names recognized by the Claude Agent SDK.
+// These are the tools the SDK can register via Options.tools (string[] form).
+// Anything not in this set is either an MCP tool (mcp__*), a Bash() pattern,
+// or an unknown identifier — none of which belong in the registration list.
+// Source: @anthropic-ai/claude-agent-sdk type defs and CLI reference docs.
+const BUILT_IN_TOOL_NAMES = new Set([
+  "Bash",
+  "Read",
+  "Write",
+  "Edit",
+  "MultiEdit",
+  "Glob",
+  "Grep",
+  "LS",
+  "NotebookRead",
+  "NotebookEdit",
+  "Task",
+  "TodoWrite",
+  "WebFetch",
+  "WebSearch",
+]);
+
+/**
+ * Extract the set of built-in tool names that should be registered with the SDK
+ * (`Options.tools`) given a merged allow-list.
+ *
+ * The SDK's `Options.tools` field is the *registration* knob — it controls which
+ * built-in tools are wired into the spawned CLI session. `Options.allowedTools`
+ * is a separate, downstream auto-approve gate-list. Historically this action
+ * only set `allowedTools`, leaving `tools` undefined; in agent mode the result
+ * was that the SDK init reported `tools: ["Bash", "Read"]` regardless of what
+ * the user passed in `claude_args: --allowedTools …`. The model literally could
+ * not call Edit/Glob/Grep/Write/etc. even when allow-listed.
+ *
+ * Filtering rules (kept conservative for backwards compatibility):
+ *  - Strip any `(…)` suffix, e.g. `Bash(git:*)` → `Bash`.
+ *  - Drop entries that start with `mcp__` (MCP tool names, not base tools).
+ *  - Keep only entries whose stem is in `BUILT_IN_TOOL_NAMES`.
+ *  - Deduplicate, preserving input order.
+ *
+ * If the resulting set is empty, returns `undefined` so callers can leave
+ * `Options.tools` unset and let the SDK fall back to its default preset.
+ */
+function extractBuiltInTools(
+  mergedAllowedTools: string[],
+): string[] | undefined {
+  const result: string[] = [];
+  const seen = new Set<string>();
+  for (const entry of mergedAllowedTools) {
+    if (!entry) continue;
+    if (entry.startsWith("mcp__")) continue;
+    // Strip any `(…)` suffix used for Bash command patterns.
+    const parenIdx = entry.indexOf("(");
+    const stem = parenIdx >= 0 ? entry.slice(0, parenIdx) : entry;
+    if (!BUILT_IN_TOOL_NAMES.has(stem)) continue;
+    if (seen.has(stem)) continue;
+    seen.add(stem);
+    result.push(stem);
+  }
+  return result.length > 0 ? result : undefined;
+}
+
 type McpConfig = {
   mcpServers?: Record<string, unknown>;
 };
@@ -253,6 +315,14 @@ export function parseSdkOptions(options: ClaudeOptions): ParsedSdkOptions {
     };
   }
 
+  // Derive the set of built-in tools to register with the SDK. The SDK uses
+  // `Options.tools` as the *registration* knob; `Options.allowedTools` is only
+  // the auto-approve gate-list. Without this, agent mode init reports
+  // `tools: ["Bash", "Read"]` regardless of what the user passes via
+  // `claude_args: --allowedTools …` — the model literally cannot call
+  // Edit/Glob/Grep/Write/etc. (See issues #690, #181, #533, #264.)
+  const builtInTools = extractBuiltInTools(mergedAllowedTools);
+
   // Build SDK options - use merged tools from both direct options and claudeArgs
   const sdkOptions: SdkOptions = {
     // Direct options from ClaudeOptions inputs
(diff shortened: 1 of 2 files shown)


jeffschwane:
This is affecting us in production-like GitHub Actions runs for a custom-built agent.

Our workflow passes built-in tools via `--allowedTools`, including `Read`, `Write`, `Grep`, and `Glob`, but recent runs show the action runtime returning:

- `Error: No such tool available: Grep`
- `Error: No such tool available: Glob`

That causes the agent to burn turns retrying or working around unavailable native search tools, and in some cases the run fails with `Reached maximum number of turns`. These are especially important tools for our use case: the agent is doing repository research against a checked-out workspace before writing structured output, so `Grep`/`Glob` are central to the workflow rather than optional conveniences.

We can add `Bash(rg:*)` as a temporary fallback, but that’s less precise than the native Claude Code tools and means callers have to grant shell access just to recover search capability that was already requested through `--allowedTools`.

This PR’s distinction between tool registration and approval matches the behavior we need: if a workflow explicitly passes built-in tools through `--allowedTools`, those tools should be registered in the SDK session so the model can actually call them.

Would appreciate this getting fixed soon 🙏

PR #1277, opened 2026-04-30
fix: exclude .claude-pr snapshot from git staging

### LINKED ISSUE #1275: `restoreConfigFromBase` snapshot at `.claude-pr/` gets committed to PR branch when bot pushes a fix
## Summary

The `.claude-pr/` snapshot directory created by `restoreConfigFromBase` (introduced in #1172) is unintentionally committed to the user's PR branch whenever the Claude bot makes a fix-up commit during a PR-context run. This results in stealth commits that the bot's own reply comment doesn't acknowledge, polluting the PR with unrelated files.

## Reproduction

Repository conditions:
- Has any of the `SENSITIVE_PATHS` checked into the repo (most commonly `.claude/` and/or `CLAUDE.md`).
- Uses `claude-code-action@v1` with `@claude` mention triggers (`issue_comment` event on a PR).

Steps:
1. A PR is opened. `Claude PR Review` runs and posts a review comment with a suggestion.
2. A maintainer comments `@claude please address that suggestion`.
3. The action runs:
   - `restoreConfigFromBase` snapshots PR-authored `.claude/` and `CLAUDE.md` to `.claude-pr/.claude/` and `.claude-pr/CLAUDE.md`.
   - The CLI makes the requested fix to the actual source file (e.g. `tools/foo/bar.gs`).
   - The CLI commits with `git add -A` (or equivalent) and pushes.
4. **Two commits land on the PR branch**: the requested fix, plus a chore commit dragging in the entire `.claude-pr/` snapshot.

The bot's reply comment only mentions the requested fix — there is no mention of the `.claude-pr/` commit, so the user has no signal that this happened until they read `git log`.

## Expected

`.claude-pr/` is described in `restore-config.ts` as data \"for review agents (not executed)\" — i.e. an ephemeral runner-side artifact. It should never be committed to the user's repository.

## Actual

The directory is committed verbatim, often containing dozens to hundreds of files (every file under `.claude/skills/` etc.). On subsequent `@claude` runs, the now-tracked `.claude-pr/` shows diffs against the freshly-snapshotted version, causing repeated chore commits.

## Existing partial mitigation

`restoreConfigFromBase` already runs `git reset -- ${SENSITIVE_PATHS}` to unstage the base-branch revert of `.claude/`, `CLAUDE.md`, etc., precisely to keep them out of the CLI's later commits. The same protection is missing for `.claude-pr/` itself.

## Suggested fix

In `restoreConfigFromBase`, after creating `.claude-pr/`, append it to `.git/info/exclude` so it is invisible to `git add` regardless of the CLI's staging strategy:

```typescript
if (existsSync(\".claude-pr\")) {
  appendFileSync(\".git/info/exclude\", \"\\n.claude-pr/\\n\");
}
```

`.git/info/exclude` is preferred over modifying `.gitignore` because it does not mutate any tracked file in the user's repo, matching the design intent that `.claude-pr/` is a pure runner-side artifact.

## Real-world incidents

Two stealth commits in our private organization repo, both from bot runs:
- A chore commit adding the initial `.claude-pr/` (~100 files) on a PR about hardware OEM project structure — completely unrelated to the PR.
- A second chore commit re-syncing `.claude-pr/` on a different PR after a `@claude` fix-up request.

In both cases the bot's reply summarized only the requested change, so the maintainer only noticed the extra commit on review.

## Workaround

Add `.claude-pr/` to the consuming repo's `.gitignore` and `git rm -r --cached .claude-pr/`. This works but ideally the action would not require every adopter to discover and apply this fix individually — the symptom is delayed (only appears when `@claude` is invoked on a PR with sensitive paths) and the root cause is opaque from the repo side.

cvan20191:
## Summary

Fixes the `.claude-pr/` snapshot being staged by broad Git commands such as `git add -A`.

## Details

- Adds `ensureClaudePrExcludedFromGit()` in `src/github/operations/restore-config.ts`.
- Writes `/.claude-pr/` to `.git/info/exclude` after `restoreConfigFromBase()` creates the snapshot.
- Leaves the working-tree `.gitignore` unchanged.
- Adds `test/restore-config.test.ts` coverage for repeated restores, existing `.gitignore` contents, `git add -A`, and committed files.

## Related Issues

Fixes #1275.

## Validation

1. `TMPDIR=/tmp bun test test/restore-config.test.ts`
2. `TMPDIR=/tmp bun test`
3. `TMPDIR=/tmp bun run typecheck`
4. `TMPDIR=/tmp bun run format:check`

### DIFF
--- a/src/github/operations/restore-config.ts
+++ b/src/github/operations/restore-config.ts
@@ -1,5 +1,13 @@
 import { execFileSync } from "child_process";
-import { cpSync, existsSync, rmSync } from "fs";
+import {
+  appendFileSync,
+  cpSync,
+  existsSync,
+  mkdirSync,
+  readFileSync,
+  rmSync,
+} from "fs";
+import { dirname } from "path";
 
 // Paths that are both PR-controllable and read from cwd at CLI startup.
 //
@@ -19,6 +27,30 @@ const SENSITIVE_PATHS = [
   "CLAUDE.local.md",
   ".husky",
 ];
+
+const CLAUDE_PR_EXCLUDE_PATTERN = "/.claude-pr/";
+
+function ensureClaudePrExcludedFromGit(): void {
+  const excludePath = execFileSync(
+    "git",
+    ["rev-parse", "--git-path", "info/exclude"],
+    { encoding: "utf8" },
+  ).trim();
+
+  const excludeContents = existsSync(excludePath)
+    ? readFileSync(excludePath, "utf8")
+    : "";
+
+  if (excludeContents.split(/\r?\n/).includes(CLAUDE_PR_EXCLUDE_PATTERN)) {
+    return;
+  }
+
+  mkdirSync(dirname(excludePath), { recursive: true });
+
+  const prefix =
+    excludeContents.length === 0 || excludeContents.endsWith("\n") ? "" : "\n";
+  appendFileSync(excludePath, `${prefix}${CLAUDE_PR_EXCLUDE_PATTERN}\n`);
+}
 
 /**
  * Restores security-sensitive config paths from the PR base branch.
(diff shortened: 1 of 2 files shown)


cvan20191:
@ashwin-ant @OctavianGuzu would you mind reviewing this when you have a moment?

cvan20191:
@ashwin-ant friendly follow-up. This is a follow-on to #1172, which added the `.claude-pr/` snapshot.

This prevents that snapshot from being staged by broad commands like `git add -A` after restore, while keeping the exclusion local via `.git/info/exclude` rather than modifying repo `.gitignore`.

ashwin-ant (approved):

PR #1279, opened 2026-05-01
fix: make trigger_phrase match case-insensitive

### LINKED ISSUE #910: trigger_phrase matching is case-sensitive, should be case-insensitive
## Problem

The `trigger_phrase` matching in `src/github/validation/trigger.ts` is case-sensitive. The regex is constructed without the `i` flag:

```typescript
const regex = new RegExp(
  `(^|\\s)${escapeRegExp(triggerPhrase)}([\\s.,!?;:]|$)`,
);
```

This means `trigger_phrase: "@claude"` will not match `@Claude`, `@CLAUDE`, etc.

## Why this matters

On mobile (iOS autocorrect), typing `@claude` often gets capitalized to `@Claude`. The workflow-level `if` guard uses GitHub's `contains()` function which **is** case-insensitive, so the workflow runs — but the action's internal trigger check rejects it silently:

```
trigger_phrase: @claude
No trigger was met for @claude
Trigger result: false
No trigger found, skipping remaining steps
```

This is confusing because the workflow ran (meaning the comment did contain `@claude` per GitHub's case-insensitive `contains()`), but the action decided it didn't match.

## Suggested fix

Add the `i` flag to the RegExp constructor:

```typescript
const regex = new RegExp(
  `(^|\\s)${escapeRegExp(triggerPhrase)}([\\s.,!?;:]|$)`,
  'i',
);
```

This applies to all the trigger checks in `checkContainsTrigger`.

## Environment

- Action version: v1
- Trigger phrase: `@claude` (default)
- Comment text: `@Claude do this` (capital C, typed on iOS)

JustinBis:
Closes #910.

`@Claude` (capital C — from GitHub's @-mention autocomplete or iOS autocorrect) doesn't trigger the action because the regex in `checkContainsTrigger` has no `i` flag:

```
trigger_phrase: @claude
No trigger was met for @claude
Trigger result: false
No trigger found, skipping remaining steps
```

The workflow's outer `if: contains(github.event.comment.body, '@claude')` is case-insensitive, so the job runs but the action silently exits — looks to the user like Claude saw the mention and ignored it.

**Fix:** add `"i"` to the four `RegExp` constructions. Word-boundary behavior (`email@claude.com` → false) is unchanged. Added test cases for `@Claude` and `@CLAUDE`.

```
$ bun test
665 pass, 0 fail
```

### DIFF
--- a/src/github/validation/trigger.ts
+++ b/src/github/validation/trigger.ts
@@ -51,6 +51,7 @@ export function checkContainsTrigger(context: ParsedGitHubContext): boolean {
     // Check for exact match with word boundaries or punctuation
     const regex = new RegExp(
       `(^|\\s)${escapeRegExp(triggerPhrase)}([\\s.,!?;:]|$)`,
+      "i",
     );
 
     // Check in body
@@ -77,6 +78,7 @@ export function checkContainsTrigger(context: ParsedGitHubContext): boolean {
     // Check for exact match with word boundaries or punctuation
     const regex = new RegExp(
       `(^|\\s)${escapeRegExp(triggerPhrase)}([\\s.,!?;:]|$)`,
+      "i",
     );
 
     // Check in body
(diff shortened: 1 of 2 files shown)


JustinBis:
@ashwin-ant — small one closing #910. You'd flagged in Slack back in Jan that you didn't think this was case-sensitive; turns out it has been since the initial commit.

ashwin-ant (approved):

Issue #702, opened 2025-11-26
Bug: Image media type detection fails for GIF images saved with .png extension


nackyAir:
**Describe the bug**

When Claude Code Action downloads images from GitHub comments, it incorrectly determines the media type based on the file extension rather than the actual file content. This causes API errors when the actual image format differs from the extension.

The problematic image is ironically Claude's own spinner animation GIF that it adds to comments during processing.

**To Reproduce**

1. Set up a workflow with `direct_prompt` to trigger on PR open
2. Claude starts processing and posts a comment with the spinner GIF: `<img src="https://github.com/user-attachments/assets/5ac382c7-e004-429b-8e35-7feb3e8f9c6f" />`
3. The action downloads the image and saves it as `/tmp/github-images/image-xxx-0.png`
4. Claude attempts to read the image file
5. The image is sent to Claude API with `media_type: "image/png"`, but the actual data is GIF format
6. API returns error: `Image does not match the provided media type image/png`

**Expected behavior**

The action should detect the actual image format by inspecting the file content (magic bytes) rather than relying on the file extension or URL.

**Screenshots**

N/A

**Workflow yml file**

```yaml
name: Claude PR Review

on:
  pull_request:
    types: [opened, synchronize]

jobs:
  claude-review:
    runs-on: blacksmith-4vcpu-ubuntu-2404
    permissions:
      contents: read
      pull-requests: read
      issues: read
      id-token: write
    steps:
      - name: Checkout code
        uses: actions/checkout@v4
        with:
          fetch-depth: 1

      - name: Run Claude PR Review
        uses: anthropics/claude-code-action@beta
        with:
          claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
          direct_prompt: |
            Please review this pull request and provide feedback on:
            - Code quality and best practices
            - Potential bugs or issues
            - Performance considerations
            - Security concerns
            - Test coverage
            Be constructive and helpful in your feedback.
```

**API Provider**

[x] Anthropic First-Party API (default)
[ ] AWS Bedrock
[ ] GCP Vertex

**Additional context**

Error from logs:
```
API Error: 400 {"type":"error","error":{"type":"invalid_request_error","message":"messages.1.content.10.image.source.base64.data: Image does not match the provided media type image/png"},"request_id":"req_011CVVczVbPxv12AWe44tPoC"}
```

Relevant log showing the issue:
```
[...]

Issue #495, opened 2025-08-27
v1 cannot process issues with JPEGs in them


AnthonyRonning:
Describe the bug                                                                                                                                                                                                                                   
  Claude GitHub Action fails with an API error when processing GitHub issues that contain uploaded images. The error indicates a media type mismatch where the base64 image data doesn't match the expected PNG format.                              
                                                                                                                                                                                                                                                     
  To Reproduce                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                     
  1. Create a GitHub issue in a repository with Claude GitHub Action enabled                                                                                                                                                                         
  2. Upload an image directly to the issue (drag and drop or paste)                                                                                                                                                                                  
  3. Mention @claude in the issue to trigger the action                                                                                                                                                                                              
  4. Claude responds with an API error about image format mismatch                                                                                                                                                                                   
                                                                                                                                                                                                                                                     
  Expected behavior                                                                                                                                                                                                                                  
  Claude should be able to process issues containing images, either by:                                                                                                                                                                              
  - Correctly handling various image formats (JPEG, PNG, etc.)                                                                                                                                                                                       
  - Gracefully skipping image content if unable to process                                                                                                                                                                                           
  - Converting images to the correct format before sending to the API                                                                                                                                                                                
                                                                                                                                                                                                                                                     
  Screenshots                                                                                                                                                                                                                                        
  The error message received: 
```
{"type":"user","message":{"role":"user","content":[{"tool_use_id":"toolu_01WnM9ZBxs2vCwhKiSdzoJmn","type":"tool_result","content":[{"type":"image","source":{"type":"base64","data":"/9j/4AAQSkZJRgABAQAASABIAAD/4QBARXhpZgAATU0AKgAAAA
...
                                                                                                                                                                                                                       
  API Error: 400 {                                                                                                                                                                                                                                   
    "type":"error",                                                                                                                                                                                                                                  
    "error":{                                                                                                                                                                                                                                        
      "type":"invalid_request_error",                                                                                                                                                                                                                
      "message":"messages.1.content.6.image.source.base64.data: Image does not match the provided media type image/png"                                                                                                                              
    },                                                                                                                                                                                                                                               
    "request_id":"req_011CSYPDcYC4wR98Ljz3ke1u"                                                                                                                                                                                                      
  }                                                                                                                                                                                                                                                  
```                                                                              
API Provider

  [x] Anthropic First-Party API (default)
  [ ] AWS Bedrock
[...]

PR #1280, opened 2026-05-02
fix(image-downloader): detect media type from magic bytes instead of URL extension

bakrsabeeh:
# Fix: Detect image media type from magic bytes instead of file extension

## Problem

When Claude Code Action downloads images from GitHub comments, it determines
the `media_type` field (sent to the Anthropic API) from the **file extension**
in the URL — not from the actual file content.

This causes a `400 invalid_request_error` whenever the real format doesn't
match the extension. The most common trigger is **Claude's own spinner GIF**,
which the action adds to comments during processing and then downloads back
as `/tmp/github-images/image-xxx-0.png`.  Because the file is saved as `.png`
but contains GIF bytes, the API rejects it with:

```
API Error: 400
"Image does not match the provided media type image/png"
```

Fixes: #702, #495

Related issues in `claude-code`: #13396, #30124, #31642, #39146, #11931

---

## Root cause

`src/github/image-downloader.ts` contained two functions:

```ts
// BEFORE — extension-based (broken)
function getImageExtension(url: string): string {
  const match = url.match(/\.([a-zA-Z0-9]+)(?:\?|$)/);
  if (match && ["png","jpg","jpeg","gif","webp"].includes(match[1].toLowerCase()))
    return match[1].toLowerCase();
  return "png"; // ← always falls back to "png", even for GIFs
}

function getMediaType(ext: string): string {
  const map: Record<string, string> = {
    png: "image/png", jpg: "image/jpeg", jpeg: "image/jpeg",
    gif: "image/gif", webp: "image/webp",
  };
  return map[ext] ?? "image/png"; // ← trusts the extension
}
```

The downloaded file was also saved with the URL-derived extension (e.g.
`.png`), so even checking the saved filename would reproduce the same error.

---

## Fix

Replace extension-based detection with **magic byte sniffing**:

1. Save downloaded files with a neutral `.bin` extension — no format assumption.
2. `detectMediaTypeFromBytes(buffer)` reads the first 12 bytes of the file and
   matches against known signatures:
   | Format | Signature |
   |--------|-----------|
   | PNG    | `\x89PNG\r\n\x1a\n` (8 bytes) |
   | JPEG   | `\xFF\xD8\xFF` (3 bytes) |
   | GIF87a | `GIF87a` (6 bytes) |
   | GIF89a | `GIF89a` (6 bytes) |
   | WebP   | `RIFF????WEBP` (bytes 0-3 + 8-11) |
3. `getMediaType(filePath)` opens the file, reads 12 bytes, calls
   `detectMediaTypeFromBytes`, and falls back to extension-based detection
   only when the magic bytes are inconclusive.

Only 12 bytes are read per file — zero performance impact.

---

## Files changed

- `src/github/image-downloader.ts` — core fix (magic byte detection)

---

## Tests

Unit tests cover:
- All 5 supported formats detected correctly from magic bytes
- **Core bug scenario**: GIF bytes in a `.png` file → correctly returns `image/gif`
- JPEG bytes in a `.png` file → correctly returns `image/jpeg`
- WebP bytes in a `.jpg` file → correctly returns `image/webp`
- Extension fallback when magic bytes are inconclusive
- Default `image/png` when both magic bytes and extension are unknown
- Edge case: buffer shorter than 4 bytes returns `null`

```
✓ detectMediaTypeFromBytes > detects PNG
✓ detectMediaTypeFromBytes > detects JPEG
✓ detectMediaTypeFromBytes > detects GIF87a
✓ detectMediaTypeFromBytes > detects GIF89a
✓ detectMediaTypeFromBytes > detects WebP
✓ detectMediaTypeFromBytes > returns null for unknown bytes
✓ detectMediaTypeFromBytes > returns null for buffer < 4 bytes
✓ getMediaType > returns image/gif for a GIF saved with .png extension (core bug)
✓ getMediaType > returns image/jpeg for a JPEG saved with .png extension
✓ getMediaType > returns image/webp for a WebP saved with .jpg extension
✓ getMediaType > returns image/png for a genuine PNG file
✓ getMediaType > falls back to extension when magic bytes unknown
✓ getMediaType > falls back to image/png when both unknown
```

### DIFF
--- a/src/github/utils/image-downloader.ts
+++ b/src/github/utils/image-downloader.ts
@@ -192,10 +192,6 @@ export async function downloadCommentImages(
           continue;
         }
 
-        const fileExtension = getImageExtension(originalUrl);
-        const filename = `image-${Date.now()}-${i}${fileExtension}`;
-        const localPath = path.join(downloadsDir, filename);
-
         try {
           console.log(`Downloading ${originalUrl}...`);
 
@@ -208,6 +204,16 @@ export async function downloadCommentImages(
 
           const arrayBuffer = await imageResponse.arrayBuffer();
           const buffer = Buffer.from(arrayBuffer);
+
+          // Detect the real image format from magic bytes in the downloaded
+          // buffer. Fall back to the URL extension only when the format cannot
+          // be identified — this fixes 400 errors caused by mismatches between
+          // the URL extension and the actual file content (e.g. Claude's own
+          // spinner GIF that GitHub serves with a .png URL).
+          const fileExtension =
+            getExtensionFromBuffer(buffer) ?? getImageExtension(originalUrl);
+          const filename = `image-${Date.now()}-${i}${fileExtension}`;
+          const localPath = path.join(downloadsDir, filename);
 
           await fs.writeFile(localPath, buffer);
           console.log(`✓ Saved: ${localPath}`);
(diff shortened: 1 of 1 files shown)


ashwin-ant:
Thanks for this! This was fixed in #1396 (magic-byte detection), so closing as resolved.

PR #1327, opened 2026-05-19
fix: preserve symlinks in .config/ during restore to prevent ENOENT on dangling links

### LINKED ISSUE #1321: `restoreConfigFromBase` crashes on dangling symlinks in `.claude/` directory
**Describe the bug**

The `restoreConfigFromBase` step fails with `ENOENT` when `.claude/skills/` contains symlinks whose targets don't exist at the time the action runs. This started with Claude Code v2.1.142 — v2.1.141 on the same repo/branch worked fine.

Our project shares skill definitions across many repos via a package dependency. `.claude/skills/` contains symlinks into that dependency:

.claude/skills/check -> ../../vendor/package/.claude/skills/check

These resolve after dependency installation, but the action's restore step runs before that, so the symlink target is missing.

**To Reproduce**

1. Create a repo where `.claude/skills/` contains a symlink to a file outside the repo (e.g., inside `node_modules` or a similar dependency directory)
2. Do *not* install dependencies before the `claude-code-action` step
3. Open a PR
4. Action fails at the restore step:

Restoring .claude, .mcp.json, .claude.json, .gitmodules, .ripgreprc, CLAUDE.md, CLAUDE.local.md, .husky from origin/master (PR head is untrusted)
##[error]Action failed with error: ENOENT: no such file or directory, statx '.claude/skills/check'

**Expected behavior**

The restore step should preserve symlinks as-is without following/resolving them. Dangling symlinks in `.claude/` should not cause a hard failure — they're expected to resolve later in the workflow.

**Workflow yml file**

```yaml
jobs:
  review:
    runs-on: [self-hosted, ubuntu-22.04]
    steps:
      - uses: actions/checkout@v4
        with:
          fetch-depth: 50

      # Note: no dependency install step before claude-code-action

        with:
          use_vertex: 'true'
          prompt: "Review this PR"
```

API Provider

- GCP Vertex

Additional context

- The related fix in #1311 addresses a different ENOENT path (missing .claude-pr/ parent directory) and does not cover this case.
- Working: Claude Code v2.1.141 (runs before May 15)
- Broken: Claude Code v2.1.142+ (runs from May 15 onward)
- The cpSync call in restoreConfigFromBase appears to follow symlinks via statx. Using lstat-based copying or tolerating dangling symlinks would fix this.

anishesg:
The `restoreConfigFromBase` function crashes with ENOENT when `.config/` contains symlinks pointing to files that don't exist yet (e.g., symlinks into dependency directories before installation).

The root cause is on line 89 of `src/github/operations/restore-config.ts`, where `cpSync` is called with `dereference: true`. This option forces Node.js to follow symlinks and stat their targets, causing the copy operation to fail when the target doesn't exist.

The fix changes `dereference: true` to `dereference: false`, which preserves symlinks as-is without attempting to resolve them. This allows dangling symlinks in `.config/` to be copied to `.config-pr/` and later restored from the base branch without crashing. The symlinks can resolve naturally after dependency installation later in the workflow.

A regression test was added to verify that dangling symlinks in `.config/skills/` no longer cause the restore step to crash.

Fixes #1321

### DIFF
--- a/src/github/operations/restore-config.ts
+++ b/src/github/operations/restore-config.ts
@@ -86,7 +86,7 @@ export function restoreConfigFromBase(baseBranch: string): void {
   rmSync(".claude-pr", { recursive: true, force: true });
   for (const p of SENSITIVE_PATHS) {
     if (existsSync(p)) {
-      cpSync(p, `.claude-pr/${p}`, { recursive: true, dereference: true });
+      cpSync(p, `.claude-pr/${p}`, { recursive: true, dereference: false });
     }
   }
   if (existsSync(".claude-pr")) {
--- a/test/restore-config.test.ts
+++ b/test/restore-config.test.ts
@@ -134,6 +134,27 @@ describe("restoreConfigFromBase", () => {
     expect(countClaudePrExcludeEntries()).toBe(1);
   });
 
+  test("handles dangling symlinks in .claude/ without crashing", () => {
+    const skillsDir = join(repoDir, ".claude/skills");
+    mkdirSync(skillsDir, { recursive: true });
+
+    // Create a symlink pointing to a non-existent file (simulating a dependency
+    // directory that hasn't been installed yet)
+    execFileSync("ln", ["-s", "../../vendor/package/.claude/skills/check", join(skillsDir, "check")], {
+      cwd: repoDir,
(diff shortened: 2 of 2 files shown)


anishesg:

PR #1356, opened 2026-05-26
docs: fix execution log parsing example

sarvesh1327:
## Summary
- Updates the `execution_file` parsing example in the base-action README to match the SDK message shape written by the action.
- Extracts assistant text from `entry.message.content` text blocks instead of looking for top-level `role`/`content` fields.

## Test Plan
- [x] Ran a local `node -e` sample parse check for SDK-style assistant messages.
- [x] `git diff --check -- base-action/README.md`

## Notes
- Scope intentionally limited to the documented PR review example.


ashwin-ant:
fixed in #1297

PR #1407, opened 2026-06-11
fix: share one exchanged WIF credential across spawned Claude processes

### LINKED ISSUE #1406: WIF federation fails with 401 (jti_reused) when plugins are configured — each spawned claude process re-exchanges the single-use OIDC token
## Describe the bug

When workload identity federation (WIF) is used together with the `plugins` / `plugin_marketplaces` inputs, the action reliably fails with:

```text
Claude Code returned an error result: API Error: Token exchange failed with status 401
(request-id req_...): {"error":{"type":"authentication_error","message":"Authentication failed"}}
Ensure your federation rule matches your identity token.
```

The run dies approximately 3 minutes after start with `num_turns: 1` and `total_cost_usd: 0` — no API call ever succeeds. The Claude Console authentication history fills up with `platform_federated_authentication` failure events with reason **`jti_reused`**.

Removing the `plugins` input, with no other workflow changes, makes the same workflow succeed.

## Root cause

This was traced with an eBPF process/network sensor on the runner and reproduced locally against a mock that enforces single-use `jti`.

1. A GitHub OIDC token is **single-use** at the Anthropic token-exchange endpoint. The same `jti` cannot be exchanged twice.

2. With plugins configured, the action spawns **several short-lived `claude` processes per job**:

   * `claude plugin marketplace add`
   * one `claude plugin install` per plugin
   * the main query, via the Agent SDK

3. Each process resolves federation from the bare env vars:

   * `ANTHROPIC_FEDERATION_RULE_ID`
   * other related federation env vars
   * `ANTHROPIC_IDENTITY_TOKEN_FILE`

   Each process then performs **its own token exchange using the same identity-token file**.

   The first process succeeds. Every subsequent process re-presents the now-consumed `jti` and receives `401 jti_reused`.

4. By the time the main query runs, the token has already been burned. It retries the exchange with exponential backoff for approximately 3 minutes. The same file still contains the same `jti`, because the action's background refresh runs every 4 minutes, and the job eventually fails.

eBPF process trace from a failing run — five separate `claude` processes each contact `api.anthropic.com` with the same identity-token file:

```text
16:06:36  pid 2225  claude (native binary, install-time handshake)
16:06:45  pid 2259  claude plugin marketplace add
16:06:46  pid 2306  claude plugin install differential-review@trailofbits
16:06:47  pid 2315  claude plugin install fp-check@trailofbits
16:06:48  pid 2324  claude (main query, agent-sdk)        ← fails, token already consumed
```

The federation rule itself is fine. Exchanging the same workflow's OIDC token once via a raw `curl` to `POST /v1/oauth/token` returns `200`, and a second exchange of the same JWT returns the exact `401` above.

## To reproduce

1. Configure WIF using a federation rule for a GitHub Actions repo, following the docs.
2. Run the workflow below on a GitHub-hosted runner using `ubuntu-24.04`.
3. The run fails after approximately 3 minutes with the token-exchange `401`.
4. The Console shows one successful exchange followed by repeated `jti_reused` failures for the same `jti`.
5. Remove the `plugins` / `plugin_marketplaces` inputs.
6. The run succeeds.

This was reproduced on:

* `v1.0.139`, which bundles Claude Code `2.1.167`
* `v1.0.144`, which bundles Claude Code `2.1.173`

This does not appear to be fixed by version bumps.

In our org, this happened on 100% of runs with plugins — 10+ runs across two repos and three workflows — and 0% of runs without plugins.

## Expected behavior

Plugins and WIF should work together.

The exchanged access token should be shared across the `claude` processes spawned by the action, so the single-use identity token is only exchanged once per token lifetime.

## Workflow YAML file

Minimal reproduction:

```yaml
name: wif-plugins-repro

on: workflow_dispatch

permissions:
  id-token: write
  contents: read

jobs:
  repro:
    runs-on: ubuntu-24.04

    steps:
      - uses: actions/checkout@v4

      - uses: anthropics/claude-code-action/base-action@v1.0.144
        with:
          prompt: "Say hi in one word and stop. Do not use any tools."
          anthropic_federation_rule_id: "fdrl_***"
          anthropic_organization_id: "***"
          anthropic_service_account_id: "svac_***"
          anthropic_workspace_id: "wrkspc_***"
          claude_args: '--model claude-opus-4-6 --allowedTools "Read"'
          plugin_marketplaces: |
            https://github.com/trailofbits/skills.git
          plugins: |
            differential-review@trailofbits
            fp-check@trailofbits
```

## API Provider

* [x] Anthropic First-Party API — default
* [ ] AWS Bedrock
* [ ] GCP Vertex

## Additional context

Sample failing request IDs, in UTC, in case server-side logs help:

* `req_011CbwgekzRwSukwKoVFBMnm` — 2026-06-11 15:09
* `req_011CbwiRpxvNGfJSs13mMVnT` — 2026-06-11 around 16:09

Happy to share the federation rule / org IDs privately.

The `jti_reused` events are misleading during diagnosis. The **first** event per `jti` is a success, or the real error, and everything after that is retry noise. We initially chased the federation rule configuration because of this.

## Proposed fix

Happy to open a PR.

The Anthropic SDK only enables its **on-disk credentials cache**:

```text
<config_dir>/credentials/<profile>.json
```

This cache is shared across processes, but it is only enabled when federation is loaded from a profile config file — not from bare env vars.

If `setupWorkloadIdentity()` additionally writes a profile pointing at the identity-token file and selects it via:

* `ANTHROPIC_CONFIG_DIR`
* `ANTHROPIC_PROFILE`

then the first process exchanges once, and every other process reuses the cached access token.

The env vars can remain as a fallback for older CLIs.

Implemented and verified here:

```text
https://github.com/KeisukeYamashita/claude-code-action/compare/main...fix/wif-shared-credentials-cache
```

With that change, the exact configuration above, which previously failed 100% of the time, succeeds:

* `is_error: false`
* main query completes in approximately 6 seconds
* the eBPF trace shows the same five processes, but only one token exchange
* existing test suite passes: `702/702`
* new unit tests were added for the profile file

KeisukeYamashita:
Fixes #1406

## Problem

GitHub OIDC tokens are single-use at the Anthropic token-exchange endpoint (the same `jti` cannot be exchanged twice). When the `plugins` / `plugin_marketplaces` inputs are configured, the action spawns several short-lived `claude` processes per job (`claude plugin marketplace add`, one `claude plugin install` per plugin, then the main query). Each process resolved workload identity federation from the bare env vars and exchanged the **same identity-token file** independently: the first exchange succeeded and every later process got 401 (`jti_reused`), which the main query retried for ~3 minutes before failing the job. Details, eBPF process traces, and request-ids in #1406.

## Fix

The SDK only enables its on-disk credentials cache (`<config_dir>/credentials/<profile>.json`, shared across processes) when federation is loaded from a **profile config file**, not from bare env vars. `setupWorkloadIdentity()` now additionally writes a profile pointing at the identity-token file and selects it via `ANTHROPIC_CONFIG_DIR` + `ANTHROPIC_PROFILE`, so the first process exchanges once and every other process reuses the cached access token.

- The profile lives in an action-owned dir under `RUNNER_TEMP` (0700 dir / 0600 file), next to the identity token.
- `ANTHROPIC_IDENTITY_TOKEN_FILE` and the federation env vars are kept as a fallback for CLIs that predate profile support — older CLIs degrade to today's behavior, never worse.
- The existing 4-minute identity-token refresh is unchanged; after the cached access token expires, the next exchange picks up the refreshed (new-`jti`) token from the same file path recorded in the profile.

## Verification

- `bun test`: 702/702 pass (includes 2 new unit tests for the profile file), `bun run typecheck`, `bun run format:check` all green.
- Local reproduction against a mock exchange endpoint that enforces single-use `jti`: two sequential `claude` processes sharing one identity token — env-var path: second process retries `jti_reused` for ~3 minutes and dies (matches production failures exactly); profile path: second process reuses the cached credential with zero additional exchanges and succeeds. Verified with both Claude Code 2.1.167 and 2.1.173.
- On a GitHub-hosted runner with a real federation rule: the exact configuration that previously failed on 100% of runs (two trailofbits plugins + WIF, on both v1.0.139 and v1.0.144) succeeds with this branch — `is_error: false`, main query completes in ~6s. An eBPF trace shows the same five `claude` processes contacting `api.anthropic.com`, now with a single token exchange.

## Notes / limitations

- A cold-start race (two processes exchanging before the first cache write) is still theoretically possible, but the action spawns its subprocesses sequentially, so in practice the first one exchanges and the rest reuse. A complete fix (serializing the exchange / locking the credentials file) would belong in the CLI/SDK.
- If a user has pre-set `ANTHROPIC_CONFIG_DIR` / `ANTHROPIC_PROFILE`, they are overridden while federation inputs are configured (federation inputs are an explicit opt-in, and previously the env-var path was similarly authoritative).

### DIFF
--- a/base-action/src/index.ts
+++ b/base-action/src/index.ts
@@ -75,7 +75,8 @@ async function run() {
     core.setOutput("conclusion", "failure");
     process.exit(1);
   } finally {
-    // Stop refreshing the workload identity token file so the process can exit
+    // Stop refreshing the workload identity token file (so the process can
+    // exit) and delete the token material so it doesn't outlive this step
     workloadIdentity?.stop();
   }
 }
--- a/base-action/src/workload-identity.ts
+++ b/base-action/src/workload-identity.ts
@@ -15,7 +15,8 @@
  */
 
 import * as core from "@actions/core";
-import { mkdirSync, writeFileSync } from "fs";
+import { createHash } from "crypto";
+import { mkdirSync, rmSync, writeFileSync } from "fs";
 import { join } from "path";
 import { retryWithBackoff } from "./retry";
 
@@ -51,48 +52,57 @@ async function fetchIdentityToken(audience: string) {
 }
 
 /**
- * Writes a profile config that loads federation from the identity-token file.
- * Resolving federation through a profile (rather than bare env vars) enables
- * the SDK's on-disk credentials cache, so the several `claude` processes the
- * action spawns (plugin installs, main query) share one exchanged access token
- * instead of each re-exchanging the single-use GitHub OIDC token, which fails
- * with 401 (`jti_reused`).
+ * Writes a profile config that switches federation resolution to the
+ * file-backed path. Resolving federation through a profile (rather than bare
+ * env vars) enables the SDK's on-disk credentials cache, so the several
+ * `claude` processes the action spawns (plugin installs, main query) share
+ * one exchanged access token instead of each re-exchanging the single-use
+ * GitHub OIDC token, which fails with 401 (`jti_reused`).
+ *
+ * The profile is intentionally minimal: the SDK gap-fills the federation
+ * fields (rule, organization, identity-token file, service account, base URL)
+ * from the ANTHROPIC_* env vars the action already exports, so the file only
+ * needs to exist to turn the cache on.
+ *
+ * The config dir name embeds a fingerprint of the federation inputs. The
+ * SDK's cache reuses a token on `expires_at` alone, with no record of the
+ * config that minted it, and the token's scope is bound at mint time — so a
+ * later action step in the same job (RUNNER_TEMP is per-job) with different
+ * federation inputs must land in a different dir or it would silently reuse
+ * the first step's token.
+ *
+ * Sharing the cache is only safe while the action spawns its `claude`
+ * subprocesses sequentially: the SDK cache is not cross-process serialized,
+ * and concurrent cache misses would each re-exchange the same single-use
+ * identity token. Parallelizing the plugin installs would reintroduce the
+ * `jti_reused` failures.
  */
-function writeFederationProfile(tokenFile: string): string {
-  const configDir = join(
-    process.env.RUNNER_TEMP || "/tmp",
-    "claude-workload-identity",
-    "config",
-  );
-  const serviceAccountId = process.env.ANTHROPIC_SERVICE_ACCOUNT_ID?.trim();
-  const workspaceId = process.env.ANTHROPIC_WORKSPACE_ID?.trim();
-  const baseUrl = process.env.ANTHROPIC_BASE_URL?.trim();
-
-  const authentication: Record<string, unknown> = {
-    type: "oidc_federation",
-    federation_rule_id: process.env.ANTHROPIC_FEDERATION_RULE_ID?.trim(),
-    identity_token: { source: "file", path: tokenFile },
-  };
-  if (serviceAccountId) {
-    authentication.service_account_id = serviceAccountId;
-  }
-
-  const profile: Record<string, unknown> = {
-    version: "1.0",
-    authentication,
-    organization_id: process.env.ANTHROPIC_ORGANIZATION_ID?.trim(),
-  };
-  if (workspaceId) {
-    profile.workspace_id = workspaceId;
-  }
-  if (baseUrl) {
-    profile.base_url = baseUrl;
-  }
+function writeFederationProfile(baseDir: string): string {
+  // Every input that changes which credential the exchange mints must be in
+  // here; service_account_id and scope are sent in the exchange request body.
+  const fingerprint = createHash("sha256")
+    .update(
+      JSON.stringify([
+        process.env.ANTHROPIC_FEDERATION_RULE_ID?.trim() ?? "",
+        process.env.ANTHROPIC_ORGANIZATION_ID?.trim() ?? "",
+        process.env.ANTHROPIC_SERVICE_ACCOUNT_ID?.trim() ?? "",
+        process.env.ANTHROPIC_WORKSPACE_ID?.trim() ?? "",
+        process.env.ANTHROPIC_BASE_URL?.trim() ?? "",
+        process.env.ANTHROPIC_SCOPE?.trim() ?? "",
+      ]),
+    )
+    .digest("hex")
+    .slice(0, 16);
+  const configDir = join(baseDir, `config-${fingerprint}`);
 
   mkdirSync(join(configDir, "configs"), { recursive: true, mode: 0o700 });
   writeFileSync(
     join(configDir, "configs", "default.json"),
-    JSON.stringify(profile, null, 2),
+    JSON.stringify(
+      { version: "1.0", authentication: { type: "oidc_federation" } },
+      null,
+      2,
+    ),
     { mode: 0o600 },
   );
   return configDir;
@@ -105,7 +115,8 @@ function writeFederationProfile(tokenFile: string): string {
  *
  * Returns undefined when federation is not configured or is shadowed by a
  * higher-precedence credential. Callers must invoke stop() when execution
- * finishes.
+ * finishes; it also deletes the identity token and any cached exchanged
+ * credential.
  */
 export async function setupWorkloadIdentity(): Promise<
   WorkloadIdentityHandle | undefined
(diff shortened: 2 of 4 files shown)


ashwin-ant (changes requested):
Nice find, and the diagnosis holds up. `buildProvider` in `@anthropic-ai/sdk/lib/credentials/credential-chain.mjs` only wraps the exchange in `cachedExchangeProvider` when a `credentials_path` is set, and that only happens for file-backed configs. So env vars can't share a token across processes and a profile can. Your profile also matches `AnthropicConfig` exactly.

Two changes before merge.

**The cache path needs a config discriminator.** `configDir` is a constant under `RUNNER_TEMP`, which is per-job, but the federation rule / org / workspace inputs are per-step. `cachedExchangeProvider` reuses a cached token based on `expires_at` alone, with no fingerprint of the config that minted it, and the workspace is bound at mint time. So two steps in one job with different workspaces will silently reuse the first step's token: wrong scope, wrong billing, no error. Today's uncached path exchanges fresh each step and gets this right, so it's a regression. The SDK comment right above the cache says this is exactly why env-only configs skip it. Hashing `federation_rule_id + organization_id + workspace_id + base_url` into the config dir name fixes it.

**Don't clobber an operator-set `ANTHROPIC_PROFILE` / `ANTHROPIC_CONFIG_DIR`.** You flag this under limitations, but it's free to fix rather than a tradeoff: if `ANTHROPIC_PROFILE` is already set, the disk cache is already on, so just `core.warning()` and skip `writeFederationProfile`. Same shape as the `ANTHROPIC_API_KEY` guard above it. Worth a test.

Smaller stuff:

- `workspace_id`, `base_url`, and `service_account_id` look redundant. `loadConfig` gap-fills them from env for file-backed profiles, and the resolved base URL prefers the env var anyway. `{ version, authentication: { type: "oidc_federation" } }` should do it.
- `stop()` could `rmSync` the token dir, so the exchanged token doesn't outlive the step.
- Worth a docstring line that this depends on the plugin subprocesses spawning sequentially. `cachedExchangeProvider` isn't cross-process serialized, and its "exchanges are idempotent" assumption is what single-use `jti` breaks. Parallelize that loop later and the bug comes back.

One question. Single-use `jti` is server side and I can't confirm it from the client. Your mock assumes it and the Console events point that way, but let me check with the token exchange owners before we call this a correctness fix. If reuse is allowed it's still worth landing, just as a latency fix, and the description should say so.

KeisukeYamashita:
@ashwin-ant Hi, thank you for your kind review.

All four changes are in: https://github.com/anthropics/claude-code-action/pull/1407/commits/34d32284bab3d5d97b255c34da992a3befb3479e

**Config discriminator.** The config dir is now `config-<sha256[:16]>`, derived from the federation inputs. I added two fields beyond the ones you listed: `service_account_id` and `scope`. Both are sent in the exchange request body (`oidc-federation.mjs` builds `{ grant_type, assertion, federation_rule_id, organization_id, service_account_id }`, and `loadConfig` gap-fills `authentication.scope`), so two steps differing only in service account would hit the same silent-reuse bug as the workspace case does. `anthropic_oidc_audience` is deliberately excluded—it changes which JWT is minted, not the scope of the exchanged token.

**Operator-set `ANTHROPIC_PROFILE` / `ANTHROPIC_CONFIG_DIR`.** If either is set, the action now warns and skips `writeFederationProfile` entirely, following the same pattern as the `ANTHROPIC_API_KEY` guard. `ANTHROPIC_IDENTITY_TOKEN_FILE` is still exported so the operator’s profile—or the environment fallback—can consume the token file. Tests were added for both environment variables.

**Minimal profile.** It is now exactly:

```json
{
  "version": "1.0",
  "authentication": {
    "type": "oidc_federation"
  }
}
```

Before shrinking it, I verified that the gap-fill exists in the CLI version pinned by this action (`2.1.173`), not just in a current SDK. The bundled `loadConfig` in the `@anthropic-ai/claude-code-darwin-arm64@2.1.173` binary references `ANTHROPIC_ORGANIZATION_ID`, `ANTHROPIC_IDENTITY_TOKEN_FILE`, `ANTHROPIC_FEDERATION_RULE_ID`, `ANTHROPIC_WORKSPACE_ID`, `ANTHROPIC_BASE_URL`, `ANTHROPIC_SERVICE_ACCOUNT_ID`, and `ANTHROPIC_SCOPE` in its file-backed branch.

**`stop()` cleanup.** `stop()` now uses `rmSync` to remove the entire token directory—the identity token, profile, and SDK-written credentials cache—so nothing outlives the step. This also means that an unchanged config re-exchanges on the next step rather than reusing the cached credentials. The fingerprint still matters for the crash path where `stop()` never runs.

**Docstring.** I added the sequential-spawn dependency note to `writeFederationProfile`: the plugin loop in `install-plugins.ts` uses `for...await`, the SDK cache is not serialized across processes, and parallelizing the installs would reintroduce `jti_reused`.

On the `jti` question: agreed, that is for you to confirm with the token exchange owners. If reuse turns out to be allowed, I’ll reframe the description as a latency fix.

`bun test` passes with `705/705` tests. `bun run typecheck` and `bun run format:check` are also clean.

ashwin-ant (approved):

Issue #1081, opened 2026-03-18
Don't retry non-retryable errors (401, 4xx) in retryWithBackoff


ei-grad:
## Problem

`retryWithBackoff` in `src/utils/retry.ts` retries all errors blindly, including permanent failures like 401 Unauthorized. This wastes ~35 seconds (5s + 10s + 20s backoff) on errors that will never succeed on retry.

Common example: "Workflow validation failed" 401 when a workflow file differs from the default branch — this is deterministic and will fail identically on every attempt.

## Current behavior

```
Attempt 1 of 3...
App token exchange failed: 401 Unauthorized - Workflow validation failed...
Attempt 1 failed: Workflow validation failed...
Retrying in 5 seconds...
Attempt 2 of 3...
App token exchange failed: 401 Unauthorized - Workflow validation failed...
Attempt 2 failed: Workflow validation failed...
Retrying in 10 seconds...
Attempt 3 of 3...
App token exchange failed: 401 Unauthorized - Workflow validation failed...
```

## Root cause

`exchangeForAppToken` already handles the specific `workflow_not_found_on_default_branch` error code by throwing `WorkflowValidationSkipError`, but `retryWithBackoff` catches it and retries anyway since it has no concept of non-retryable errors.

More broadly, any 4xx HTTP error (400, 401, 403, 404, 422) is a client error and non-retryable — only 5xx and network errors are worth retrying.

## Suggested fix

Option A — minimal: re-throw `WorkflowValidationSkipError` immediately in the catch block of `retryWithBackoff`.

Option B — generic: add a `shouldRetry` predicate to `RetryOptions`:

```typescript
export type RetryOptions = {
  maxAttempts?: number;
  initialDelayMs?: number;
  maxDelayMs?: number;
  backoffFactor?: number;
  shouldRetry?: (error: Error) => boolean; // return false to stop retrying
};
```

Then callers can pass their own logic, e.g. `shouldRetry: (e) => !(e instanceof WorkflowValidationSkipError)`.

Option C — HTTP-aware: introduce a typed error (e.g. `HttpError` with a `status` field) and skip retries for any 4xx status.

---

*This issue was generated with [Claude Code](https://claude.ai/code)*

Issue #722, opened 2025-12-05
Workflow fails on first-time setup but error message says 'this is normal, ignore it'


TimEvans:
## Description

When adding the Claude workflow to a repository for the first time, the workflow fails with exit code 1 and shows as ❌ FAILED in GitHub's UI. However, the error message buried in the logs says "this is normal and you should ignore this error."

This creates a contradictory user experience that blocks users from merging their initial setup PR.

## Steps to Reproduce

1. Run `/install-github-app` in Claude Code CLI
2. Create a PR with the generated workflow files
3. Workflow runs and fails with:
```
##[error]Failed to setup GitHub token: Error: Workflow validation failed.
The workflow file must exist and have identical content to the version on
the repository's default branch. If you're seeing this on a PR when you
first add a code review workflow file to your repository, this is normal
and you should ignore this error.
```

## The Problem

**What the user sees in GitHub UI:**
- ❌ FAILED (red X, exit code 1)
- No indication this is expected behavior

**What's hidden in the failed workflow logs** (only visible if you click through and read the output):
```
Failed to setup GitHub token: Error: Workflow validation failed. The workflow
file must exist and have identical content to the version on the repository's
default branch. If you're seeing this on a PR when you first add a code review
workflow file to your repository, this is normal and you should ignore this error.

If you instead wish to use this action with a custom GitHub token or custom
GitHub app, provide a github_token in the uses section of the app in your
workflow yml file.
```

**The failure status directly contradicts the buried message telling users to ignore it.** Most users won't dig into failed workflow logs to find this explanation.

## What Actually Happens

The workflow needs to be on the default branch before GitHub App auth can validate it (security feature). So the first PR will always fail, but it works after merging.

I only discovered this after multiple reinstalls and debugging. I specifically didn't merge the PR because the workflow was failing - which to be fair is the correct response to a failing workflow!

## Suggested Fix

If this failure is expected for first-time setup, the workflow should NOT fail. Options:

1. **Pass with a warning:** Show ✅ success with a warning annotation: "First-time workflow setup detected. Workflow will activate after PR is merged to main."
2. **Skip the run:** Don't run the workflow on PRs that add it for the first time.
3. **Different status:** Use a status that isn't "failed" (maybe neutral/skipped).

## Impact

Users see "failed" and reasonably assume something is wrong, blocking them from completing the setup. The "this is normal" explanation is hidden in logs that users won't check.

Issue #522, opened 2025-09-02
v1: App token exchange failed


pmatos:
**Describe the bug**

Claude review with v1 results in

```
Run oven-sh/setup-bun@735343b667d3e6f658f44d0eca948eb6282f2b76
  with:
    bun-version: 1.2.11
    no-cache: false
Cache hit for: wMLiZ8FsTgrkPpZKjrOwUG0XVBQ=
Received 0 of 34725459 (0.0%), 0.0 MBs/sec
Received 34725459 of 34725459 (100.0%), 23.7 MBs/sec
Cache Size: ~33 MB (34725459 B)
/usr/bin/tar -xf /home/runner/work/_temp/4c1e850b-5fdc-4fff-89b2-c757dd7b02fb/cache.tzst -P -C /home/runner/work/rightkey/rightkey --use-compress-program unzstd
Cache restored successfully
/home/runner/.bun/bin/bun --revision
1.2.11+cb6abd211
Using a cached version of Bun: 1.2.11+cb6abd211
Run cd ${GITHUB_ACTION_PATH}
bun install v1.2.11 (cb6abd21)

+ @types/bun@1.2.11
+ @types/node@20.19.9
+ @types/node-fetch@2.6.12
+ @types/shell-quote@1.7.5
+ prettier@3.5.3
+ typescript@5.8.3
+ @actions/core@1.11.1
+ @actions/github@6.0.1
+ @modelcontextprotocol/sdk@1.16.0
+ @octokit/graphql@8.2.2
+ @octokit/rest@21.1.1
+ @octokit/webhooks-types@7.6.1
+ node-fetch@3.3.2
+ shell-quote@1.8.3
+ zod@3.25.76

149 packages installed [385.00ms]
Run bun run ${GITHUB_ACTION_PATH}/src/entrypoints/prepare.ts
Auto-detected mode: agent for event: pull_request
Requesting OIDC token...
Attempt 1 of 3...
OIDC token successfully obtained
Exchanging OIDC token for app token...
Attempt 1 of 3...
App token exchange failed: 401 Unauthorized - Workflow validation failed. The workflow file must exist and have identical content to the version on the repository's default branch. If you're seeing this on a PR when you first add a code review workflow file to your repository, this is normal and you should ignore this error.
Attempt 1 failed: Workflow validation failed. The workflow file must exist and have identical content to the version on the repository's default branch. If you're seeing this on a PR when you first add a code review workflow file to your repository, this is normal and you should ignore this error.
[...]

Issue #503, opened 2025-08-28
How to test this tool by pushing commits in PR?


changchiyou:
https://github.com/changchiyou/wildrift-gold-efficiency/actions/runs/17291991464/job/49081266316#step:6:281

> Warning: Skipping action due to workflow validation: Workflow validation failed. The workflow file must exist and have identical content to the version on the repository's default branch. If you're seeing this on a PR when you first add a code review workflow file to your repository, this is normal and you should ignore this error.
Action skipped due to workflow validation error. This is expected when adding Claude Code workflows to new repositories or on PRs with workflow changes. If you're seeing this, your workflow will begin working once you merge your PR.

Does the message here mean that I have to merge and wait for the tool be triggered?

ad-m-ss:
@changchiyou To react to comments, you need a tool merged to interact with it.

ashwin-ant:
Yes, you need to merge it into your main branch in order for it to begin working.

PR #1417, opened 2026-06-16
fix: skip workflow validation token exchange failures

### LINKED ISSUE #1343: Workflow-validation skip returns conclusion: failure on modifications but success on first-add
## Observed behavior

When a PR's workflow file differs from the version on `main`, the anti-tamper validator skips the action. But the resulting GitHub Actions conclusion differs based on whether the PR is *adding* or *modifying* the workflow file:

**First-add (PR introduces the workflow file for the first time):**

```
##[warning]Skipping action due to workflow validation: Workflow validation failed.
The workflow file must exist and have identical content to the version on the
repository's default branch. If you're seeing this on a PR when you first add a
code review workflow file to your repository, this is normal and you should
ignore this error.
Attempt 1 failed: Workflow validation failed. ...
Error is not retryable, giving up immediately
...
Action skipped due to workflow validation error. This is expected when adding
Claude Code workflows to new repositories or on PRs with workflow changes.
Exiting due to workflow validation skip
```

Conclusion: `success`. ✅

**Modification (PR changes an existing workflow file):**

```
App token exchange failed: 401 Unauthorized - Workflow validation failed.
The workflow file must exist and have identical content to the version on the
repository's default branch. If you're seeing this on a PR when you first add a
code review workflow file to your repository, this is normal and you should
ignore this error.
Attempt 1 failed: ...
Retrying in 5 seconds...
Attempt 2 of 3...
App token exchange failed: 401 Unauthorized - Workflow validation failed. ...
Retrying in 10 seconds...
Attempt 3 of 3...
##[error]Action failed with error: Workflow validation failed. ...
Operation failed after 3 attempts
##[error]Process completed with exit code 1.
```

Conclusion: `failure`. ❌

## Impact

Both code paths represent the same underlying anti-tamper check, and the user-facing effect is identical: the action does not run. But the surfaced status is opposite.

On modification, the `failure` status:

1. Produces a red ✗ on the PR's checks UI, suggesting something is broken when nothing is.
2. Contradicts the action's own message ("this is normal and you should ignore this error") by emitting a hard error rather than a warning/skip.
3. Makes every future workflow-file improvement PR (updating `allowed-tools`, `model`, `max-turns`, etc.) appear to fail its own Claude review, even though that failure is the expected validator skip.
4. Adds confusion for operators rolling out the action across multiple repos — the first repo gets a green skip, every subsequent workflow tweak gets a red failure.

## Suggested fix

Align the two code paths to consistently return a non-failure conclusion on validator skip. The first-add path already does this correctly (logs as warning, exits success). The modification path should follow the same shape — detect the validation failure upstream of the retry loop and skip with `success`, matching the first-add behavior.

Alternatively, both paths could surface as `skipped` (or `neutral` if `skipped` isn't available as a conclusion) so the UI clearly distinguishes "intentionally not run" from "ran and failed".

## Reproduction

1. Add `claude-code-action@v1` to a repo for the first time via PR — observe `success` outcome with skip.
2. Once merged, open a follow-up PR that modifies the workflow file (e.g., change `--max-turns`) — observe `failure` outcome with same underlying skip.

Both PRs hit the same anti-tamper check; only the conclusion field differs.

Ryanoonan:
## Summary
- Treat token exchange responses with the workflow-validation failure message as `WorkflowValidationSkipError`, matching the existing coded `workflow_not_found_on_default_branch` path.
- Preserve the existing warning/skip behavior for expected workflow validation mismatches instead of retrying and failing the run.
- Add regression coverage for both coded and message-only workflow validation responses.

Fixes #1343
Related: #443, #522, #722, #503, #1081

## Tests
- `npx --yes bun@latest test test/token.test.ts`
- `npx --yes bun@latest test`
- `npx --yes tsc --noEmit`
- `npx --yes prettier@3.5.3 --check src/github/token.ts test/token.test.ts`

### DIFF
--- a/src/github/token.ts
+++ b/src/github/token.ts
@@ -10,6 +10,49 @@ export class WorkflowValidationSkipError extends Error {
   }
 }
 
+type AppTokenExchangeErrorResponse = {
+  error?: {
+    message?: string;
+    details?: {
+      error_code?: string;
+    };
+  };
+  type?: string;
+  message?: string;
+};
+
+const WORKFLOW_VALIDATION_ERROR_CODES = new Set([
+  "workflow_not_found_on_default_branch",
+]);
+
+function getAppTokenExchangeErrorMessage(
+  responseJson: AppTokenExchangeErrorResponse,
+): string {
+  return responseJson.error?.message ?? responseJson.message ?? "Unknown error";
+}
+
+function isWorkflowValidationError(
+  status: number,
+  responseJson: AppTokenExchangeErrorResponse,
+): boolean {
+  const errorCode = responseJson.error?.details?.error_code;
+  if (
+    errorCode !== undefined &&
+    WORKFLOW_VALIDATION_ERROR_CODES.has(errorCode)
+  ) {
+    return true;
+  }
+
+  if (status !== 401) {
+    return false;
+  }
+
+  const workflowValidationMessage = "workflow validation failed";
+  return [responseJson.message, responseJson.error?.message].some((message) =>
+    message?.toLowerCase().includes(workflowValidationMessage),
+  );
+}
+
 async function getOidcToken(): Promise<string> {
   try {
     const oidcToken = await core.getIDToken("claude-code-github-action");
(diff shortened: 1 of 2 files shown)


Ryanoonan:
I found a few related issues:

- #443 tracks the same `Workflow validation failed` / app-token-exchange 401 class, especially around reusable/shared workflows; it also has several reports from workflow-update PRs hitting this behavior.
- #522 is a closed v1 app-token exchange report with the same validation message, where the guidance was to ignore the failure and merge.
- #722 covers the first-time setup version of the same UX problem: expected validator skip surfaced as a failed run.
- #503 captures user confusion around the skip message and whether the workflow starts working only after merge.
- #1081 covers the retry-loop side of this, specifically retrying deterministic workflow-validation 401s.

Just wanted to note these as related, not necessarily fixed.

ashwin-ant (approved):

cedrichld:
good work

Issue #1287, opened 2026-05-06
Stale prompt files in ${RUNNER_TEMP}/claude-prompts/ leak between jobs on non-ephemeral self-hosted runners


kyungilpark:
## Summary

On non-ephemeral self-hosted runners, `claude-code-action` writes prompt files to `${RUNNER_TEMP}/claude-prompts/`, a directory that — contrary to the GitHub Actions documented contract — is **not reliably emptied between jobs**. Files written by one invocation persist on disk and are picked up by subsequent invocations on the same runner agent, causing the model to see stale context from an unrelated prior job.

In our case, a `claude-user-request.txt` left behind by a `@claude` mention-mode invocation in repo A was repeatedly read by later `prompt`-driven review invocations in repo B running on the same runner agent. The model dutifully addressed the stale question in PR review comments — visible cross-job context bleed.

## Environment

- Action: `anthropics/claude-code-action@v1.0.67`
- Runner: self-hosted, non-ephemeral, Linux (one runner agent processes many jobs sequentially)
- GHES (but the bug is environment-agnostic — same on github.com self-hosted)

## Reproduction

1. Configure two workflows on the same self-hosted runner pool:
   - Workflow M: `claude-mention.yml`-style — invokes the action with `prompt: ""` and relies on `@claude` mention triggering
   - Workflow R: `claude-code-review.yml`-style — invokes the action with a non-empty `prompt:` (agent mode, e.g. `"You are a code reviewer. Review this PR."`)
2. Trigger Workflow M with a mention comment containing some distinctive text, e.g. `@claude what's the difference between approach X and approach Y?`. Note the runner agent name that picks it up (`Set up job` log shows `Runner name: …`).
3. Wait for the job to finish. Inspect the runner host:

   ```
   $ ls -la $RUNNER_TEMP/claude-prompts/
   -rw-r--r-- claude-prompt.txt        (the assembled prompt — overwritten each run)
   -rw-r--r-- claude-user-request.txt  (the user request from step 2 — not cleared)
   ```

4. Now trigger Workflow R against the **same runner agent** by submitting a PR. (Re-run if needed until `Runner name` matches.)
5. Observe Workflow R's review output. With non-trivial probability the review body or comment will reference the stale text from step 2 ("regarding your earlier question about X vs Y…", or "this question doesn't seem relevant to this PR — please clarify…"), demonstrating that step-2 content reached the step-4 model context.

[...]

PR #1440, opened 2026-06-27
fix(agent): preserve pre-written claude-user-request.txt in agent mode

### LINKED ISSUE #1427: Agent mode wipe of `claude-prompts/` (PR #1288) breaks composite actions that pre-write `claude-user-request.txt`
## Summary

PR #1288 added an unconditional `await rm(promptDir, { recursive: true, force: true })` at the start of agent mode in `src/modes/agent/index.ts`. This breaks any composite action that wraps `claude-code-action` and pre-writes `claude-user-request.txt` in an earlier step, because that pre-written file is deleted before the SDK ever sees it.

For my use case ([kyosei-action](https://github.com/ncaq/kyosei-action), a code-review composite action), this means `/kyosei <PR_URL>` is no longer processed as a slash command, the skill never starts, and reviews stop firing entirely on `v1.0.146+`. I had to pin to `v1.0.145` as a workaround ([ncaq/kyosei-action#173](https://github.com/ncaq/kyosei-action/pull/173)).

## Why we can't just put the slash command in the `prompt:` input

This was our original approach, and it doesn't work. The dispatch in `base-action/src/run-claude-sdk.ts` deliberately branches on whether `claude-user-request.txt` exists:

```ts
if (!hasUserRequest) {
  // No user request file - use simple string prompt
  return promptContent;
}
// User request file exists - create multi-block message
...
content: [
  { type: "text", text: promptContent }, // Instructions + GitHub context
  { type: "text", text: userRequest },   // User's request (may be a slash command)
],
```

The in-source comment is explicit that only the multi-block branch "allows the CLI to detect and process slash commands in the user request". The simple-string branch delivers the prompt verbatim and the CLI does NOT preprocess `/foo` slash commands or perform `$ARGUMENTS` substitution from the skill's frontmatter.

In a composite action whose purpose is "run a specific skill on every PR", the only stable way to get slash-command processing today is to write `claude-user-request.txt` so the SDK takes the multi-block branch.

## Why losing the file breaks us

A composite action can run steps **before** the `uses: anthropics/claude-code-action@...` step or **after** it. There is no hook in the middle for downstream code to inject a file between the wipe inside `prepareAgentMode` and the SDK invocation. So a downstream pre-write pattern like:

```yaml
- name: Write user request for slash command support
  shell: bash
  run: |
    mkdir -p "${RUNNER_TEMP}/claude-prompts"
    echo "/kyosei ${{ github.event.pull_request.html_url }}" \
      > "${RUNNER_TEMP}/claude-prompts/claude-user-request.txt"

  with:
    prompt: "Follow the user request."
```

worked on `≤ v1.0.145`, and silently no-ops on `≥ v1.0.146`: the file is wiped, the SDK takes the simple-string branch, and Claude only sees `"Follow the user request."` with no `/kyosei` ever processed.

## Suggested fixes

In rough order of preference / lowest blast radius:

1. Add a `user_request` (or `user_request_file`) input to agent mode. When set, write the value to `claude-user-request.txt` after the wipe so the SDK takes the multi-block path. This gives composite actions a sanctioned way to drive a slash command, without depending on a file-system side channel.
2. Switch to per-invocation subdirectory keyed by `${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}-${GITHUB_JOB}` — this is exactly alternative 1 in PR #1288's own description. Solves the original cross-job leak without nuking files an outer step intentionally placed.
3. Scope the wipe to known filenames (`claude-prompt.txt`, `claude-user-request.txt`) rather than the whole directory. Preserves any new files an outer step might leave while still preventing the specific stale-file leak that #1287 reported.

The current `rm -rf` is overbroad: it solves #1287 but at the cost of denying downstream composite actions a way to communicate with the SDK.

## Reproduction

- Composite action writes `${RUNNER_TEMP}/claude-prompts/claude-user-request.txt` before `uses: anthropics/claude-code-action@v1.0.146` (or any later version).
- Set `prompt:` to something generic like `"Follow the user request."`.
- Observe that the slash command in `claude-user-request.txt` is never processed (vs. `v1.0.145` where it is).

Concrete repro: kyosei-action `@v2.2.4` + claude-code-action `@v1.0.146..v1.0.153`. Real failed run: https://github.com/ncaq/dotfiles/actions/runs/27996821836. The action log shows `Context prompt: Follow the user request.` and no `/kyosei` ever appears.

## Environment

- claude-code-action: `v1.0.146` through `v1.0.153` (regression introduced in `v1.0.146`)
- Runner: `ubuntu-24.04` (GitHub-hosted)
- API Provider: Anthropic First-Party API

---

Happy to send a PR for option 1 if that direction sounds acceptable.

syf2211:
## Summary

Agent mode no longer wipes the entire `claude-prompts` directory. It only removes `claude-prompt.txt`, the file agent mode owns and rewrites.

## Motivation

Fixes #1427. Composite workflows can pre-write `claude-user-request.txt` in an earlier step so the SDK takes the multi-block path and processes slash commands. PR #1288 added an unconditional `rm -rf` on the prompt directory, which deleted that pre-written file before the SDK ran.

## Changes

- `prepareAgentMode`: replace directory `rm -rf` with targeted removal of `claude-prompt.txt` only
- Add regression test ensuring a pre-written `claude-user-request.txt` survives agent prepare

## Tests

- `bun test test/modes/agent.test.ts` — 7/7 passed
- `bun run typecheck` — passed
- `bun run format:check` — passed
- composer-2.5 review: **APPROVE**

## Notes

Tag mode still performs full directory cleanup in `create-prompt` (it writes both files). Agent mode intentionally preserves `claude-user-request.txt` for composite workflows. On non-ephemeral self-hosted runners without a pre-write step, a stale user-request file could theoretically persist if `RUNNER_TEMP` is not cleared between jobs — same class of concern noted in #1287, traded off here to unblock composite slash-command workflows.

### DIFF
--- a/src/modes/agent/index.ts
+++ b/src/modes/agent/index.ts
@@ -64,12 +64,11 @@ export async function prepareAgentMode({
     }
   }
 
-  // Create prompt directory. Clear any stale files from a prior invocation first —
-  // see src/create-prompt/index.ts for context (non-ephemeral self-hosted runners
-  // do not reliably honor the RUNNER_TEMP cleanup contract).
+  // Create prompt directory. Only remove the prompt file this mode writes — composite
+  // workflows may pre-write claude-user-request.txt in an earlier step (see #1427).
   const promptDir = `${process.env.RUNNER_TEMP || "/tmp"}/claude-prompts`;
-  await rm(promptDir, { recursive: true, force: true });
   await mkdir(promptDir, { recursive: true });
+  await rm(`${promptDir}/claude-prompt.txt`, { force: true });
 
   // Write the prompt file - use the user's prompt directly
   const promptContent =
--- a/test/modes/agent.test.ts
+++ b/test/modes/agent.test.ts
@@ -7,6 +7,7 @@ import {
   spyOn,
   mock,
 } from "bun:test";
+import { mkdir, readFile, rm, writeFile } from "fs/promises";
 import { prepareAgentMode } from "../../src/modes/agent";
 import { createMockAutomationContext } from "../mockContext";
 import * as core from "@actions/core";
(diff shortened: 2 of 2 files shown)


tarunag10:
This looks like the lowest-blast-radius fix for #1427 to me.

I had explored an alternate first-class `user_request` input in #1434, but this PR better preserves the existing composite-action behavior that regressed in v1.0.146: pre-writing `claude-user-request.txt` before invoking agent mode.

The main validation I would want is exactly what this tests: agent mode rewrites its owned `claude-prompt.txt` while leaving a pre-written `claude-user-request.txt` intact, so the SDK still takes the multi-block path for slash-command processing.

PR #1457, opened 2026-06-30
fix: support push event in automation context

### LINKED ISSUE #1456: Unsupported event type: push — action exits 1 on push-triggered workflows
Describe the bug

  anthropics/claude-code-action@v1.0.160 (commit 4633baf5267540f3f8cb58b684f79901d564c280) fails at startup with Error: Action failed with error:
  Unsupported event type: push when the workflow is triggered by a push event. The action never reaches Claude — it exits 1 before any prompt is
  run.

  This blocks any use case where Claude is driven from non-PR, non-issue automation (scheduled syncs, post-merge content generation, doc
  regeneration on push to default branch, etc.). workflow_dispatch, schedule, pull_request, and issue_comment all work; push does not.

  To Reproduce

  1. Create a workflow triggered by on: push: branches: [main] (or call a reusable workflow from such a trigger).
  2. In the workflow's job, invoke anthropics/claude-code-action@v1.0.160 with any valid prompt / claude_args / Bedrock or Anthropic credentials.
  3. Push a commit to main so the workflow runs.
  4. The "Run Claude Code Action" step fails immediately with:

  Error: Action failed with error: Unsupported event type: push
  Error: Process completed with exit code 1.

  A minimal reproducer:

  name: repro
  on:
    push:
      branches: [main]
  permissions:
    contents: read
    id-token: write
  jobs:
    claude:
      runs-on: ubuntu-latest
      steps:
        - uses: actions/checkout@v4
        - uses: anthropics/claude-code-action@v1.0.160
          with:
            use_bedrock: "true"
            claude_args: --max-turns 1
            prompt: "Say hello."

  Expected behavior

  push should be a supported event type — either:

  1. The action should treat push like workflow_dispatch (run the prompt, no PR/issue context needed), or
  2. The action should expose an explicit event_type / trigger_override input so callers can declare how they want the event to be interpreted,
  or
  3. At minimum, the docs should state which events are supported and the failure message should name the supported list so users can self-serve
  a fix.

  Today there is no way to drive the action from a push-triggered workflow without restructuring the trigger entirely.

  Screenshots

  N/A — full failure log included below.

  Job definition 

  - name: Generate ingestion doc
    uses: anthropics/claude-code-action@4633baf5267540f3f8cb58b684f79901d564c280 # v1.0.160
    with:
      github_token: ${{ steps.configure-credentials.outputs.github-token }}
      use_bedrock: "true"
      claude_args: ${{ steps.prepare-claude-args.outputs.args }}
      plugin_marketplaces: ${{ vars.PLUGIN_MARKETPLACE }}
      plugins: ${{ inputs.plugin }}
      prompt: |
        /${{ inputs.skill }} ${{ steps.prep.outputs.source-abs }}

        Asset short name: ${{ matrix.name }}
        Use this directory as OUTPUT_DIR (override the default): ${{ steps.prep.outputs.output-dir }}

        Do not write anywhere else. Do not post to GitHub. Final message must list WROTE/EMPTY lines per the skill spec.

  API Provider

  - [ ] Anthropic First-Party API (default)
  - [x] AWS Bedrock
  - [ ] GCP Vertex

  Additional context

  - Action version: anthropics/claude-code-action@4633baf5267540f3f8cb58b684f79901d564c280 (v1.0.160)
  - Runner: ubuntu-latest
  - Bun: 1.3.14+0d9b296af (cached install OK)
  - Event: push to main of a private repo, via a reusable workflow (workflow_call) — the parent event type propagates through reusable workflows,
  so the inner job still sees github.event_name == "push".
  - The action successfully completes Bun setup and bun install (140 packages installed). The failure happens immediately after, before any
  prompt execution, in the entrypoint that validates the event type.
  - The post-run "post-buffered-inline-comments" step also runs and prints No buffered inline comments, which is expected — PR_NUMBER is empty
  because the event isn't a PR. This suggests the action's design assumption is "we always have a PR/issue/comment to anchor to," and push (and
  presumably schedule?) don't fit that model.

  Full failure log:

  Run anthropics/claude-code-action@4633baf5267540f3f8cb58b684f79901d564c280
  ...
  Run cd ${GITHUB_ACTION_PATH}
  bun install v1.3.14 (0d9b296a)

  + @actions/core@1.11.1
  + @actions/github@6.0.1
  + @anthropic-ai/claude-agent-sdk@0.3.196
  + @modelcontextprotocol/sdk@1.16.0
  + @octokit/graphql@8.2.2
  + @octokit/rest@21.1.1
  + @octokit/webhooks-types@7.6.1
  + node-fetch@3.3.2
  + shell-quote@1.8.3
  + zod@3.25.76

  140 packages installed [1475.00ms]
  Run # Do NOT pass --tsconfig-override here. It triggers a Bun runtime bug
  Error: Action failed with error: Unsupported event type: push
  Error: Process completed with exit code 1.

  Use case: we run the action on push to main to regenerate ingestion docs from changed files in the merged commit, then upload them as artifacts
  for a downstream publish job. There is no PR at this point — the merge has already happened — so pull_request triggers don't fit.
  workflow_dispatch works but defeats the "regenerate on every merge" automation. Native push support (or an opt-in event override) would unblock
  this entirely.

tombrov:
## Summary

Adds `push` to the list of supported GitHub event types so the action can be driven from `on: push:` workflows (e.g. post-merge automation that regenerates artifacts after a merge to `main`).

Before this change, `parseGitHubContext()` threw `Unsupported event type: push` at startup, blocking any non-PR/non-issue automation use case where `workflow_dispatch` isn't a good fit.

## What changed

- `src/github/context.ts`
  - Import `PushEvent` from `@octokit/webhooks-types`.
  - Add `"push"` to `AUTOMATION_EVENT_NAMES` (same bucket as `workflow_dispatch`, `schedule`, `workflow_run`, `repository_dispatch`).
  - Add `PushEvent` to the `AutomationContext.payload` union.
  - Handle `case "push":` in the `parseGitHubContext()` switch.
- `test/github-context.test.ts`
  - New `push produces an automation context` test in the automation-events block.
  - Extend the `isEntityContext` / `isAutomationContext` type-guard assertions to cover push.

`isAutomationContext()` is derived from `AUTOMATION_EVENT_NAMES`, so adding `"push"` to that tuple makes the type guard recognize it automatically — no other code paths needed changes.

## Why this approach

- **No new input flag.** Push fits cleanly into the existing automation-event bucket. Agent mode is already triggered by the presence of a `prompt:` input, which is how `workflow_dispatch` / `schedule` users drive the action today.
- **No changes to `src/create-prompt/index.ts`.** The second `Unsupported event type` throw lives in tag mode's prompt builder, and tag mode requires an entity context (gated via `isEntityContext` in `detectMode()` and `run.ts`). Push events never reach that switch, so the throw stays as-is to preserve the tag-mode invariant.
- **Entity-only code paths are already gated.** `checkWritePermissions`, `restoreConfigFromBase`, and `updateCommentLink` in `run.ts` are all wrapped in `isEntityContext(context)` checks, so they correctly skip on push events.

## Behavior

With this change, a workflow like:

```yaml
on:
  push:
    branches: [main]
jobs:
  claude:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
        with:
          prompt: "Regenerate ingestion docs from the latest commit"
          claude_args: --max-turns 1
```

…now runs Claude in agent mode (no tracking comment, no PR/issue anchor required), matching how `workflow_dispatch` already works.

## Test plan

- [x] `bun run typecheck` passes
- [x] `bun test test/github-context.test.ts` passes (new push cases + existing automation events)
- [x] `bun test` full suite green
- [x] `bun run format:check` clean
- [x] Smoke test in a downstream repo: a workflow triggered by `on: push: branches: [main]` reaches Claude and runs the supplied prompt (instead of failing at startup)

Fixes #1456

### DIFF
--- a/src/github/context.ts
+++ b/src/github/context.ts
@@ -6,6 +6,7 @@ import type {
   PullRequestEvent,
   PullRequestReviewEvent,
   PullRequestReviewCommentEvent,
+  PushEvent,
   WorkflowRunEvent,
 } from "@octokit/webhooks-types";
 import { CLAUDE_APP_BOT_ID, CLAUDE_BOT_LOGIN } from "./constants";
@@ -65,6 +66,7 @@ const AUTOMATION_EVENT_NAMES = [
   "repository_dispatch",
   "schedule",
   "workflow_run",
+  "push",
 ] as const;
 
 // Derive types from constants for better maintainability
@@ -118,14 +120,15 @@ export type ParsedGitHubContext = BaseContext & {
   isPR: boolean;
 };
 
-// Context for automation events (workflow_dispatch, repository_dispatch, schedule, workflow_run)
+// Context for automation events (workflow_dispatch, repository_dispatch, schedule, workflow_run, push)
 export type AutomationContext = BaseContext & {
   eventName: AutomationEventName;
   payload:
     | WorkflowDispatchEvent
     | RepositoryDispatchEvent
     | ScheduleEvent
-    | WorkflowRunEvent;
+    | WorkflowRunEvent
+    | PushEvent;
 };
 
 // Union type for all contexts
@@ -245,6 +248,13 @@ export function parseGitHubContext(): GitHubContext {
         ...commonFields,
         eventName: "workflow_run",
         payload: context.payload as unknown as WorkflowRunEvent,
+      };
+    }
+    case "push": {
+      return {
+        ...commonFields,
+        eventName: "push",
+        payload: context.payload as unknown as PushEvent,
       };
     }
     default:
--- a/test/github-context.test.ts
+++ b/test/github-context.test.ts
@@ -297,6 +297,23 @@ describe("parseGitHubContext", () => {
       expect(context.eventName).toBe("workflow_run");
       expect(isAutomationContext(context)).toBe(true);
     });
+
+    test("push produces an automation context", () => {
+      setEvent("push", {
+        ref: "refs/heads/main",
+        before: "0000000000000000000000000000000000000000",
+        after: "1111111111111111111111111111111111111111",
+        repository: repositoryPayload,
+        pusher: { name: "test-actor", email: "test-actor@example.com" },
+      });
+
+      const context = parseGitHubContext();
+
+      expect(context.eventName).toBe("push");
+      expect(isAutomationContext(context)).toBe(true);
+      expect("entityNumber" in context).toBe(false);
+      expect("isPR" in context).toBe(false);
+    });
   });
 
   describe("invalid partition", () => {
@@ -442,6 +459,9 @@ describe("type guards", () => {
   const workflowDispatchContext = createMockAutomationContext({
     eventName: "workflow_dispatch",
   });
+  const pushContext = createMockAutomationContext({
+    eventName: "push",
+  });
 
   test("isIssuesEvent accepts only issues events", () => {
     expect(isIssuesEvent(issuesContext)).toBe(true);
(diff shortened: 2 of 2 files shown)


tombrov:
@wolffiex  @km-anthropic  I noticed it was pushed before and removed, would love seeing if this could be merged asap for the next release

Tested as a standalone in my actions and I can approve it does work

dovach (approved):

GuyBlassPapaya (approved):

diklamon (approved):

YuriSivPapaya (approved):

AgnePapaya (approved):

PR #1483, opened 2026-07-08
fix: handle null author (deleted/ghost accounts) in GitHub data formatter

HumphreySun98:
## What

`formatContext`, `formatComments`, and `formatReviewComments` (`src/github/data/formatter.ts`) crash with `TypeError: null is not an object (evaluating 'author.login')` whenever an issue, PR, comment, or review involved in a run was written by a **deleted GitHub account**.

## Why it happens

GitHub's GraphQL API types `author` as the nullable `Actor` interface and returns `null` for content authored by deleted ("ghost") accounts. But `src/github/types.ts` declared `author: GitHubAuthor` (non-null) on `GitHubComment`, `GitHubReview`, `GitHubPullRequest`, and `GitHubIssue`, so the formatter dereferenced `author.login` at four sites (`formatter.ts:24/36/74/88`) with no guard.

A single such comment/review aborts the whole prompt-generation path (`create-prompt` calls `formatComments` / `formatContext` / `formatReviewComments`) and fails the run. `filterCommentsByActor` returns early when no actor filters are configured (the default), so nothing upstream catches the null before it reaches the formatter.

## Fix

- **Make the type honest:** `author: GitHubAuthor | null` on the four GraphQL types, with a comment explaining GitHub's ghost-account behavior. This makes `tsc` enforce null-handling from here on (it surfaced exactly the dereference sites below and no others).
- **Guard every dereference:** `author?.login ?? "ghost"` in `formatter.ts`, and `author?.login ?? ""` in `filterCommentsByActor` (so a ghost author matches no include/exclude username pattern instead of throwing). `"ghost"` mirrors the login GitHub itself shows for deleted accounts.

## Tests

Added a `null author (deleted/ghost GitHub accounts)` suite to `test/data-formatter.test.ts` covering all four call sites (PR context, issue context, comment, review). Each case throws `TypeError` before the fix and renders the `ghost` fallback after.

## Verification

- `bun test` → **773 pass, 0 fail** (4 new).
- `bun run typecheck` → clean (the now-nullable type surfaced no other unguarded dereference).
- `bun run format:check` → clean.

Happy to open a tracking issue as well if the maintainers prefer.

### DIFF
--- a/src/github/data/fetcher.ts
+++ b/src/github/data/fetcher.ts
@@ -204,11 +204,9 @@ export function isBodySafeToUse(
  * @param excludeActors - Comma-separated actors to exclude
  * @returns Filtered array of comments
  */
-export function filterCommentsByActor<T extends { author: { login: string } }>(
-  comments: T[],
-  includeActors: string = "",
-  excludeActors: string = "",
-): T[] {
+export function filterCommentsByActor<
+  T extends { author: { login: string } | null },
+>(comments: T[], includeActors: string = "", excludeActors: string = ""): T[] {
   const includeParsed = parseActorFilter(includeActors);
   const excludeParsed = parseActorFilter(excludeActors);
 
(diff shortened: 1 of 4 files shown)


ashwin-ant:
Thanks for this. #1490 fixed the same deleted-account crash and just merged, so closing this one as a duplicate.

HumphreySun98:
> Thanks for this. #1490 fixed the same deleted-account crash and just merged, so closing this one as a duplicate.

Thank you so much for your review.

Issue #1137, opened 2026-04-01
Branch name validation rejects valid git branch names containing #


pbosch:
## Bug

The branch name validation regex rejects `#`, which is a valid character in git branch names. This causes the action to fail on PRs from branches like `put-back-arm64-#2` with:

```
##[error] Invalid branch name: "put-back-arm64-#2". Branch names must start with an
alphanumeric character and contain only alphanumeric characters, forward slashes,
hyphens, underscores, or periods.
```

## Context

The validation at `src/github/operations/branch.ts` uses regex `^[a-zA-Z0-9][a-zA-Z0-9/_.-]*$` which excludes `#`. Per [git-check-ref-format](https://git-scm.com/docs/git-check-ref-format), `#` is allowed in branch names.

This validation makes sense for branches the action *creates*, but for `pull_request` events where the action only *reads* an existing branch (checked out by `actions/checkout`), there's no reason to reject it. The branch already exists in git — rejecting it just prevents the review from running.

## Expected behavior

For `pull_request` events (review mode), the action should accept whatever branch name GitHub provides via the event context, since it's already a valid git ref.

## Reproduction

1. Create a branch with `#` in the name: `git checkout -b "fix-arm64-#2"`
2. Push and open a PR
3. Run `claude-code-action` on the PR → fails with the error above

## Suggested fix

Either:
- Relax the regex to allow `#` and other git-valid characters
- Skip branch name validation for `pull_request` events where the action is reading (not creating) branches
- Add an input to override/skip branch validation

PR #1486, opened 2026-07-09
fix: allow leading underscore in branch names (valid per git-check-ref-format)

riley-mete-db:
## Problem

`validateBranchName()` rejects branch names that start with an underscore, even though they are valid per `git check-ref-format`:

```
Error: Action failed with error: Invalid branch name: "_release/v0.4.15". Branch names must start with an alphanumeric character or '@' and contain only alphanumeric characters, forward slashes, hyphens, underscores, periods, hashes (#), plus signs (+), commas (,), or at signs (@).
```

Underscores are already allowed everywhere in the name *except* the first character. Because `setupBranch()` validates the PR's `baseRefName` after checking out an open PR (`src/github/operations/branch.ts`), the action fails on **every open PR targeting a branch like `_release/v1.2.3`** — a common convention for release/internal branches. There is no input that bypasses this path.

## Fix

Add `_` to the first-character class of the whitelist pattern:

```diff
-  const validPattern = /^[a-zA-Z0-9@][a-zA-Z0-9/_.#+,@-]*$/;
+  const validPattern = /^[a-zA-Z0-9@_][a-zA-Z0-9/_.#+,@-]*$/;
```

A leading underscore carries no option-injection risk — only a leading dash does, and that is still rejected separately. All git calls already go through `execFileSync`, so there is no shell-interpolation risk either. The error message and doc comments are updated to match.

This follows the same pattern as prior fixes relaxing this whitelist for other git-valid characters: `#` (#1137), `@` (#998), `+` (#1248), `,` (#1310).

## Testing

- Added cases to `test/validate-branch-name.test.ts` covering `_release/v1.2.3`, `_internal`, and `_wip/feature-x`
- `bun test`: 770 pass, 0 fail
- `bun run typecheck`: clean
- `prettier --check` on changed files: clean

### DIFF
--- a/src/github/operations/branch.ts
+++ b/src/github/operations/branch.ts
@@ -27,7 +27,7 @@ function extractFirstLabel(githubData: FetchDataResult): string | undefined {
  * This prevents command injection by ensuring only safe characters are used.
  *
  * Valid branch names:
- * - Start with alphanumeric character or @ (not dash, to prevent option injection)
+ * - Start with alphanumeric character, underscore, or @ (not dash, to prevent option injection)
  * - Contain only alphanumeric, forward slash, hyphen, underscore, period, hash (#), plus (+), comma (,), or at sign (@)
  * - Do not start or end with a period
  * - Do not end with a slash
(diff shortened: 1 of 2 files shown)


riley-mete-db:
Hey @ashwin-ant, could you please review this when you get a chance?

It's similar to #1411, which you reviewed and merged a few weeks ago: another git-valid character added to the branch name whitelist, in this case a leading `_`. I updated the regex's first-character class, added test cases for `_release/v1.2.3` style names, and verified the full suite passes (770 tests, typecheck and prettier clean).

Thanks!

ashwin-ant (approved):

PR #1497, opened 2026-07-11
feat: support ref pinning for plugin_marketplaces via .git#<ref> suffix

### LINKED ISSUE #1229: Support version pinning for plugin_marketplaces
## Summary

I'd like the `plugin_marketplaces` input parameter to support pinning a specific version of the marketplace repository.

## Motivation

Currently, `plugin_marketplaces` only accepts URLs in the form `https://github.com/user/marketplace.git` and always clones the HEAD of the default branch.

This means that whenever the marketplace repository is updated, compatibility with downstream repositories using the action can break. Just as GitHub Actions itself supports version pinning with `uses: actions/checkout@v6`, plugin marketplaces should also support version pinning.

Claude Code already supports version pinning via the `ref` field in `extraKnownMarketplaces` in `settings.json`:

```json
{
  "extraKnownMarketplaces": {
    "konoka": {
      "source": {
        "source": "github",
        "repo": "ncaq/konoka",
        "ref": "v5.0.0"
      }
    }
  }
}
```

Having equivalent functionality in claude-code-action's `plugin_marketplaces` would enable consistent version management between the CLI and the Action.

### Option 1: Fragment URL support (lightweight approach)

Accept fragment-style URLs like `https://github.com/user/marketplace.git#v5.0.0`.

Currently, `MARKETPLACE_URL_REGEX` in `base-action/src/install-plugins.ts` requires the URL to end with `.git`, so any content after `.git` causes a validation error:

```typescript
// Current
const MARKETPLACE_URL_REGEX = /^https:\/\/[a-zA-Z0-9\-._~:/?#[\]@!$&'()*+,;=%]+\.git$/;
```

The fix would be to allow an optional fragment part and parse it to pass to `git clone --branch <ref>` or similar.

### Option 2: Structured ref support aligned with settings.json (comprehensive approach)

Support structured marketplace definitions, similar to how `extraKnownMarketplaces` works in settings.json. For example:

```yaml
plugin_marketplaces: |
  - url: https://github.com/user/marketplace.git
    ref: v5.0.0
  - url: https://github.com/other/marketplace.git
    ref: v2.1.0
```

Or a simpler key-value style:

```yaml
marketplace_refs: |
  konoka=v5.0.0
```

## Use case

```yaml
# Option 1
  with:
    plugin_marketplaces: "https://github.com/user/marketplace.git#v5.0.0"
    plugins: "my-plugin@marketplace"

# Option 2
  with:
    plugin_marketplaces: |
      - url: https://github.com/user/marketplace.git
        ref: v5.0.0
    plugins: "my-plugin@marketplace"
```

onurcelep:
Implements **Option 1 (fragment URL)** proposed in #1229.

## What

`plugin_marketplaces` entries may now append `#<branch|tag|sha>` after the `.git` suffix:

```yaml
plugin_marketplaces: |
  https://github.com/user/marketplace.git#v1.2.3
  https://github.com/other/marketplace.git
```

Pinned entries are fetched at that ref into a temp directory and added as a **local** marketplace (a path the action already supports); unpinned URLs take exactly the previous code path — fully backward-compatible.

## Why

#1229 covers the compatibility motivation. The sharper version of the argument: CI is the *unattended* environment running with workflow-granted permissions, and today it is the only one that cannot pin — Claude Code's `extraKnownMarketplaces` already supports `ref` (branch/tag) on marketplace sources in `settings.json` for local sessions. Every run currently re-fetches whatever is on the marketplace's default branch at that moment, which is the same third-party-code-in-CI exposure that `uses: owner/action@sha` pinning exists to prevent, one layer down. This PR brings the action input to parity with the settings format — and, because the action performs the fetch itself, it can additionally honor exact commit SHAs, which marketplace sources in settings cannot.

## Implementation notes

- **Fetch-by-ref instead of `git clone --branch`**: the pinned checkout uses `init → remote add → fetch --depth 1 origin <ref> → checkout FETCH_HEAD`, because `--branch` accepts only branch/tag names — this path uniformly handles branches, tags, **and unadvertised commit SHAs** (verified live against github.com: fetching a non-tip commit SHA succeeds and yields a valid marketplace checkout). SHA pinning is the security-grade variant, so it's supported from the start.
- **Ref validation**: must start with an alphanumeric (a ref can never be parsed as a git flag), `..` rejected, length-capped; all subprocesses use argument arrays, no shell involved.
- `.git#` is unambiguous as a separator since valid marketplace URLs must end in `.git`.

## Tests

8 new cases in `base-action/test/install-plugins.test.ts` following the existing spawn-spy conventions: full pinned-flow argument assertions, SHA pinning, mixed pinned/unpinned lists, git-failure surfacing, and rejection of empty refs, flag injection (`#-rev`), path traversal (`#v1..v2`), and shell metacharacters. Full suite: **777 pass / 0 fail**; `tsc --noEmit` clean; prettier applied (the `docs/usage.md` table realignment is prettier's, from the one changed row).

Closes #1229


onurcelep:
Adding a concrete data point for reviewers, since this has been quiet since July.

I run a small fleet of repos whose PR review workflow loads a plugin through
plugin_marketplaces. Because the input accepts only a bare .git URL, every run
re-fetches the marketplace's default branch at that moment. This week the
same workflow file went green at 12:24 and red at 12:39 on the same repo with
no change on my side; I had no way to hold the plugin still while I found out
why. For the Action itself I can write uses: owner/action@sha; one layer down
there is no equivalent, and CI is the one unattended environment running
with workflow-granted permissions.

#1229 asks for exactly this, and this PR implements its option 1 without
touching the unpinned code path. Happy to rebase or adjust the syntax if the
maintainers prefer a separate input over the #<ref> fragment.

PR #1498, opened 2026-07-13
fix(branch): sanitize {{label}} in branch_name_template for git-safe names

### LINKED ISSUE #1491: Unsanitized {{label}} in branch_name_template can abort the run with an invalid branch name
When `branch_name_template` uses `{{label}}`, the raw GitHub label name is substituted into the branch name without any sanitization. Labels are free-form and frequently scoped with a colon or slash, so a template like `{{prefix}}{{label}}/{{description}}` on an issue whose first label is `area:permissions` produces a branch name containing `:`.

That name then reaches `validateBranchName` in `src/github/operations/branch.ts`, which rejects `:` (along with the other special git characters). The error is caught by the surrounding `try/catch` in the branch setup block, which calls `process.exit(1)` — so instead of a bad-branch-name warning, the whole action run dies.

This is easy to hit because scoped labels are common — this repo itself uses labels like `area:permissions` and `provider:1p`. Any consumer that opts into a `{{label}}`-based template and runs on an issue carrying such a label will crash.

The inconsistency is with `{{description}}`, which is already sanitized: `generateBranchName` runs the title through `extractDescription`, which kebab-cases it and strips everything outside `[a-z0-9-]`. `{{label}}` is the only free-text template variable that skips that step — it's substituted verbatim. `action.yml` even documents `{{description}}` as being kebab-cased but says nothing about `{{label}}`, so the raw substitution isn't the intended contract.

Repro: set `branch_name_template: "{{prefix}}{{label}}/{{entityNumber}}"` and trigger the action on an issue whose first label is `area:permissions`. The generated branch name is `claude/area:permissions/123`, `validateBranchName` throws, and the run exits with code 1.

Fix is to sanitize the label into a git-safe segment before substitution, the same way descriptions are handled. I'll open a PR that does this and adds regression tests.

(Related, out of scope for the fix I'm sending: a title made up entirely of non-alphanumerics makes `{{description}}` sanitize to an empty string, so a template like `{{prefix}}{{description}}/{{entityNumber}}` yields a `//` double slash, which `validateBranchName` also rejects. Same failure mode, different variable — worth a follow-up.)

syf2211:
## Summary

Sanitize GitHub label values before substituting them into `branch_name_template`, so scoped labels like `area:permissions` no longer produce branch names containing `:` that fail `validateBranchName` and abort the action run.

## Motivation

Fixes #1491. Labels are free-form and often scoped with `:` or `/` (e.g. `area:permissions`, `provider:bedrock`). The `{{description}}` variable was already normalized to kebab-case, but `{{label}}` was substituted verbatim. A template like `{{prefix}}{{label}}/{{entityNumber}}` could crash the entire run with `process.exit(1)`.

## Changes

- `src/utils/branch-template.ts`: add shared `sanitizeBranchSegment()` helper; route `extractDescription()` and new `sanitizeLabel()` through it
- `generateBranchName()`: substitute sanitized label, falling back to `entityType` when sanitization yields empty
- `test/branch-template.test.ts`: regression tests for scoped labels, colon-containing labels, and empty-sanitize fallback; assert `validateBranchName` passes

## Tests

- `bun test test/branch-template.test.ts` — pass (23/23)
- `bun test` — pass (772/772)
- `bun run typecheck` — pass
- composer-2.5 review: COMMENT (addressed JSDoc + fallback test)

## Notes

- Intentionally scoped to `{{label}}` only; custom templates with literal punctuation in `{{prefix}}` are unchanged (pre-existing behavior).
- Labels that sanitize to empty (e.g. `:::`) fall back to `entityType`, matching the documented fallback when no label is provided.

Fixes #1491

### DIFF
--- a/src/utils/branch-template.ts
+++ b/src/utils/branch-template.ts
@@ -7,7 +7,27 @@
 const NUM_DESCRIPTION_WORDS = 5;
 
 /**
+ * Normalizes a string into a git-safe lowercase hyphenated segment.
+ */
+function sanitizeBranchSegment(value: string): string {
+  if (!value || value.trim() === "") {
+    return "";
+  }
+
+  return value
+    .trim()
+    .toLowerCase()
+    .replace(/[:/\s]+/g, "-") // Scoped tokens (area:permissions) or path-like segments
+    .replace(/[^a-z0-9-]/g, "") // Remove non-alphanumeric except hyphens
+    .replace(/-+/g, "-") // Replace multiple hyphens with single
+    .replace(/^-|-$/g, ""); // Remove leading/trailing hyphens
+}
+
+/**
  * Extracts the first 5 words from a title and converts them to kebab-case
+ */
+/**
+ * Extracts the first 5 words from a title and converts them to kebab-case.
  */
 function extractDescription(
   title: string,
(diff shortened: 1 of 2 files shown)


ashwin-ant:
Thanks. #1492 fixes the same {{label}} sanitization issue and merged first, so closing this as a duplicate.

Issue #1482, opened 2026-07-08
Feature request: expose per-run cost and token usage as Action outputs


hudaif-eg:
## Context

We're evaluating Claude Code cost across models in GitHub Actions. After inspecting the Action source and testing the released `anthropics/claude-code-action@v1`, we couldn't find a supported way to retrieve per-run usage metrics.

## Our investigation found

- The Action completes successfully but does not expose cost/token metrics as outputs.
- The Claude CLI is not available after the Action, so `/usage` and `/cost` cannot be called in later workflow steps.
- `show_full_output: true` did not emit usage information in our logs.

## Ask

Is there a supported way to retrieve per-run metrics such as:

- Total cost (USD)
- Input/output tokens
- Cache read/write tokens

If not, would you consider exposing these as official Action outputs if the underlying SDK provides them?

Proposed output names (for discussion):

```
total_cost_usd
input_tokens
output_tokens
cache_read_tokens
cache_write_tokens
```

This would be valuable for teams monitoring API spend per PR review run without needing to correlate against the Anthropic Console manually.

PR #1506, opened 2026-07-14
feat(comment): show cost alongside duration in Claude's tracking comment

### LINKED ISSUE #59: Cost report

kenchan:
## Summary
Adds the run's cost next to its duration in Claude's tracking comment on issues/PRs, e.g.:

> **Claude finished @user's task in 31s ($0.1234)**

## Why
`total_cost_usd` is already computed for every run and read into `executionDetails` in `update-comment-link.ts` — it just isn't rendered where people actually look. The tracking comment shows duration but not cost; cost only appears in the Actions Step Summary, which most people reading a PR/issue never open. Surfacing it in the comment makes cost visible to everyone in the loop, and brings the comment to parity with the Step Summary (`format-turns.ts`), which already displays it. This directly answers #59 ("Is there a way to make claude report the cost after a run?"), and speaks to the same cost-visibility need behind #1482.

## Changes
- `src/github/operations/comment-logic.ts`: compute `costStr` from `executionDetails.total_cost_usd` (same `$X.XXXX` formatting already used in the Step Summary) and append it in parentheses after the duration, in both the success and error headers.
- `test/comment-logic.test.ts`: cover duration-only, cost-only, and combined duration+cost cases for both success and error headers.

## Notes
Cost is shown unconditionally (no opt-in flag), matching how duration is already handled and keeping the diff minimal. The number is already visible in the Step Summary on public repos, so this doesn't expose anything new — happy to gate it behind an input if you'd prefer that.

## Test plan
- [x] `bun run typecheck`
- [x] `bun test test/comment-logic.test.ts` — 26 pass
- [x] `bun run format:check`
- [x] Full `bun test` suite — no new failures (2 pre-existing failures in `install-pipefail.test.ts` are network-dependent and unrelated to this change)

Closes #59
Related: #1482

### DIFF
--- a/src/github/operations/comment-logic.ts
+++ b/src/github/operations/comment-logic.ts
@@ -109,6 +109,12 @@ export function updateCommentBody(input: CommentUpdateInput): string {
     durationStr = minutes > 0 ? `${minutes}m ${seconds}s` : `${seconds}s`;
   }
 
+  // Calculate cost string if available
+  let costStr = "";
+  if (executionDetails?.total_cost_usd !== undefined) {
+    costStr = `$${executionDetails.total_cost_usd.toFixed(4)}`;
+  }
+
   // Build the header
   let header = "";
 
@@ -116,6 +122,9 @@ export function updateCommentBody(input: CommentUpdateInput): string {
     header = "**Claude encountered an error";
     if (durationStr) {
       header += ` after ${durationStr}`;
+    }
+    if (costStr) {
+      header += ` (${costStr})`;
     }
     header += "**";
   } else {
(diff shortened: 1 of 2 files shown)


kenchan:
Would love to get a maintainer's take on whether there's room for this to land (as-is, or gated behind an opt-in input as suggested in the PR description).

In the meantime, for anyone who wants this today without waiting on the action itself: the `execution_file` output already exposes `total_cost_usd`/`duration_ms` per run, so you can add a follow-up step that parses it and posts the cost as a separate comment, e.g.:

```yaml
- name: Comment cost
  if: always() && steps.claude.outputs.execution_file != ''
  run: |
    cost=$(jq -r 'last | select(.type=="result") | .total_cost_usd // empty' "${{ steps.claude.outputs.execution_file }}")
    [ -n "$cost" ] && gh api "repos/${{ github.repository }}/issues/${{ github.event.issue.number || github.event.pull_request.number }}/comments" \
      -f body="💰 Cost: \$$(printf '%.4f' "$cost")"
  env:
    GH_TOKEN: ${{ github.token }}
```

Not as clean as having it inline in the tracking comment, but it works without forking the action.

PR #1525, opened 2026-07-17
fix: keep the triggering review/comment in prompt context (#1522)

### LINKED ISSUE #1522: pull_request_review trigger: the triggering review is always filtered out of the prompt context (submittedAt >= triggerTime)
### Summary

When a workflow is triggered by `pull_request_review` (submitted), the review that triggered the run is always excluded from the prompt context: `<review_comments>` renders as "No review comments" even though the triggering review contains inline comments. If the review was submitted with an empty body (very common — reviewers often only leave inline comments and hit "Submit review"), Claude receives **no information at all** about the review it is supposed to respond to, and typically ends up asking the reviewer to re-post their comments with an explicit `@claude` mention.

### Root cause

`extractTriggerTimestamp()` returns `review.submitted_at` for `pull_request_review` events ([src/github/data/fetcher.ts](https://github.com/anthropics/claude-code-action/blob/main/src/github/data/fetcher.ts)):

```ts
} else if (isPullRequestReviewEvent(context)) {
    return context.payload.review.submitted_at || undefined;
```

`filterReviewsToTriggerTime()` then drops every review with `submittedAt >= triggerTimestamp`:

```ts
// Review must have been submitted before trigger (not at or after)
const submittedTimestamp = new Date(review.submittedAt).getTime();
if (submittedTimestamp >= triggerTimestamp) {
  return false;
}
```

The triggering review's `submittedAt` **equals** the trigger timestamp, so the `>=` comparison always filters out the very review the run is about — together with all of its inline comments. The anti-injection intent of the filter (drop content created/edited *after* the trigger) is sound, but the entity *at* the trigger timestamp is the trigger itself, not post-trigger injection.

The review **body** is separately injected as `<trigger_comment>` (from the webhook payload), which masks the bug when reviewers write a summary text — but reviews that consist only of inline comments produce a completely empty context.

The same `>=` pattern exists in `filterCommentsToTriggerTime()`, which affects `pull_request_review_comment` triggers analogously (the triggering comment is dropped from the fetched context; there the payload-injected `<trigger_comment>` carries the text, so the impact is smaller).

### Observed behavior (v1 @ `1298632c`, still present on `main` @ `3e807ec`)

Workflow: `pull_request_review` (types: submitted) → `anthropics/claude-code-action@v1` with a custom `prompt` and `track_progress: true`. Reviewer submits a review containing one inline comment and an empty body. Action log:

```
Mode: tag
Preparing with mode: tag for event: pull_request_review
...
Successfully fetched PR #118 data
...
===== FINAL PROMPT =====
...
<comments>
No comments
</comments>

<review_comments>
No review comments
</review_comments>
```

The inline comment (created three minutes *before* submission, `submitted_at` equal to the trigger time) is verifiably present via `GET /repos/{owner}/{repo}/pulls/118/reviews/{id}/comments`.

### Suggested fix

In `filterReviewsToTriggerTime()` / `filterCommentsToTriggerTime()`, keep the entity that *is* the trigger:

- compare creation/submission time with `>` instead of `>=` (a same-second non-trigger injection is indistinguishable anyway and would still be caught by the edited-after-trigger check), **or**
- explicitly retain the review/comment whose id matches the triggering payload entity (most precise), while keeping the `lastEditedAt/updatedAt >= trigger` check as-is. Note the edit-check needs the same exemption for the trigger entity if GitHub bumps `updated_at` to the submission time on submit.

### Workaround for others hitting this

A pre-step that fetches the triggering review + its inline comments via `gh api` into files under `$RUNNER_TEMP`, with the `prompt` pointing Claude at those files (file handoff avoids interpolating untrusted comment text into workflow YAML).

TemRevil:
Fixes #1522.

## Problem

`filterReviewsToTriggerTime()` / `filterCommentsToTriggerTime()` drop any entity whose submit/creation time is `>= triggerTime`, to keep content injected at/after the trigger out of the prompt (TOCTOU protection). But the entity that *is* the trigger is submitted/created **exactly** at `triggerTime`, so `>=` always filtered it out.

For a `pull_request_review` trigger this dropped the triggering review **and all of its inline comments** (they hang off the review node, which was removed). When a reviewer submits only inline comments with an empty body, `<review_comments>` renders "No review comments" and Claude receives nothing about the review it is supposed to respond to.

The edit-time check compounds it: reviews fetch `updatedAt`, which GitHub stamps at submit time, so even relaxing the submit comparison alone wouldn't be enough.

## Fix

Retain the entity whose `databaseId` matches the triggering webhook payload id, exempting **only that one entity** from both the submit/create and edit-time checks. The `>=` comparison is left untouched for every other entity, so post-trigger injection is still filtered — this is deliberately narrower than relaxing `>=` to `>`, which would admit *any* same-second entity.

- New `extractTriggerEntityId(context)` returns `review.id` / `comment.id` from the payload (mirrors `extractTriggerTimestamp`).
- Threaded through `fetchGitHubData` to both filters (covers `pull_request_review`, `pull_request_review_comment`, and `issue_comment` triggers).
- `databaseId` is string-compared: the GraphQL id is a `string`, the webhook id is a `number`.

## Tests

Added 4 tests to `test/data-fetcher.test.ts`:
- the triggering review is dropped without the id and **kept** when the id is passed;
- it's kept even when `updatedAt` is bumped to the submit time (edit-check bypass);
- a *non-trigger* review submitted at the same time is **still dropped** (the exemption is precise);
- the triggering comment created exactly at trigger time is kept.

`bun test test/data-fetcher.test.ts` → 80 pass / 0 fail. `bun run typecheck` clean. Only the two touched source files + the test are changed.

### DIFF
--- a/src/github/data/fetcher.ts
+++ b/src/github/data/fetcher.ts
@@ -181,6 +181,30 @@ async function findIssueEventTime(
 }
 
 /**
+ * Extracts the numeric id of the entity that triggered the run (the comment or
+ * review from the webhook payload). It matches the GraphQL `databaseId` of the
+ * fetched comment/review, letting the trigger-time filters retain the entity
+ * that *is* the trigger — which is created/submitted exactly at the trigger
+ * time and would otherwise be dropped by the `>= triggerTime` comparison.
+ *
+ * @param context - Parsed GitHub context from webhook
+ * @returns The numeric entity id, or undefined if the event has no such entity
+ */
+export function extractTriggerEntityId(
+  context: ParsedGitHubContext,
+): number | undefined {
+  if (isIssueCommentEvent(context)) {
+    return context.payload.comment.id;
+  } else if (isPullRequestReviewEvent(context)) {
+    return context.payload.review.id;
+  } else if (isPullRequestReviewCommentEvent(context)) {
+    return context.payload.comment.id;
+  }
+
+  return undefined;
+}
+
+/**
  * Extracts the original title from the GitHub webhook payload.
  * This is the title as it existed when the trigger event occurred.
  *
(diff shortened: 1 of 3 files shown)


TemRevil:
Rebased onto main, the conflict was from #1592 touching this same file for an unrelated timestamp feature. Trigger-entity-id fix is back on top, typecheck and the data-fetcher/tag suites are green.

TemRevil:
@ashwin-ant #1526 went in on the 20th, this one has been sitting since July without a look. Still rebased and green as of the 7th, happy to redo it if it has drifted since.

PR #1526, opened 2026-07-17
fix: neutralize the checkout credential under actions/checkout v6+ include layout (#1510)

### LINKED ISSUE #1510: Checkout credential removal in configureGitAuth is a no-op since actions/checkout v6
## Summary

`configureGitAuth` removes the credential persisted by `actions/checkout` before installing the action's own app token ([src/github/operations/git-config.ts](https://github.com/anthropics/claude-code-action/blob/v1.0.169/src/github/operations/git-config.ts)):

```ts
await $`git config --unset-all http.${GITHUB_SERVER_URL}/.extraheader`;
```

Since actions/checkout v6.0.0 ("Persist creds to a separate file", actions/checkout#2286, backported in v5.0.1 and v4.3.1), the persisted header is no longer stored in the repository's `.git/config`. It is written to a separate config file under `RUNNER_TEMP` and referenced from the repository config via an include. `git config --unset-all` operates on the local config file and cannot remove values supplied by includes, so the removal is a no-op: the action logs `No existing authentication headers to remove` and the checkout credential (the workflow `GITHUB_TOKEN`) remains active for all git operations for the rest of the job.

## Consequences

1. **The workflow `GITHUB_TOKEN` stays usable by git while Claude executes.** With the default `persist-credentials: true` and a workflow granting `contents: write`, `git push` during Claude's execution authenticates with the workflow token even after the action replaced the origin URL with its own short-lived, scoped app token. The design intent of the removal (only the app token available during execution) is not achieved, and the success-looking log line is misleading during a security review.

2. **It masks #1236.** Tag mode's `setupBranch` fetches before `configureGitAuth` runs, which only works on default setups because the checkout credential survives. Users who set `persist-credentials: false` — the mitigation recommended by zizmor's `artipacked` audit — get an immediate failure.

## Verification

Two runs of the same workflow shape (two invocations of this action in one job), both on actions/checkout v7 and claude-code-action v1.0.169:

- `persist-credentials: true`: both invocations log `No existing authentication headers to remove`; the second invocation's `setupBranch` fetch succeeds even though the only credential in the origin URL was already revoked by the first invocation (#1509) — i.e. the surviving checkout header is what authenticates.
- `persist-credentials: false`: the second invocation fails in `setupBranch` with `remote: Invalid username or token` (#1509).

## Suggested direction

Removing an include-provided credential from inside the repository config is not possible with `--unset-all`, so either:

- treat `persist-credentials: false` as the supported configuration: make no fetch depend on ambient credentials (#1236, PR #1340) and document the recommendation; or
- neutralize the checkout credential through a mechanism that works with the include-based storage (e.g. removing the include entry that checkout adds to the repository config).

## Version

claude-code-action v1.0.169 (code unchanged on current `main`); actions/checkout v6.0.0+, v5.0.1, v4.3.1.

TemRevil:
Fixes #1510.

## Problem

`configureGitAuth()` removes the credential that `actions/checkout` persists with:

```ts
git config --unset-all http.<server>/.extraheader
```

That edits only the repo-local config. Since **actions/checkout v6.0.0** ([#2286](https://github.com/actions/checkout/pull/2286), backported to v5.0.1 / v4.3.1) the auth header is no longer stored in `.git/config` — it's written to a separate file under `RUNNER_TEMP` and pulled in via `include.path`. `git config --unset-all` cannot remove an include-provided value, so the call is a **no-op**: the action logs `No existing authentication headers to remove` while the checkout credential (typically the workflow `GITHUB_TOKEN`) stays usable by git for the rest of the job — exactly while Claude executes.

## Fix

Clear the header from the local config **and** from every file referenced by `include.path`, so it's neutralized under both layouts:

- pre-v6 (header in local config) → `--unset-all` on the local config (unchanged);
- v6+ (header in an included file) → `git config --file <inc> --unset-all …` on each include.

Includes that don't define the header are left untouched (a no-op `--unset-all` there), so legitimate includes survive.

## Verification

Reproduced the include-based layout with plain git and confirmed the behavior, since git operations aren't unit-tested in this repo (integration lives in `install-test`):

| Case | Before | After fix |
|---|---|---|
| v6+ credential behind `include.path` | `--unset-all` exits 5, header still resolves | header **gone** |
| pre-v6 credential in local config | removed | removed (unchanged) |
| legit include (no header) present | — | **preserved** (include kept, its settings intact) |
| mixed cred-include + legit-include | — | cred **gone**, legit **preserved** |
| no auth configured at all | — | no error (idempotent) |

`bun run typecheck` clean; touched file is prettier-clean. (The repo-wide `format:check` only flags files due to a Windows CRLF checkout, not this change.)

### DIFF
--- a/src/github/operations/git-config.ts
+++ b/src/github/operations/git-config.ts
@@ -56,6 +56,16 @@ export async function configureGitAuth(
  * action's own token instead (a credential helper when non-write users are
  * allowed, otherwise the origin URL). This applies to every mode, including API
  * commit signing where no other git configuration is needed.
+ *
+ * actions/checkout < v6 stored the header directly in the repo-local config,
+ * where `git config --unset-all` removes it. Since v6.0.0 (backported to
+ * v5.0.1 and v4.3.1) the header is written to a separate file under
+ * RUNNER_TEMP that the repo config pulls in via `include.path`; `--unset-all`
+ * on the local config cannot touch an include-provided value, so the removal
+ * was a silent no-op and the checkout credential (typically the workflow
+ * GITHUB_TOKEN) stayed usable by git for the rest of the job. Clear the
+ * header from the local config AND from every included file so it can no
+ * longer authenticate while Claude runs.
  */
 export async function replaceCheckoutCredentials(
   githubToken: string,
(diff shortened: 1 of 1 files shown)


TemRevil:
Rebased onto main, the conflict was from #1597 which refactored this function for an unrelated commit-signing case. Fix is reapplied on top of the refactor, typecheck and the git-config tests pass (except the one pre-existing Windows file-mode assertion, confirmed unrelated by reproducing it against a clean upstream checkout).

ashwin-ant (approved):

TemRevil:
@ashwin-ant this got approved on the 14th but looks like it never went in with the rest of that batch. Anything you need from me? Can rebase if it has drifted.

TemRevil:
Thanks @ashwin-ant, appreciate it.

PR #1528, opened 2026-07-19
fix: normalize empty branch template segments

### LINKED ISSUE #1527: branch_name_template with empty {{description}} produces '//' and aborts the run
### Summary

When a `branch_name_template` places `{{description}}` next to a slash and the entity's title contains no ASCII-alphanumeric characters (emoji-only, CJK-only, or punctuation-only titles), `{{description}}` sanitizes to an empty string. The template then yields consecutive slashes (e.g. `claude//123`), which `validateBranchName` rejects — aborting the entire run with an unhandled error.

This is **distinct from #1491**. That issue (and its open PRs #1492 / #1498) sanitizes `{{label}}`; both deliberately leave `{{description}}` untouched on the reasoning that it is "already sanitized." It is sanitized *character-wise*, but it can still be **empty**, which is what triggers this crash. So this remains broken after either label PR merges. The original reporter of #1491 flagged this as a follow-up.

### Reproduction

Config:

```yaml
branch_name_template: "{{prefix}}{{description}}/{{entityNumber}}"
```

Trigger the action on an issue/PR whose title has no `[a-z0-9]` content, e.g. `🎉🎉🎉`, `日本語のタイトル`, or `!!! ???`.

Minimal standalone repro against the current source:

```ts
import { generateBranchName } from "./src/utils/branch-template";
import { validateBranchName } from "./src/github/operations/branch";

const template = "{{prefix}}{{description}}/{{entityNumber}}";
const branch = generateBranchName(template, "claude/", "issue", 123, undefined, undefined, "🎉🎉🎉");
console.log(branch);        // "claude//123"
validateBranchName(branch); // throws: Branch names cannot contain consecutive slashes
```

Output:

```
claude//123
error: Invalid branch name: "claude//123". Branch names cannot contain consecutive slashes.
```

### Expected behavior

A title that happens to be all-emoji/CJK/punctuation should not crash the run. The generated branch name should be valid (e.g. the empty segment is dropped, or the run falls back to the default branch format).

### Root cause

- `extractDescription()` in `src/utils/branch-template.ts` strips to `[a-z0-9-]` and returns `""` when nothing survives.
- `applyBranchTemplate()` substitutes that empty string verbatim, leaving the surrounding literal `/` characters adjacent → `//`.
- `generateBranchName()`'s empty-result fallback only fires when the **whole** result is empty; `claude//123` is non-empty, so it is returned as-is.
- `validateBranchName()` (`src/github/operations/branch.ts:315`) then throws on the consecutive slashes, and the throw is not caught in `setupBranch` — the run aborts.

### Possible fix directions (non-prescriptive)

- Collapse consecutive slashes and trim leading/trailing slashes in the final generated name, so any empty segment (from `description`, `label`, `sha`, …) degrades gracefully rather than crashing — this fixes the whole class, independent of which `{{label}}` PR lands.
- Or: when `{{description}}` resolves empty, fall back to `entityType` (mirroring the empty-label fallback the #1491 PRs add), for consistency.

Happy to open a PR once there's agreement on the preferred approach — I have a fix and regression tests ready.

### Environment

- Reproduced against `main` (`src/utils/branch-template.ts`, `src/github/operations/branch.ts`).

cyphercodes:
## Summary

- Normalize generated branch template results to drop empty slash-delimited segments before validation
- Preserve the default-format fallback when slash normalization leaves no usable branch name
- Add regression coverage for titles whose `{{description}}` sanitizes to an empty string

Fixes #1527

## Test plan

- `bun test test/branch-template.test.ts`
- `bun test`
- `bun run typecheck`
- `bun run format:check`
- `git diff --check`

### DIFF
--- a/src/utils/branch-template.ts
+++ b/src/utils/branch-template.ts
@@ -42,6 +42,15 @@ function sanitizeLabel(label: string): string {
     .replace(/^-|-$/g, ""); // Remove leading/trailing hyphens
 }
 
+/**
+ * Drops empty branch path segments that can be produced by optional template
+ * variables (for example an emoji-only title causing {{description}} to be
+ * empty in "{{prefix}}{{description}}/{{entityNumber}}").
+ */
+function normalizeBranchTemplateResult(branchName: string): string {
+  return branchName.replace(/\/+/g, "/").replace(/^\/|\/$/g, "");
+}
+
 export interface BranchTemplateVariables {
   prefix: string;
   entityType: string;
@@ -97,9 +106,11 @@ export function generateBranchName(
   };
 
   if (template?.trim()) {
-    const branchName = applyBranchTemplate(template, variables);
+    const branchName = normalizeBranchTemplateResult(
+      applyBranchTemplate(template, variables),
+    );
 
-    // Some templates could produce empty results- validate
+    // Some templates could produce empty results or only empty path segments.
     if (branchName.trim().length > 0) return branchName;
 
     console.log(
--- a/test/branch-template.test.ts
+++ b/test/branch-template.test.ts
@@ -268,6 +268,39 @@ describe("branch template utilities", () => {
       expect(result).toBe("test/-101");
     });
 
+    it("should drop empty slash segments when description sanitizes to empty", () => {
+      const template = "{{prefix}}{{description}}/{{entityNumber}}";
+      const result = generateBranchName(
+        template,
+        "claude/",
+        "issue",
+        123,
+        undefined,
+        undefined,
+        "🎉🎉🎉",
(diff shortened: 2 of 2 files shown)


ashwin-ant (dismissed):

ashwin-ant:
@cyphercodes there's a conflict, mind fixing?

cyphercodes:
Resolved the merge conflict with current main and pushed 47d600c78f0d5757c1a812468304da88f13786b7.

Local verification:
- `bun test test/branch-template.test.ts`
- `bun run typecheck`
- `bun run format:check`
- `git diff --check`

Issue #1236, opened 2026-04-18
claude-code-action fails when actions/checkout uses persist-credentials: false


Guthman:
## Summary

`anthropics/claude-code-action@v1` fails during branch setup when the preceding `actions/checkout` step is configured with `persist-credentials: false`. The action's `setupBranch` runs `git fetch origin <ref>` before configuring its own git auth, so with no persisted credentials the fetch exits with `could not read Username for 'https://github.com'`.

This blocks users from following zizmor's [`artipacked`](https://docs.zizmor.sh/audits/#artipacked) recommendation, which is the mitigation for the well-known class of bugs where a persisted `GITHUB_TOKEN` leaks via artifact upload or a later step reading `.git/config`.

## Reproduction

Use the template generated by `/install-github-app`, but add `persist-credentials: false` to the checkout step:

```yaml
steps:
  - name: Checkout repository
    uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
    with:
      fetch-depth: 1
      persist-credentials: false

  - name: Run Claude Code
    uses: anthropics/claude-code-action@38ec876110f9fbf8b950c79f534430740c3ac009 # v1
    with:
      claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
```

Trigger the workflow (e.g. by `@claude` in an issue comment).

## Actual behaviour

```
Source branch SHA: <sha>
Creating local branch claude/issue-<n>-<ts> for issue #<n> from source branch: master...
Fetching and checking out source branch: master
fatal: could not read Username for 'https://github.com': No such device or address
Error in branch setup: ...
    at execGit (/home/runner/work/_actions/anthropics/claude-code-action/<sha>/src/github/operations/branch.ts:123:3)
    at setupBranch (/home/runner/work/_actions/anthropics/claude-code-action/<sha>/src/github/operations/branch.ts:299:5)
    at async prepareTagMode (/home/runner/work/_actions/anthropics/claude-code-action/<sha>/src/modes/tag/index.ts:66:28)
    at async run (/home/runner/work/_actions/anthropics/claude-code-action/<sha>/src/entrypoints/run.ts:216:17)

error: Command failed: git fetch origin master --depth=1
```

## Expected behaviour

[...]

PR #1534, opened 2026-07-21
fix: restore credential-free origin URL after token revocation

### LINKED ISSUE #1509: Revoked app token is left in the origin remote URL, breaking later git operations in the same job
## Summary

When no `github_token` input is provided, the action mints a GitHub App installation token, `configureGitAuth` embeds it in the origin remote URL ([src/github/operations/git-config.ts](https://github.com/anthropics/claude-code-action/blob/v1.0.169/src/github/operations/git-config.ts)), and the composite's final step revokes that token ("Revoke app token" in [action.yml](https://github.com/anthropics/claude-code-action/blob/v1.0.169/action.yml)). The origin URL is never restored, so after the action completes, `origin` still points to `https://x-access-token:<revoked-token>@github.com/<owner>/<repo>.git`.

Any later step in the same job that contacts `origin` then fails — including a second invocation of this action itself, whose tag-mode `setupBranch` runs `git fetch origin` before configuring its own auth (#1236).

## Reproduction

```yaml
steps:
  - uses: actions/checkout@v6
    with:
      persist-credentials: false

  # First invocation (agent mode) - completes fine, revokes its token on exit
    with:
      prompt: "Summarize the repository layout."

  # Any later git operation against origin fails
  - run: git fetch origin main
```

In our case the failing step was a second invocation of the action (tag mode). Its `setupBranch` fetch used the revoked token left in the origin URL:

```
This is an open PR, checking out PR branch...
remote: Invalid username or token. Password authentication is not supported for Git operations.
Error: Action failed with error: Command failed: git fetch origin --depth=20 <branch>
```

Note: with the default `persist-credentials: true` and actions/checkout >= v6 (or v5.0.1 / v4.3.1), the failure can be masked, because the credential persisted by checkout is stored in a separate config file that survives `configureGitAuth`'s cleanup and still authenticates the fetch despite the revoked token in the URL. I will file that cleanup defect separately.

## Expected behaviour

The cleanup that revokes the token should also restore a credential-free origin URL, e.g.:

```
git remote set-url origin "https://github.com/<owner>/<repo>.git"
```

leaving the workspace usable for later steps, the same way `actions/checkout` cleans up after itself.

## Version

Observed on v1.0.169; the relevant code is unchanged on current `main`.

gunjanjaswal:
Fixes #1509.

When no `github_token` input is given, the action mints an app installation token and `configureGitAuth` embeds it in the origin remote URL. The final composite step revokes that token but never restores the URL, so the job ends with `origin` pointing at a dead credential. Any later git operation against `origin` fails with "Invalid username or token", most visibly a second invocation of this action, whose tag-mode `setupBranch` fetches before configuring its own auth. That second-invocation symptom is what #1236 reports, so this should close that one out too.

This adds a cleanup step next to the revoke step, under the same guard, that strips credentials from the origin URL, drops the credential helper this action configured, and deletes the helper script. Both auth branches are handled, and the cleanup is idempotent so running the action twice in one job is safe. The clean URL now comes from a shared helper used by both branches so the two paths cannot drift.

I kept the guard identical to the revoke step rather than making it unconditional. A user-supplied `github_token` stays valid after the run, and stripping it could break workflows that push to `origin` afterwards.

The cleanup rewrites whatever URL is actually configured rather than rebuilding one from event context, so a remote pointing somewhere other than the event repo is left intact. That is the concern #966 raises, so there is some overlap. Happy to rebase on it if you would rather land #966 first. Worth flagging too that #1526 edits the same region of `git-config.ts` for a different defect, so whichever lands second will need a rebase.

Tests cover both auth branches, the idempotency case, and assert the resulting `.git/config` carries no `x-access-token` and no dangling `credential.helper`. I checked they fail without the source change.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -445,6 +445,20 @@ runs:
           --config="${GITHUB_ACTION_PATH}/bunfig.toml" \
           run ${GITHUB_ACTION_PATH}/src/entrypoints/post-buffered-inline-comments.ts
 
+    - name: Restore git authentication
+      # Same guard as the revoke step below: the credentials configured for the
+      # run are about to stop working, so they must not be left behind for later
+      # steps (or a second invocation of this action) to pick up (issue #1509).
+      if: always() && inputs.github_token == '' && steps.run.outputs.github_token != '' && steps.run.outputs.skipped_due_to_workflow_validation_mismatch != 'true'
+      shell: bash
+      run: |
+        BUN_BIN="${GITHUB_ACTION_PATH}/bin/bun"
+        [ -x "$BUN_BIN" ] || BUN_BIN="bun"
+        # No --tsconfig-override: see the "Run Claude Code Action" step above.
+        "$BUN_BIN" --no-env-file \
+          --config="${GITHUB_ACTION_PATH}/bunfig.toml" \
+          run ${GITHUB_ACTION_PATH}/src/entrypoints/cleanup-git-config.ts || true
+
     - name: Revoke app token
       if: always() && inputs.github_token == '' && steps.run.outputs.github_token != '' && steps.run.outputs.skipped_due_to_workflow_validation_mismatch != 'true'
       shell: bash
--- a/src/entrypoints/cleanup-git-config.ts
+++ b/src/entrypoints/cleanup-git-config.ts
@@ -1,0 +1,23 @@
+#!/usr/bin/env bun
+
+/**
+ * Undo the git authentication configured for the run.
+ *
+ * Runs as a post step next to the token revocation so the workspace is left
+ * without a reference to a token that is about to stop working (issue #1509).
+ */
+
+import { cleanupGitAuth } from "../github/operations/git-config";
+
+async function run() {
+  try {
+    await cleanupGitAuth();
+  } catch (error) {
+    // Don't fail the action if cleanup fails, just log it
+    console.error("Failed to clean up git authentication:", error);
+  }
+}
+
+if (import.meta.main) {
+  run();
+}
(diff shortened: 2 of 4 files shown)


gunjanjaswal:
Rebased onto `main` after the credential-config refactor landed. #1509 is still open and the bug is still present — the setup path embeds the token in the origin URL via `replaceCheckoutCredentials`, and nothing restores it after the revoke step — so I reworked the fix to fit the new structure rather than close it.

- `cleanupGitAuth()` in `git-config.ts` strips credentials from the origin URL, removes only the credential helper this action configured, and deletes the helper script. It never throws, since it runs in an `always()` step.
- Wired as a `Restore git authentication` post step right before `Revoke app token`, under the same guard.
- Setup and cleanup now share `getCredentialHelperPath()` so the helper path can't drift between them.
- Tests are in `test/git-config-cleanup.test.ts` (kept separate from the rewritten `git-config.test.ts`), covering both auth paths, the token-URL restore, leaving a workflow-set helper alone, idempotency, ssh remotes, and a missing origin.

`typecheck`, the new tests (9), and the existing `git-config` tests all pass. Mergeable now.

gunjanjaswal:
Gentle bump. This restores the credential-free origin URL after a token revocation so the remote stays usable, a small scoped change, still green and mergeable. Happy to rebase if needed, just flagging it's ready for review whenever someone can take a look.

gunjanjaswal:
Friendly bump. This has been open a couple of months without a first review. It's a focused fix that restores the credential-free origin URL after token revocation, so a later push doesn't end up carrying the revoked token. Happy to rebase or walk through the change if that helps get it on someone's radar.

Issue #1357, opened 2026-05-26
Agent mode: MCP-server install decision uses a different parser (`parseAllowedTools`) than the one that builds the actual tool list, and the parsers diverge on comments and multi-line flag values


KenSCompareCredit:
## Summary

In agent mode, the action decides **which bundled MCP servers to install** using one parser (`parseAllowedTools`, a regex scanner in `src/modes/agent/parse-tools.ts`), but builds the **actual `allowedTools` the SDK receives** using a different parser (`parseSdkOptions` → `parseClaudeArgsToExtraArgs`, a `shell-quote`-based parser in `base-action/src/parse-sdk-options.ts`).

These two parsers interpret the same `--allowedTools` / `--disallowedTools` content differently. As a result, the set of tools used to decide MCP-server provisioning can differ from the set actually granted to the model. When they disagree on an `mcp__github*` entry, the action either fails to start a server whose tools the model is nonetheless told it has, or starts a server based on a tool reference the model never actually receives.

I've confirmed **2 distinct divergences** (repros below):

1. **Comment lines** — `parseAllowedTools` does not strip `#` comment lines, so a `--allowedTools "…"` pattern appearing inside a comment is counted. `parseSdkOptions` correctly ignores it (it calls `stripShellComments`). This is a direct consequence of the #802 fix being applied to only one of the two parsers.
2. **Multi-line single-flag values** — when one `--allowedTools` is followed by multiple quoted values on subsequent lines, `parseAllowedTools` captures only the first value, while `parseSdkOptions` (whose accumulating-flag logic consumes all consecutive non-flag tokens) captures all of them.

## Context on the two parsers

- `src/modes/agent/index.ts` calls `parseAllowedTools(userClaudeArgs)` and passes the result to `prepareMcpConfig(...)`. `prepareMcpConfig` (`src/mcp/install-mcp-server.ts`) uses it to decide whether to spin up MCP servers like `github`, `github_inline_comment`, `github_ci`, `github_comment`, etc. (e.g. `hasGitHubMcpTools = allowedToolsList.some(t => t.startsWith("mcp__github__"))`).
- Separately, `base-action/src/parse-sdk-options.ts` (`parseClaudeArgsToExtraArgs` + the allowed/disallowed merge logic in `parseSdkOptions`) parses the same `claude_args` to build the `allowedTools` / `disallowedTools` actually handed to the Agent SDK.

[...]

PR #1554, opened 2026-07-27
fix: parse claude args with shell parser

webpro:
## Summary

`claude_args` is tokenized by dropping any line whose first non-whitespace character is `#`, then running `shell-quote`. That loses information a shell would have kept, and the two places that read `claude_args` disagree about the result.

## Three bugs

**1. A `#` line inside a multi-line quoted value is deleted.**

```
--append-system-prompt "Follow these rules:
# 1. never force push
# 2. never delete branches
"
```

reaches Claude as `Follow these rules:\n`. Both instructions are gone, with no error. A numbered list written with `#` is an ordinary way to write one, and `stripShellComments` says in its own doc comment that inline `#` inside a quoted value is left untouched, which is the intent this breaks.

**2. A line continuation drops the value entirely.**

```
--disallowedTools \
  "Bash(rm:*)" "Bash(sudo:*)"
```

`shell-quote` emits an empty token for the `\`, and the empty token lands where the accumulating flag expects its first value. The result is that `disallowedTools` comes back `undefined`, so neither deny rule applies.

**3. The two `claude_args` readers disagree.**

`src/modes/agent/parse-tools.ts` states that it MUST agree with `parseClaudeArgsToExtraArgs`, because otherwise a tool is granted without its MCP server being installed, or a server is installed for a tool that was never granted (#1357). It never applied the metacharacter escaping the SDK parser applies, so:

```
--allowedTools Bash(gh:*)

  SDK parser      -> ["Bash(gh:*)"]
  MCP installer   -> ["Bash", "gh:*"]
```

## A fourth case, in its own commit

An unterminated quote runs to the end of the input, so this:

```
--allowedTools "Read
--max-turns 5
```

granted the single tool `Read\n--max-turns 5` and silently dropped `--max-turns`. Nothing surfaced, because the previous tokenizer had no way to say the input was malformed. The parser reports a position, so `splitClaudeArgs` now raises it:

```
Could not parse claude_args: unterminated double quote at position 15 near `"Read`.
Values containing spaces or shell characters need quoting.
```

I checked this against every `claude_args` in the test suite and docs, including MCP config JSON, glob tools, nested quotes, line continuations, `#` inside values and escaped spaces. None of them parse with an error. When one does, the token list is already wrong, so failing is better than granting a tool nobody asked for.

This is deliberately a separate commit (`37f020f`). It changes a silent mis-parse into a hard failure, which is a behaviour change rather than a bug fix, so if you would rather keep this PR to the three bugs above, revert that one commit and the rest stands on its own.

## What this changes

`stripShellComments` becomes `splitClaudeArgs`, backed by a shell parser and exported so `parse-tools.ts` can share it instead of keeping a second copy. Comments, quoting and line continuations are resolved by the grammar rather than by scanning lines, so there is one canonical answer for what `claude_args` contains.

The private-use-codepoint escaping for `()|&;<>` is unchanged. Those are genuine control operators in shell grammar, so a parser splits on them too. Keeping the escaping is what makes `--allowedTools Bash(gh:*)` and `Read(path/**)` round-trip, and applying it in the shared tokenizer is what fixes bug 3.

`shell-quote` is now unused and is dropped from both packages. Its test file is updated in place to cover the same behaviours against the tokenizer that replaced it, plus the regressions above.

## About the dependency

This swaps `shell-quote` for [unbash](https://github.com/webpro-nl/unbash), so the dependency count is unchanged. I wrote unbash. I also wrote [knip](https://github.com/webpro-nl/knip), which depends on it, so it already runs against a large amount of real world shell: knip is at about 11.7M downloads a week and unbash at about 8.5M. It has no dependencies of its own, is synchronous, and needs no WASM or async initialisation.

## Verification

```
bun test            810 pass, 0 fail, 44 files   (804 on main, +6 regressions)
bun run typecheck   clean
bun run format      clean
```

All three bugs were reproduced against `main` first and confirmed fixed after.

## Related issues

#1357 is the divergence bug 3 describes.

### DIFF
--- a/base-action/bun.lock
+++ b/base-action/bun.lock
@@ -7,7 +7,7 @@
       "dependencies": {
         "@actions/core": "^1.10.1",
         "@anthropic-ai/claude-agent-sdk": "^0.3.220",
-        "unbash": "^4.0.3",
+        "unbash": "^4.0.5",
       },
       "devDependencies": {
         "@types/bun": "^1.2.12",
@@ -243,7 +243,7 @@
 
     "typescript": ["typescript@5.8.3", "", { "bin": { "tsc": "bin/tsc", "tsserver": "bin/tsserver" } }, "sha512-p1diW6TqL9L07nNxvRMM7hMMw4c5XOo/1ibL4aAIGmSAt9slTE1Xgw5KWuof2uTOvCg9BY7ZRi+GaF+7sfgPeQ=="],
 
-    "unbash": ["unbash@4.0.3", "", {}, "sha512-3cudTErfToSc4Ggv8XGXVNVli/xHKUtUZvaY5UVwhOcUPbQGz7PeaEnT/SAVgNziZtX67KEN9swMUYkLghxA1w=="],
+    "unbash": ["unbash@4.0.5", "", {}, "sha512-EE9xv9cr93DSppe086Rnbq0jwG7MBCEe22JZ4UQ8Bn9RyUwDSoUpBmzdb5+7PzocdqBEb/K73FGPaQUMmLxTQQ=="],
 
     "undici": ["undici@5.29.0", "", { "dependencies": { "@fastify/busboy": "^2.0.0" } }, "sha512-raqeBD6NQK4SkWhQzeYKd1KmIG6dllBOTt55Rmkt4HtI9mwdWtJljnrXjAFUBLTSN67HWrOIZ3EPF4kjUw80Bg=="],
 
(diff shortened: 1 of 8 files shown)


rishavnaskar:
Closed #1583 in favour of this, since the grammar approach here covers the same root cause more thoroughly.

One case from #980 you might want as an explicit regression test — a mid-word `#`, which `shell-quote` truncates:

```ts
expect(splitClaudeArgs("--branch-name fix/#123-description")).toEqual([
  "--branch-name",
  "fix/#123-description",
]);
```

On `main` the value comes back as `fix/`. Worth pinning because `#` is explicitly valid in branch names — `validateBranchName` documents `fix/#123-description` as a supported example.

webpro:
Thanks Rishav, added the test. Also bumped unbash itself (not needed but doesn't hurt either as it contains perf improvements and edge case fixes).

PR #108, opened 2025-06-02
Fix 401 error on GitHub Enterprise Server by passing GITHUB_API_URL to Octokit

uc4w6c:
## Summary
This PR fixes a 401 Unauthorized error that occurs in GitHub Enterprise Server by specifying the baseUrl when creating the Octokit instance. #107 

## Changes
- Adds `baseUrl: GITHUB_API_URL` to Octokit initialization

## Purpose
To ensure authenticated API requests work properly in GitHub Enterprise Server environments.

PR #1575, opened 2026-07-31
fix(github): honor GITHUB_GRAPHQL_URL for the GraphQL client

rover0811:
## Problem

`createOctokit()` builds the REST client from `GITHUB_API_URL` (correct) but builds the **GraphQL** client from `GITHUB_API_URL` too, ignoring the standard `GITHUB_GRAPHQL_URL` variable the Actions runner provides.

On standard GitHub Enterprise Server this happens to work, because `@octokit/graphql` rewrites a `.../api/v3` REST base to `.../api/graphql`. But any deployment whose GraphQL endpoint is **not derivable** from the REST base by that rule (a custom proxy, or GraphQL served from a separate host) has its GraphQL requests sent to the wrong URL.

## Fix

Honor `GITHUB_GRAPHQL_URL` independently, falling back to `GITHUB_API_URL` when it is unset so behavior is **unchanged for github.com and standard GHES**. A single trailing `/graphql` is stripped because `@octokit/graphql` appends its own (avoids `/graphql/graphql`); the trailing-slash trim keeps the existing `.../api/v3` → `.../api/graphql` rewrite working on the fallback path.

REST stays on `GITHUB_API_URL`. This is the GraphQL analogue of #108 / #243, which established honoring `GITHUB_API_URL` for REST on GHES.

## Tests

Added wire-level regression tests that run the real client factory in a fresh process and assert the **final request URL + Authorization header** — constructor-option assertions are insufficient because `@octokit/graphql` rewrites the path after construction. Coverage:

- both env vars unset → github.com for REST and GraphQL
- `GITHUB_API_URL` only (GHES) → GraphQL still resolves to `/api/graphql`
- `GITHUB_GRAPHQL_URL` only → GraphQL honors it, REST stays public
- both set → REST and GraphQL route independently
- `GITHUB_GRAPHQL_URL` overriding a derivable base
- trailing-slash normalization
- no `/graphql/graphql` doubling
- Authorization header preserved on both clients

`bun test` (full suite), `bun run typecheck`, and `bun run format:check` all pass.

### DIFF
--- a/src/github/api/client.ts
+++ b/src/github/api/client.ts
@@ -1,6 +1,6 @@
 import { Octokit } from "@octokit/rest";
 import { graphql } from "@octokit/graphql";
-import { GITHUB_API_URL } from "./config";
+import { GITHUB_API_URL, GITHUB_GRAPHQL_URL } from "./config";
 
 export type Octokits = {
   rest: Octokit;
@@ -14,7 +14,7 @@ export function createOctokit(token: string): Octokits {
       baseUrl: GITHUB_API_URL,
     }),
     graphql: graphql.defaults({
-      baseUrl: GITHUB_API_URL,
+      baseUrl: GITHUB_GRAPHQL_URL,
       headers: {
         authorization: `token ${token}`,
       },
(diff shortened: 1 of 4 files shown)


ashwin-ant (approved):

rover0811:
@ashwin-ant this was approved on the 14th but seems to have been left out when the rest of that batch went in. Still conflict free against main and nothing has been pushed since the review. Anything you need from me here? Happy to rebase or trim the test fixture.

PR #1581, opened 2026-08-02
fix(format): keep all text blocks in step summary tool results

### LINKED ISSUE #1572: Step summary drops later text blocks from structured tool results
**Describe the bug**

The step-summary formatter recognizes structured tool output shaped like
`[{ type: "text", text: "..." }]`, but it only keeps the first text block.

If a successful `tool_result.content` contains multiple text items, the GitHub
Actions step summary only shows the first one and silently drops the rest.

That means follow-up details such as extra findings, file paths, or secondary
instructions can disappear from the visible `Claude Code Report` even though
they are still present in the execution transcript.

**To Reproduce**

1. Feed `formatTurnsFromData()` a transcript with:
   - one assistant `tool_use`
   - one successful `tool_result`
   - `tool_result.content` set to:

```json
[
  { "type": "text", "text": "first line" },
  { "type": "text", "text": "second line" },
  { "type": "text", "text": "third line" }
]
```

2. Generate the step-summary markdown.
3. Observe that the output contains `first line` only.
4. Observe that `second line` and `third line` are missing.

Relevant logic today:

```ts
if (
  Array.isArray(parsedContent) &&
  parsedContent.length > 0 &&
  typeof parsedContent[0] === "object" &&
  parsedContent[0]?.type === "text"
) {
  contentStr = parsedContent[0]?.text || "";
}
```

**Expected behavior**

Expected behavior would be for all text blocks in a structured text array to
remain visible in the rendered summary, not only the first one.

**Screenshots**

Not included.

**Workflow yml file**

Not provider- or workflow-specific. This appears to be in the shared summary
formatting path used after execution output is written.

**API Provider**

[ ] AWS Bedrock
[ ] GCP Vertex

**Additional context**

- The summary path is `src/entrypoints/run.ts -> formatTurnsFromData()`.
- The formatter code is in `src/entrypoints/format-turns.ts`.
- There does not appear to be a regression test for multi-block successful
  `tool_result.content`.
- A straightforward fix would be to join all `{ type: "text" }` items rather
  than taking only `parsedContent[0]`.

rishavnaskar:
Fixes #1572.

`formatResultContent` only read `parsedContent[0].text`, so a `tool_result` whose `content` holds multiple text blocks rendered only the first one in the step summary — the rest were silently dropped from the Claude Code Report even though they're present in the transcript.

**Before**

```
[{"type":"text","text":"first line"},{"type":"text","text":"second line"}]
→ "**→** first line"
```

**After**

```
→ "**→** first line\nsecond line"
```

Non-text blocks (e.g. images) are skipped rather than stringified, and the single-block case is unchanged.

### Tests

Three cases added to `test/format-turns.test.ts`: multiple text blocks, mixed text/image blocks, and structured content passed as an array rather than a JSON string. All three fail on `main` and pass with this change.

```
bun run typecheck     clean
bun run format:check  clean
```

### DIFF
--- a/src/entrypoints/format-turns.ts
+++ b/src/entrypoints/format-turns.ts
@@ -163,8 +163,11 @@ export function formatResultContent(content: any): string {
       typeof parsedContent[0] === "object" &&
       parsedContent[0]?.type === "text"
     ) {
-      // Extract the text field from the first item
-      contentStr = parsedContent[0]?.text || "";
+      // Join every text block; a tool result may carry more than one.
+      contentStr = parsedContent
+        .filter((block: any) => block?.type === "text")
+        .map((block: any) => block?.text || "")
+        .join("\n");
     } else {
       contentStr = String(content).trim();
     }
--- a/test/format-turns.test.ts
+++ b/test/format-turns.test.ts
@@ -111,6 +111,39 @@ describe("formatResultContent", () => {
     const result = formatResultContent(JSON.stringify(structuredContent));
     expect(result).toBe("**→** Hello world\n\n");
   });
+
+  test("keeps every text block, not just the first", () => {
+    const structuredContent = [
+      { type: "text", text: "first line" },
+      { type: "text", text: "second line" },
+      { type: "text", text: "third line" },
+    ];
+    const result = formatResultContent(JSON.stringify(structuredContent));
+    expect(result).toContain("first line");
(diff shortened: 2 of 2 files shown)


rishavnaskar:
Duplicate of #1574, which was opened first and also handles arrays where the text block isn't the first element. Closing in favour of that one.

PR #1582, opened 2026-08-02
fix(branch): validate generated branch name under commit signing

### LINKED ISSUE #1573: `use_commit_signing` path skips generated branch-name validation and fails later in `github_file_ops`
**Describe the bug**

When `branch_name_template` produces an invalid git ref name, the non-signing
path rejects it early with `validateBranchName(newBranch)`, but the
`use_commit_signing: true` path does not perform the same validation.

Instead, the invalid branch name is handed to `github-file-ops-server`, and the
first commit attempt fails later with a GitHub API error such as:

```text
Failed to create branch: 422 - {"message":"Reference update failed","errors":[{"message":"Reference name is not valid"}]}
```

This makes the behavior inconsistent across commit modes and delays a
configuration error until the first remote file-ops call.

**To Reproduce**

1. Configure a workflow that creates a new Claude branch, with:

```yaml
use_commit_signing: true
branch_name_template: "{{prefix}}release:{{entityNumber}}"
```

2. Trigger a flow that creates a fresh Claude branch (for example an issue
   flow, or a closed/merged PR flow).
3. Let Claude make a change that reaches the first
   `mcp__github_file_ops__commit_files` call.
4. Observe that branch creation fails with a late GitHub API error because the
   generated branch name contains `:`.

For comparison:

- with `use_commit_signing: false`, the generated `newBranch` is validated
  locally before checkout/creation;
- with `use_commit_signing: true`, the same `newBranch` is not pre-validated
  before being passed to the file-ops server.

**Expected behavior**

Both commit modes should validate the generated branch name consistently
before handing it off to downstream git or GitHub API operations.

If the template yields an invalid branch name, the run should fail early with a
clear validation error rather than waiting for the first file-ops branch-create
attempt.

**Screenshots**

Not included.

**Workflow yml file**

Minimal relevant fragment:

```yaml
on:
  issues:
    types: [opened]

jobs:
  claude:
    runs-on: ubuntu-latest
    steps:
      - uses: anthropics/claude-code-action@main
        with:
          use_commit_signing: true
          branch_name_template: "{{prefix}}release:{{entityNumber}}"
          prompt: "Make a small change"
```

**API Provider**

[ ] AWS Bedrock
[ ] GCP Vertex

**Additional context**

- `branch_name_template` is a documented action input.
- A template like `{{prefix}}release:{{entityNumber}}` currently generates
  `claude/release:123`.
- In `src/github/operations/branch.ts`, the non-signing path validates
  `newBranch`, but the `use_commit_signing` path validates only `sourceBranch`
  before returning the unvalidated `claudeBranch`.
- A local reproduction of the real `src/mcp/github-file-ops-server.ts` against
  a mock GitHub API produced the following branch-create body:

```json
{"ref":"refs/heads/claude/release:123","sha":"base-sha-1"}
```

- A likely fix is to run `validateBranchName(newBranch)` before the
  `use_commit_signing` early return, so both commit modes enforce the same
  branch-name rules.

rishavnaskar:
Fixes #1573.

`setupBranch` validated the generated branch name on the non-signing path only:

```ts
validateBranchName(sourceBranch);
validateBranchName(newBranch);   // non-signing path only
```

On the `use_commit_signing: true` path the name went straight to the file ops server, so a `branch_name_template` that yields an invalid ref surfaced late as:

```
Failed to create branch: 422 - {"message":"Reference update failed",
"errors":[{"message":"Reference name is not valid"}]}
```

Moved the check to just after the name is resolved (including the existing-branch fallback), so both commit modes fail early with the same local message. The duplicate call on the non-signing path is removed.

### Tests

`test/setup-branch-validation.test.ts` drives `setupBranch` with `branch_name_template: "{{prefix}}release:{{entityNumber}}"` under both commit modes and asserts the run fails on the name itself rather than on a later git or API call.

On `main` the signing case fails the new assertion — it gets past validation and dies later with `Command failed: git fetch origin main --depth=1` — while the non-signing case passes, which is the control that shows the test is measuring the asymmetry and not something incidental.

```
bun test              806 pass, 0 fail
bun run typecheck     clean
bun run format:check  clean
```

### DIFF
--- a/src/github/operations/branch.ts
+++ b/src/github/operations/branch.ts
@@ -288,6 +288,11 @@ export async function setupBranch(
       // Branch doesn't exist (non-zero exit code), continue with generated name
     }
 
+    // Validate before either path uses the name. The signing path hands it to
+    // the file ops server rather than to git, so without this an invalid
+    // template only surfaces as a 422 on the first commit.
+    validateBranchName(newBranch);
+
     // For commit signing, defer branch creation to the file ops server
     if (context.inputs.useCommitSigning) {
       console.log(
@@ -315,7 +320,6 @@ export async function setupBranch(
     // Fetch and checkout the source branch first to ensure we branch from the correct base
     console.log(`Fetching and checking out source branch: ${sourceBranch}`);
     validateBranchName(sourceBranch);
-    validateBranchName(newBranch);
     execGit(["fetch", "origin", sourceBranch, "--depth=1"]);
     execGit(["checkout", sourceBranch, "--"]);
 
--- a/test/setup-branch-validation.test.ts
+++ b/test/setup-branch-validation.test.ts
@@ -1,0 +1,80 @@
+import { describe, expect, test, beforeEach, afterEach } from "bun:test";
+import { mkdtempSync, rmSync } from "fs";
+import { join } from "path";
+import { setupBranch } from "../src/github/operations/branch";
+import { createMockContext } from "./mockContext";
+
+const octokits = {
+  rest: {
+    repos: { get: async () => ({ data: { default_branch: "main" } }) },
+    git: { getRef: async () => ({ data: { object: { sha: "abc1234" } } }) },
+  },
+} as any;
+
+const githubData = {
+  contextData: { title: "Add feature", labels: { nodes: [] } },
+} as any;
+
+// ':' is rejected by validateBranchName. The signing path used to skip that
+// check and only fail on the file ops server's first commit (a 422).
+const INVALID_TEMPLATE = "{{prefix}}release:{{entityNumber}}";
+
+const loggedErrors: string[] = [];
+
+describe("setupBranch generated branch name validation", () => {
+  let originalCwd: string;
+  let tempDir: string;
+  let exitCode: number | undefined;
+  let originalExit: typeof process.exit;
+  let originalError: typeof console.error;
+
+  beforeEach(() => {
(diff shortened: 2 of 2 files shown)


rishavnaskar:
#1578 was opened first for this issue and takes the same approach with a one-line check inside the signing branch.

Two differences here in case they're useful:

- Validation happens once after the branch name is resolved (including the existing-branch fallback), so the two commit modes can't drift apart again.
- The test drives `setupBranch` rather than calling `validateBranchName` on `generateBranchName` output directly. The latter passes on unpatched `main`, so it won't catch a regression if the check is removed.

Happy to close this if #1578 is preferred.

ashwin-ant (approved):

rishavnaskar:
On the failing checks — they're the credential-gated integration suites, not this change.

They abort before running any code:

```
Secret source: None
##[error]Environment variable validation failed:
  - Either ANTHROPIC_API_KEY, CLAUDE_CODE_OAUTH_TOKEN, or workload identity federation
    (ANTHROPIC_FEDERATION_RULE_ID and ANTHROPIC_ORGANIZATION_ID) is required
```

Fork PRs don't get repository secrets, so these can't pass from a fork. Same pattern on unrelated fork PRs (e.g. #1576): six passing, thirty failing.

All six secret-free checks are green here — `test`, `typecheck`, `prettier` (and their `ci /` duplicates).

PR #1583, opened 2026-08-02
fix(parse-sdk-options): treat mid-word # in claude_args as literal

### LINKED ISSUE #980: shell-quote treats # in claude_args values as comments, causing null values and AJV crash

rishavnaskar:
Fixes #980.

`shell-quote` starts a comment at an unquoted `#` **anywhere** in a token, but a POSIX shell only does so when `#` begins a word — `echo fix/#123` prints `fix/#123` in bash. The parser dropped the resulting `{comment}` object, so a `#` inside a value silently truncated it:

| `claude_args` | before | after |
|---|---|---|
| `--branch-name fix/#123-description` | `"fix/"` | `"fix/#123-description"` |
| `--append-system-prompt "# Use best practices"` | `"# Use best practices"` | unchanged |
| whole-line `# comment` | dropped | unchanged |

The silent truncation matters here because `#` is explicitly valid in branch names — `validateBranchName` documents `fix/#123-description` as a supported example.

`#` is now escaped to a private-use codepoint when it follows a non-space character, reusing the mechanism already in place for `()|&;<>`. A `#` that begins a word still starts a comment, so `stripShellComments` and trailing-comment behaviour are untouched.

This also corrects the assumption in the comment above the token mapping, which stated a glob op was the only non-string `shell-quote` could still emit.

### Tests

Three cases added to `base-action/test/parse-sdk-options.test.ts`. On `main` the mid-word case fails with `Expected: "fix/#123-description"` / `Received: "fix/"`; the quoted-value and whole-line-comment cases pass both before and after, which is the control showing existing behaviour is preserved.

```
bun run typecheck     clean
bun run format:check  clean
```

### DIFF
--- a/base-action/src/parse-sdk-options.ts
+++ b/base-action/src/parse-sdk-options.ts
@@ -44,14 +44,27 @@ const SHELL_META_PAIRS: [string, string][] = [
 const SHELL_META_ESCAPE = new Map(SHELL_META_PAIRS);
 const SHELL_META_UNESCAPE = new Map(SHELL_META_PAIRS.map(([k, v]) => [v, k]));
 const SHELL_META_ESCAPE_RE = /[()|&;<>]/g;
+
+// shell-quote starts a comment at an unquoted `#` anywhere in a token, but a
+// POSIX shell only does so at the start of a word: `fix/#123` is literal in
+// bash, while shell-quote truncates the token to `fix/`. Escape `#` only when
+// it follows a non-space character, so mid-word hashes survive and a `#` that
+// begins a word still starts a comment.
+const HASH_PLACEHOLDER = "\uE007";
+const MIDWORD_HASH_ESCAPE_RE = /(?<=\S)#/g;
+const HASH_PLACEHOLDER_RE = /\uE007/g;
 const SHELL_META_UNESCAPE_RE = /[-]/g;
 
 function escapeShellMeta(s: string): string {
-  return s.replace(SHELL_META_ESCAPE_RE, (c) => SHELL_META_ESCAPE.get(c)!);
+  return s
+    .replace(SHELL_META_ESCAPE_RE, (c) => SHELL_META_ESCAPE.get(c)!)
+    .replace(MIDWORD_HASH_ESCAPE_RE, HASH_PLACEHOLDER);
 }
 
 function unescapeShellMeta(s: string): string {
-  return s.replace(SHELL_META_UNESCAPE_RE, (c) => SHELL_META_UNESCAPE.get(c)!);
+  return s
+    .replace(SHELL_META_UNESCAPE_RE, (c) => SHELL_META_UNESCAPE.get(c)!)
+    .replace(HASH_PLACEHOLDER_RE, "#");
 }
 
 type McpConfig = {
--- a/base-action/test/parse-sdk-options.test.ts
+++ b/base-action/test/parse-sdk-options.test.ts
@@ -596,4 +596,43 @@ describe("parseSdkOptions", () => {
       }
     });
   });
+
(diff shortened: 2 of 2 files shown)


rishavnaskar:
Superseded by #1554, which replaces the shell-quote tokenizer with a real shell grammar and covers this case along with `#` inside multi-line quoted values and line continuations. Closing in favour of it.

PR #1025, opened 2026-03-06
fix: allow Unicode letters in branch name validation

Jah-yee:
## Summary

This PR adds multiple fixes for branch validation and permission checking:

### 1. Unicode Branch Names (#1020)
- Allows branch names with Unicode characters (e.g., Japanese like `機能/fix`)
- Uses Unicode property escapes (`\p{L}` for letters, `\p{N}` for numbers)

### 2. Azure DevOps # Character (#1024)
- Allows `#` in branch names for Azure DevOps ticket linking
- Azure DevOps uses `#` for ticket linking (e.g., `feature/AB#1992-sentry`)

### 3. Shell Comment Stripping (#1019)
- Strips lines starting with `#` before parsing `claude_args`
- Allows users to add documentation comments without breaking flags

### 4. Copilot Permission Check (#1018)
- Adds `Copilot` to the bypass list for permission checking
- Copilot-initiated workflows return 404 for user permission lookup

## Changes

- Modified `validPattern` regex for Unicode and # support
- Added comment stripping in `parseClaudeArgsToExtraArgs()`
- Added `Copilot` check alongside `[bot]` accounts

## Related

Fixes #1020, Fixes #1024, Fixes #1019, Fixes #1018

PR #1584, opened 2026-08-02
feat(context): include diffHunk in PR review comment context

### LINKED ISSUE #855: Include diffHunk in PR review comment context
# Include `diffHunk` in PR review comment context

## Problem

When Claude responds to a PR review comment attached to a specific line, it doesn't receive the actual code being discussed. This causes Claude to misinterpret questions that reference the code—without seeing it, Claude guesses what the question refers to based on other PR context, often incorrectly.

## Example

This happened on [worktrunk PR #809](https://github.com/max-sixty/worktrunk/pull/809). A reviewer commented on line 25 of `src/llm.rs` asking about a function that formats reproduction commands:

> @claude wouldn't it be easier to just do a single format? why have both?

The PR also included config format migration work. Without seeing the code at line 25, Claude interpreted "both formats" as referring to config file formats and gave a detailed answer about backward compatibility—completely missing that the question was about the `format_reproduction_command` function.

The reviewer had to follow up: "do you see the line that I'm commenting on? my comment is very specific to this area of code"

<details>
<summary>Full details: what Claude received vs what it needed</summary>

**Job run:** https://github.com/max-sixty/worktrunk/actions/runs/21281928575
**Comment:** https://github.com/max-sixty/worktrunk/pull/809#discussion_r2720505487

### What Claude received

From the [job logs](https://github.com/max-sixty/worktrunk/actions/runs/21281928575), the review comment context was:

```
<review_comments>
[Review by max-sixty at 2026-01-23T09:51:55Z]: COMMENTED
  [Comment on src/llm.rs:25]: @claude wouldn't it be easier to just do a single format? why have both?
</review_comments>
```

Claude saw the file path and line number, but not the code.

### What Claude needed

The `diffHunk` from GitHub's API would have shown:

```diff
+/// Format a reproduction command, only wrapping with `sh -c` if needed.
+///
+/// Simple commands like `llm -m haiku` are shown as-is.
+/// Complex commands with shell syntax are wrapped: `sh -c 'complex && command'`
+fn format_reproduction_command(base_cmd: &str, llm_command: &str) -> String {
```

With this context, "why have both formats?" clearly refers to simple vs `sh -c` wrapped command formats—not config file formats.

### What Claude answered

Claude gave a detailed response about config file backward compatibility, including:

> Line 25 isn't actually where both formats are handled - that's just a helper function for reproduction commands.

Claude correctly identified that line 25 was about reproduction commands, but still answered about config formats because that was the dominant theme in the PR.

</details>

## Current behavior

The GraphQL query for `PullRequestReviewComment` in `src/github/api/queries/github.ts` fetches:

```
path, line, body, author, createdAt, updatedAt, ...
```

But not `diffHunk`.

The formatter in `src/github/data/formatter.ts` outputs:

```
[Comment on src/file.rs:25]: question text here
```

## Requested behavior

1. Add `diffHunk` to the GraphQL query for `PullRequestReviewComment`
2. Include it in the formatted output, something like:

~~~
[Comment on src/file.rs:25]:
```diff
+fn format_reproduction_command(base_cmd: &str, llm_command: &str) -> String {
```
question text here
~~~

## Technical notes

- `diffHunk` is available in GitHub's GraphQL API for `PullRequestReviewComment`
- The REST API (`GET /repos/{owner}/{repo}/pulls/{pull_number}/comments`) includes `diff_hunk` by default
- The field contains the diff context showing the code where the comment was placed

rishavnaskar:
Fixes #855.

Review comments were passed to Claude with only `path` and `line`, so the code the comment was actually written against never reached the prompt. `diffHunk` is available on `PullRequestReviewComment` in the GraphQL API but wasn't being selected.

**Before**
```
  [Comment on src/index.ts:42]: This can overflow
```

**After**
```
  [Comment on src/index.ts:42]: This can overflow
  Diff context:
  ```diff
  @@ -40,3 +40,3 @@
  -const a = 1;
  +const a = 2;
  ```
```

Three changes: select `diffHunk` in the reviews query, add it to `GitHubReviewComment` as optional, and render it under the comment.

Two notes on the implementation:

- The hunk is PR-authored content reaching the prompt, so it goes through `sanitizeContent` the same way the comment body does.
- It's optional and the render is guarded, so comments without a hunk produce byte-identical output to before. All existing `formatReviewComments` tests pass unchanged.

Verified `diffHunk` against the live schema rather than assuming:

```
gh api graphql -f query='query { __type(name: "PullRequestReviewComment") { fields { name } } }'
→ diffHunk, line, path
```

### Tests

Two cases added to `test/data-formatter.test.ts`: a comment with a hunk, and one without. On `main` the first fails (`Expected to contain: "Diff context:"`); the second passes both before and after, which is the control showing existing output is unaffected.

```
bun test              806 pass, 0 fail
bun run typecheck     clean
bun run format:check  clean
```

### DIFF
--- a/src/github/api/queries/github.ts
+++ b/src/github/api/queries/github.ts
@@ -84,6 +84,7 @@ export const PR_QUERY = `
                 body
                 path
                 line
+                diffHunk
                 author {
                   login
                 }
--- a/src/github/data/formatter.ts
+++ b/src/github/data/formatter.ts
@@ -118,7 +118,16 @@ export function formatReviewComments(
 
           body = sanitizeContent(body);
 
-          return `  [Comment on ${comment.path}:${comment.line || "?"}]: ${body}`;
+          let formatted = `  [Comment on ${comment.path}:${comment.line || "?"}]: ${body}`;
+
+          // The diff hunk is the code the comment was left on. Without it the
+          // comment arrives without the context it was written against.
+          if (comment.diffHunk) {
+            const diffHunk = sanitizeContent(comment.diffHunk);
+            formatted += `\n  Diff context:\n\`\`\`diff\n${diffHunk}\n\`\`\``;
+          }
+
+          return formatted;
         })
         .join("\n");
       if (comments) {
--- a/src/github/types.ts
+++ b/src/github/types.ts
@@ -22,6 +22,7 @@ export type GitHubComment = {
 export type GitHubReviewComment = GitHubComment & {
   path: string;
   line: number | null;
+  diffHunk?: string | null;
 };
 
 export type GitHubCommit = {
--- a/test/data-formatter.test.ts
+++ b/test/data-formatter.test.ts
@@ -540,6 +540,43 @@ describe("formatReviewComments", () => {
     expect(result).not.toContain("Diff context:");
   });
 
+  // GitHub returns line: null and diffHunk: "" for outdated comments whose
+  // line no longer exists in the diff (observed on anthropics/claude-code-action#1025).
+  test("omits the diff context for an outdated comment with an empty diff hunk", () => {
+    const reviewData = {
+      nodes: [
+        {
+          id: "review1",
+          databaseId: "300001",
+          author: { login: "reviewer1" },
+          body: "",
+          state: "COMMENTED",
+          submittedAt: "2023-01-01T00:00:00Z",
+          comments: {
+            nodes: [
+              {
+                id: "comment1",
+                databaseId: "200001",
+                body: "Outdated comment",
+                author: { login: "reviewer1" },
+                createdAt: "2023-01-01T00:00:00Z",
+                path: "src/index.ts",
+                line: null,
+                diffHunk: "",
+              },
+            ],
+          },
+        },
+      ],
+    };
+
+    const result = formatReviewComments(reviewData);
+
+    expect(result).toContain("[Comment on src/index.ts:?]: Outdated comment");
+    expect(result).not.toContain("Diff context:");
+    expect(result).not.toContain("```diff");
+  });
+
   test("formats review with only body (no comments) correctly", () => {
     const reviewData = {
       nodes: [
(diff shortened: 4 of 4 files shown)


rishavnaskar:
Verified end to end rather than only in unit tests — ran the modified `PR_QUERY` against real public PRs and piped the live response through `formatReviewComments`:

| PR | review comments | populated `diffHunk` on `main` | on this branch |
|---|---|---|---|
| `cli/cli#13988` | 1 | 0 | 1 |
| `denoland/deno#36111` | 3 | 0 | 3 |
| `anthropics/claude-code-action#1025` | 1 | 0 | 0 (see below) |

Real output for `cli/cli#13988`:

```
  [Comment on api/client_test.go:193]: Interesting that we didn't have a test for this! 🤔
  Diff context:
```diff
@@ -138,6 +139,100 @@ func TestRESTError(t *testing.T) {
 	}
 }

+func TestRESTWithNextError(t *testing.T) {
```

That third row turned up an edge case worth pinning: #1025's comment is **outdated**, and GitHub returns `line: null` with `diffHunk: ""` — an empty string, not null. The guard already skips it because `""` is falsy, but a later change to `diffHunk != null` would emit an empty ```` ```diff ```` block. Added a regression test for it in 8021217.

```
bun run typecheck     clean
bun run format:check  clean
```

rishavnaskar:
On the failing checks — they're the credential-gated integration suites, not this change.

They abort before running any code:

```
Secret source: None
##[error]Environment variable validation failed:
  - Either ANTHROPIC_API_KEY, CLAUDE_CODE_OAUTH_TOKEN, or workload identity federation
    (ANTHROPIC_FEDERATION_RULE_ID and ANTHROPIC_ORGANIZATION_ID) is required
```

Fork PRs don't get repository secrets, so these can't pass from a fork. Same pattern on unrelated fork PRs (e.g. #1576): six passing, thirty failing.

All six secret-free checks are green here — `test`, `typecheck`, `prettier` (and their `ci /` duplicates), so the full unit suite passes in CI as well as locally.

Happy to rebase or adjust if anything in the diff needs it.

ashwin-ant (approved):

PR #1608, opened 2026-08-07
Document 1M gateway models and surface resolved limits

### LINKED ISSUE #1600: Model ID `claude-opus-5` resolves to a 200k context window (1M only with the `[1m]` suffix), and the action has no way to set it
**Describe the bug**

A workflow that sets the model to `claude-opus-5` — the model ID exactly as documented — runs a session that plans against a **200k** context window instead of the model's real **1M**. Auto-compaction therefore triggers at a fraction of the true limit, and long agent runs (large-diff PR reviews) compact repeatedly mid-task and fail.

Appending a `[1m]` suffix (`claude-opus-5[1m]`) resolves the window to 1,000,000 and fixes the behavior, but that suffix is not mentioned anywhere in this repo's docs, and the action exposes no input for the context window. Nothing in the action's output indicates which window was resolved, so the failure is silent.

Per the [models overview](https://platform.claude.com/docs/en/about-claude/models/overview), `claude-opus-5` and `claude-sonnet-5` are **1M context / 128k max output**, with no beta header and no opt-in — so `model: claude-opus-5` is the correct-looking configuration and the one users will write.

Secondary symptom: `maxOutputTokens` reports `64000` for both models with **and** without the suffix, while the documented cap is 128k.

**To Reproduce**

1. Configure the action with `ANTHROPIC_MODEL: claude-opus-5` (and `ANTHROPIC_SMALL_FAST_MODEL: claude-sonnet-5`), plus `show_full_output: true` so the result message is visible.
2. Run it on a task large enough to produce a long session (ours is a PR review over a ~3.5k-line diff).
3. In the job log, find the `modelUsage` object in the `result` message.
4. See `"contextWindow": 200000` for both models — and premature auto-compaction during the run.
5. Change only the model strings to `claude-opus-5[1m]` / `claude-sonnet-5[1m]` and re-run the same commit.
6. See `"contextWindow": 1000000`, no premature compaction, and the run completes.

Three consecutive attempts on the same PR. Attempts 2 and 3 hold the action version constant and vary only the suffix, so the version bump is ruled out as the cause:

| attempt | action version | model string | resolved `contextWindow` | outcome |
|---|---|---|---|---|
| 1 | v1.0.183 (CC 2.1.220) | `claude-opus-5` | 200000 | failed |
| 2 | **v1.0.185** (CC 2.1.222) | `claude-opus-5` | **200000** | failed |
| 3 | **v1.0.185** (CC 2.1.222) | `claude-opus-5[1m]` | **1000000** | succeeded |

**Expected behavior**

`claude-opus-5` and `claude-sonnet-5` should resolve to `contextWindow: 1000000` and `maxOutputTokens: 128000` — their documented limits — without needing a suffix. Auto-compaction should be sized against the real window.

If the suffix genuinely is the supported way to select the 1M variant, it should be documented in this repo, and the resolved limits should be visible in the run output so a wrong value is noticeable before a session degrades.

**Screenshots**

Attaching two screenshots of the `modelUsage` block, without and with the suffix. The same data as text, copied verbatim from the `result` message in the job log with `show_full_output: true` (timestamps stripped; `costUSD` and `provider` fields removed, nothing else changed):

Without the suffix — `contextWindow: 200000`:

```json
    "inference_geo": "global",
    "iterations": [],
    "speed": "standard"
  },
  "modelUsage": {
    "claude-sonnet-5": {
      "inputTokens": 13574,
      "outputTokens": 20,
      "cacheReadInputTokens": 0,
      "cacheCreationInputTokens": 0,
      "webSearchRequests": 0,
      "contextWindow": 200000,
      "maxOutputTokens": 64000,
      "canonicalModel": "claude-sonnet-5"
    },
    "claude-opus-5": {
      "inputTokens": 96209,
      "outputTokens": 55324,
      "cacheReadInputTokens": 1135701,
      "cacheCreationInputTokens": 149043,
      "webSearchRequests": 0,
      "contextWindow": 200000,
      "maxOutputTokens": 64000,
      "canonicalModel": "claude-opus-5"
    }
  }
```

With `[1m]` appended — same action version, same commit, `contextWindow: 1000000`:

```json
    "inference_geo": "global",
    "iterations": [],
    "speed": "standard"
  },
  "modelUsage": {
    "claude-sonnet-5[1m]": {
      "inputTokens": 13574,
      "outputTokens": 20,
      "cacheReadInputTokens": 0,
      "cacheCreationInputTokens": 0,
      "webSearchRequests": 0,
      "contextWindow": 1000000,
      "maxOutputTokens": 64000,
      "canonicalModel": "claude-sonnet-5"
    },
    "claude-opus-5[1m]": {
      "inputTokens": 26349,
      "outputTokens": 25382,
      "cacheReadInputTokens": 4252063,
      "cacheCreationInputTokens": 194796,
      "webSearchRequests": 0,
      "contextWindow": 1000000,
      "maxOutputTokens": 64000,
      "canonicalModel": "claude-opus-5"
    }
  }
```

`canonicalModel` resolves to the same model in both runs — the only field that moves is `contextWindow`.

**Workflow yml file**

Our real workflow calls an internal reusable workflow, so here is a minimal equivalent that shows the only difference between a failing and a working run:

```yaml
name: AI Code Review
on:
  pull_request:
    types: [opened, synchronize, reopened]

jobs:
  review:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: anthropics/claude-code-action@v1.0.185
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}
          show_full_output: true
          claude_args: |
            --max-turns 50
          prompt: <a long PR-review prompt>
        env:
          # 200k — silently wrong
          ANTHROPIC_MODEL: claude-opus-5
          ANTHROPIC_SMALL_FAST_MODEL: claude-sonnet-5

          # 1M — works, but undocumented in this repo
          # ANTHROPIC_MODEL: claude-opus-5[1m]
          # ANTHROPIC_SMALL_FAST_MODEL: claude-sonnet-5[1m]

          CLAUDE_AUTOCOMPACT_PCT_OVERRIDE: 60
```

**API Provider**

- [x] Anthropic First-Party API (default)
- [ ] AWS Bedrock
- [ ] GCP Vertex

First-party, reached through a custom `ANTHROPIC_BASE_URL` — an Anthropic-Messages-compatible gateway that forwards the request body to `api.anthropic.com` verbatim. Worth noting for this report: the gateway neither adds nor strips the `[1m]` suffix. Its own request logs record only the canonical name (`claude-opus-5`) across 425 forwards spanning both the failing and the working runs, so the suffix is resolved client-side before the request is sent and nothing in between is responsible for the difference.

**Additional context**

*Environment*

- `anthropics/claude-code-action@v1.0.185` (Claude Code 2.1.222, Agent SDK 0.3.222); also reproduced on v1.0.183 (Claude Code 2.1.220)
- GitHub Actions, `ubuntu-latest`, non-interactive (`--print`, stream-json)
- `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE=60`

*Impact*

With `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE=60`, the compaction threshold lands at **~120k instead of ~600k**. Our review agent needs ~195k of context on a large diff (largest single request measured: **194,798** tokens), so under the 200k assumption it compacted mid-review, lost context it had already gathered, and the job failed twice before we found the suffix. Both failed runs billed in full.

The failure mode is silent, and CI makes it worse than interactive use: there's no status line or `/context` to notice `…/200k`, so it presents as "the agent keeps forgetting" rather than as a misconfiguration. And because this action is how teams standardize agents across repos, one wrong default propagates — a shared wrapper action setting `model: claude-opus-5` puts every consuming repo on a 200k assumption at once.

*Where the resolution happens*

The action forwards the model string verbatim and holds no context-window logic of its own:

- `src/entrypoints/run.ts:293` — `model: process.env.ANTHROPIC_MODEL`
- `base-action/src/index.ts:56` — `model: process.env.ANTHROPIC_MODEL`
- grep for `contextWindow`, `canonicalModel`, `1m`, `200000`, `autocompact` across `src/` and `base-action/src/` → no hits

So the resolution itself looks like it lives in Claude Code's model registry, and I'm happy to file a companion issue there and cross-link. What seems actionable here is the workflow-author surface: one lever, no documentation of it, and no visibility into what got resolved.

*Prior art in this repo*

- #737 (open, no replies) — "how do i provide sonnet-4-5 with the ability to use up to 1m tokens?" Same gap, unanswered.
- #981 (closed, unmerged) — proposed a `context_window: "1m"` input setting the `context-1m-2025-08-07` beta header. That header is the mechanism for the pre-5 models where 1M was a header-gated beta; on the 5-series 1M is native, so that shape wouldn't fix this case even if it had landed.

*What would help*

1. Document the `[1m]` suffix in the action's model configuration docs, including which models it applies to.
2. Log the resolved model limits (`contextWindow`, `maxOutputTokens`) at startup, next to the model name the action already prints.
3. Decide the supported path — an explicit input, or confirmation that the suffix is supported so workflow authors can rely on it.
4. If the bare 5-series IDs get their real limits upstream, items 1 and 3 become moot and only the logging remains useful.

ghost:
Fixes #1600.

Document the `[1m]` selector for Claude Opus 5 and Claude Sonnet 5 when an `ANTHROPIC_BASE_URL` gateway prevents Claude Code from confirming native 1M context support.

Also retain the SDK-resolved `contextWindow` and `maxOutputTokens` values in the default sanitized result log. Only capacity metadata is exposed; token counts, costs, and message content remain hidden. The action does not rewrite model IDs or hard-code model limits, leaving Claude Code as the source of truth.

Validation:
- Added regression coverage for sanitized model usage output.
- `git diff --check`
- Focused TypeScript compile and runtime validation with SDK-compatible stubs

Full Bun tests were not run because the supplied bundle omits the tracked Bun lockfiles and installed dependencies, and Bun is unavailable in the execution environment.

### DIFF
--- a/base-action/src/run-claude-sdk.ts
+++ b/base-action/src/run-claude-sdk.ts
@@ -82,6 +82,35 @@ async function createPromptConfig(
   return createMultiBlockMessage();
 }
 
+type ModelUsageSummary = Record<
+  string,
+  {
+    contextWindow: number;
+    maxOutputTokens: number;
+  }
+>;
+
+/**
+ * Keep resolved model limits visible without exposing token usage or cost details.
+ */
+function sanitizeModelUsage(
+  modelUsage: SDKResultMessage["modelUsage"] | undefined,
+): ModelUsageSummary | undefined {
+  if (!modelUsage) {
+    return undefined;
+  }
+
+  return Object.fromEntries(
+    Object.entries(modelUsage).map(([model, usage]) => [
+      model,
+      {
+        contextWindow: usage.contextWindow,
+        maxOutputTokens: usage.maxOutputTokens,
+      },
+    ]),
+  );
+}
+
 /**
  * Sanitizes SDK output to match CLI sanitization behavior
  */
@@ -119,6 +148,7 @@ function sanitizeSdkOutput(
         num_turns: resultMsg.num_turns,
         total_cost_usd: resultMsg.total_cost_usd,
         permission_denials_count: resultMsg.permission_denials?.length ?? 0,
+        modelUsage: sanitizeModelUsage(resultMsg.modelUsage),
       },
       null,
       2,
--- a/base-action/test/run-claude-sdk.test.ts
+++ b/base-action/test/run-claude-sdk.test.ts
@@ -64,6 +64,95 @@ describe("runClaudeWithSdk", () => {
     }
   });
 
+  test("logs resolved model limits without exposing token usage", async () => {
+    const consoleLogSpy = spyOn(console, "log").mockImplementation(() => {});
+
+    tempDir = await mkdtemp(join(tmpdir(), "claude-sdk-"));
+    process.env.RUNNER_TEMP = tempDir;
+
+    const promptPath = join(tempDir, "prompt.txt");
+    await writeFile(promptPath, "test prompt");
+
+    const initMessage = {
+      type: "system",
+      subtype: "init",
+      session_id: "session-123",
+      model: "claude-opus-5",
+    };
+
+    const resultMessage = {
+      type: "result",
+      subtype: "success",
+      is_error: false,
+      duration_ms: 434,
+      num_turns: 1,
+      total_cost_usd: 1.23,
+      permission_denials: [],
+      modelUsage: {
+        "claude-opus-5": {
+          inputTokens: 96209,
+          outputTokens: 55324,
+          cacheReadInputTokens: 1135701,
+          cacheCreationInputTokens: 149043,
+          webSearchRequests: 0,
+          costUSD: 1.23,
+          contextWindow: 200000,
+          maxOutputTokens: 64000,
+        },
+      },
+    };
+
+    mock.module("@anthropic-ai/claude-agent-sdk", () => ({
+      query: async function* () {
+        yield initMessage;
+        yield resultMessage;
+      },
+    }));
+
+    try {
+      const { runClaudeWithSdk } = await import("../src/run-claude-sdk");
+
+      await expect(
+        runClaudeWithSdk(promptPath, {
+          sdkOptions: {},
+          showFullOutput: false,
+          hasJsonSchema: false,
+        }),
+      ).resolves.toMatchObject({ conclusion: "success" });
+
+      const sanitizedResult = consoleLogSpy.mock.calls
+        .map(([message]) => message)
+        .find(
+          (message) =>
+            typeof message === "string" && message.includes('"type": "result"'),
+        );
+
+      expect(sanitizedResult).toBeDefined();
+      if (typeof sanitizedResult !== "string") {
+        throw new Error("Sanitized result output was not logged");
+      }
+      expect(JSON.parse(sanitizedResult)).toEqual({
+        type: "result",
+        subtype: "success",
+        is_error: false,
+        duration_ms: 434,
+        num_turns: 1,
+        total_cost_usd: 1.23,
+        permission_denials_count: 0,
+        modelUsage: {
+          "claude-opus-5": {
+            contextWindow: 200000,
+            maxOutputTokens: 64000,
+          },
+        },
+      });
+      expect(sanitizedResult).not.toContain("inputTokens");
+      expect(sanitizedResult).not.toContain("costUSD");
+    } finally {
+      consoleLogSpy.mockRestore();
+    }
+  });
+
   test("fails when result subtype is success but is_error is true", async () => {
     const consoleErrorSpy = spyOn(console, "error").mockImplementation(
       () => {},
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -275,6 +275,29 @@ For provider-specific models:
     # ... other inputs
 ```
 
+### 1M context models through an API gateway
+
+When `ANTHROPIC_BASE_URL` points to an Anthropic-compatible API gateway,
+Claude Code may not be able to verify that the gateway supports a model's native
+1M context window and can budget the session at 200K instead. Append the
+`[1m]` selector to explicitly use the 1M context window for supported models,
+including Claude Opus 5 and Claude Sonnet 5:
+
+```yaml
+- uses: anthropics/claude-code-action@v1
+  with:
+    claude_args: |
+      --model "claude-opus-5[1m]"
+    # ... other inputs
+```
+
+Use the same selector when setting a model through `ANTHROPIC_MODEL` or another
+Claude Code model environment variable. The selector is resolved by Claude Code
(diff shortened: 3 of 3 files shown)


ranyhb:
Have u tested that?

ghost:
> Have u tested that?

Thanks for the review. I should have included live-cluster validation before submitting, and I’ll add that evidence.

One naming point I’d like to clarify: on this PR’s exact head commit, operator/pkg/model/translation/gateway-api/translator.go sets the generated GAMMA name to source.Name, and cec_translator.go assigns that value directly to metadata.name.

The existing GAMMA fixture also has metadata.name: echo, and the linked issue reproduces the resource with `kubectl get cec my-app`.

I may be overlooking another generation path. Could you point me to where the `cec-<svcname>` prefix is added? I’ll verify the observed CEC name and the detach/deletion behavior on a live Kind cluster before updating the PR.

ghost:
> > 你測試過嗎？
> 
> 感謝您的審閱。我應該在提交前加入即時叢集驗證，我會添加相關證據。
> 
> 我想澄清一個命名點：在這個 PR 的確切提交中，operator/pkg/model/translation/gateway-api/translator.go 將產生的 GAMMA 名稱設為 source.Name，而 cec_translator.go 將該值直接賦給 metadata.name。
> 
> 現有的 GAMMA fixture 也具有 metadata.name: echo，連結的問題重現了該資源`kubectl get cec my-app`。
> 
> 我可能忽略了另一種生成路徑。您能否指出`cec-<svcname>`前綴是在哪裡加上的？在更新 PR 之前，我會在實際的 Kind 叢集上驗證觀察到的 CEC 名稱以及分離/刪除行為。

I tested the change itself locally on commit 8b59f2b, but I have not tested the live 1M gateway path end-to-end.

Results:

- `bun test base-action/test/run-claude-sdk.test.ts` — 3 passed, 0 failed
- `bun run typecheck` — passed
- `git diff --check` — passed

The focused regression test verifies that the sanitized SDK result retains
`contextWindow` and `maxOutputTokens` while excluding token-usage and
per-model cost fields.

I also ran the full suite on Windows: 842 passed and 32 failed. The failures
were in other test files and were concentrated in Windows/POSIX-sensitive
areas such as Unix path expectations, symlinks, POSIX file modes, bash
availability, and line endings. I am not counting the full suite or the
formatting check as passing.

So to be precise: the sanitized logging behavior introduced by this PR is
tested. I have not independently reproduced an actual
`claude-opus-5[1m]` request through an `ANTHROPIC_BASE_URL` gateway yet.

ashwin-ant (approved):

PR #1624, opened 2026-08-10
fix(images): isolate downloads across concurrent runs

### LINKED ISSUE #1622: Image downloads use a shared /tmp directory and collision-prone filenames across concurrent runs
﻿## Summary

Image downloads are written to a process-wide `/tmp/github-images` directory using filenames based only on `Date.now()` and the image index. On a persistent self-hosted runner, concurrent action jobs can overwrite one another's files.

## Reproduction

1. Run two Claude Code Action jobs concurrently on the same self-hosted runner.
2. Have both jobs process a comment or PR body containing an image.
3. If both downloads reach the filename expression in the same millisecond, or if the same generated name is reused by concurrent invocations, both jobs target the same path such as `/tmp/github-images/image-<timestamp>-0.png`.
4. Let one job finish writing while the other job is still using the path supplied in its image URL map.

A deterministic unit test can reproduce the collision by mocking `Date.now()` to return the same value for two concurrent calls to `downloadCommentImages`.

## Expected behavior

Each action invocation should use an isolated temporary directory and collision-resistant filenames. A job should never be able to replace an image file that another job is about to send to Claude.

## Actual behavior

`src/github/utils/image-downloader.ts:95` hard-codes `/tmp/github-images`, and `src/github/utils/image-downloader.ts:264-267` creates and writes `image-${Date.now()}-${i}${extension}`. The per-invocation URL-to-path map does not prevent another invocation from writing the same path. Files are also not removed after the invocation, so persistent runners accumulate prior job artifacts.

## Why this matters

The wrong image can be supplied to Claude for a different repository or workflow run, producing incorrect analysis. On shared self-hosted runners, this also breaks job isolation and can leave user-provided issue/PR attachments in a shared directory after the job ends.

## Suggested fix

Create a per-invocation directory below `RUNNER_TEMP` (for example with `mkdtemp`), use a UUID/random component or exclusive file creation for each filename, and clean the directory in a `finally` block after all consumers have finished reading the images. Add a concurrent-invocation regression test.

## Related

This is separate from #702 and #1351, which concern the image MIME type, and from the merged #1588, which concerns pairing source URLs with signed asset URLs.

abhinavkr26104:
﻿## Summary

- isolate downloaded GitHub images in a unique temporary directory per invocation
- prefer RUNNER_TEMP and fall back to the operating system temp directory
- add a regression test covering concurrent invocations that share the same timestamp

This prevents concurrent action jobs on a persistent self-hosted runner from overwriting one another's downloaded image files.

Fixes #1622

## Testing

- bun test test/image-downloader.test.ts -t "download directory|isolate concurrent"
- bun run typecheck
- bunx prettier --check src/github/utils/image-downloader.ts test/image-downloader.test.ts
- git diff --check
- bun test (850 passed, 44 pre-existing Windows/platform/environment failures, including hard-coded POSIX image-path assertions, symlink/permission checks, CRLF-sensitive metadata/README tests, and environment-dependent cases)

### DIFF
--- a/src/github/utils/image-downloader.ts
+++ b/src/github/utils/image-downloader.ts
@@ -1,4 +1,5 @@
 import fs from "fs/promises";
+import { tmpdir } from "os";
 import path from "path";
 import type { Octokits } from "../api/client";
 import { GITHUB_SERVER_URL } from "../api/config";
@@ -92,9 +93,8 @@ export async function downloadCommentImages(
   comments: CommentWithImages[],
 ): Promise<Map<string, string>> {
   const urlToPathMap = new Map<string, string>();
-  const downloadsDir = "/tmp/github-images";
-
-  await fs.mkdir(downloadsDir, { recursive: true });
+  const tempRoot = process.env.RUNNER_TEMP || tmpdir();
+  const downloadsDir = await fs.mkdtemp(path.join(tempRoot, "github-images-"));
 
   const commentsWithImages: Array<{
     comment: CommentWithImages;
--- a/test/image-downloader.test.ts
+++ b/test/image-downloader.test.ts
@@ -12,6 +12,7 @@ import fs from "fs/promises";
 import { downloadCommentImages } from "../src/github/utils/image-downloader";
 import type { CommentWithImages } from "../src/github/utils/image-downloader";
 import type { Octokits } from "../src/github/api/client";
+import path from "path";
 
 // Asset URLs and their signed download URLs share the asset's GUID.
 const GUID_1 = "f871c23e-a84d-4f1f-b9a0-86626c63f161";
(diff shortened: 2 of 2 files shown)


ashwin-ant (dismissed):

ashwin-ant:
@abhinavkr26104 mind fixing the conflict?

abhinavkr26104:
> @abhinavkr26104 mind fixing the conflict?

i'll fix it

abhinavkr26104:
Conflict resolved by merging current main in 2d60496. The isolation change is preserved alongside the newer timeout and size-limit handling. I also made the image-path assertions cross-platform. Verified with 28/28 focused tests, TypeScript typecheck, Prettier, and git diff --check.

PR #1630, opened 2026-08-10
fix(file-ops): preserve binary file contents

### LINKED ISSUE #1627: commit_files corrupts binary files whose extensions are not in its hardcoded allowlist
﻿## Summary

The ``commit_files`` MCP tool decides whether a file is binary using a short extension allowlist. Any binary file with an unlisted extension is read as UTF-8 text and sent to GitHub as a text tree entry, corrupting its bytes.

## Reproduction

1. Enable API commit signing so the ``github_file_ops`` MCP server is available.
2. Create a binary file in the workspace with an extension not listed in the allowlist, such as ``assets/sample.bmp``, ``data.sqlite``, ``archive.docx``, or ``payload.dat``.
3. Ask Claude to call ``mcp__github_file_ops__commit_files`` for that path.
4. Compare the committed blob bytes with the original file.

## Expected behavior

Every committed file should preserve its contents byte-for-byte, regardless of filename extension.

## Actual behavior

At ``src/mcp/github-file-ops-server.ts:261-305``, binary handling is selected only for:

``````
png jpg jpeg gif webp ico pdf zip tar gz exe bin woff woff2 ttf eot
``````

For every other extension, the code executes ``readFile(fullPath, "utf-8")`` and sends the resulting string as the Git tree ``content``. Invalid UTF-8 bytes are decoded/replaced before the GitHub API receives them.

## Why this matters

The tool can silently commit corrupted assets, databases, archives, fonts, media, and other non-text files. The commit may succeed while downstream consumers receive unusable data.

## Suggested fix

Detect text/binary content from bytes rather than relying on an extension allowlist. Send binary content through the blob API and add regression coverage for an unlisted binary extension.

## Related

This is separate from #796 and active PR #1006, which address path validation, and active PR #1200, which addresses symlink detection.

abhinavkr26104:
## Summary

- remove extension-based binary detection from `commit_files`
- read every file as raw bytes and create a base64 Git blob
- add regression coverage for binary bytes with an unlisted extension

Fixes #1627

## Why

`commit_files` previously decoded files outside a short binary extension allowlist as UTF-8 text. Invalid bytes were replaced before being sent to GitHub, silently corrupting committed assets. All files now use the Git blob API with base64 encoding, preserving their bytes regardless of filename.

## Testing

- `bun test test/github-file-content.test.ts`
- `bun run typecheck`
- targeted Prettier check for changed files
- `git diff --check`
- `bun test` — 851 passed, 44 pre-existing Windows/platform/environment failures
- `bun run format:check` — pre-existing repository-wide formatting failures in 183 files; changed files pass targeted Prettier checks

### DIFF
--- a/src/mcp/github-file-content.ts
+++ b/src/mcp/github-file-content.ts
@@ -1,0 +1,6 @@
+export function createBase64BlobPayload(content: Uint8Array) {
+  return {
+    content: Buffer.from(content).toString("base64"),
+    encoding: "base64" as const,
+  };
+}
--- a/src/mcp/github-file-ops-server.ts
+++ b/src/mcp/github-file-ops-server.ts
@@ -9,6 +9,7 @@ import { constants } from "fs";
 import fetch from "node-fetch";
 import { GITHUB_API_URL } from "../github/api/config";
 import { validatePathWithinRepo } from "./path-validation";
+import { createBase64BlobPayload } from "./github-file-content";
 import { updateGitReference } from "./update-git-reference";
 
 type GitHubRef = {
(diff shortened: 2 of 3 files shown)


abhinavkr26104:
Closing as superseded by #1633, which fixed the same extension-based binary detection issue with content-based detection and regression coverage.

PR #1631, opened 2026-08-10
fix(images): bound attachment download size

### LINKED ISSUE #1628: Image attachment downloads have no size limit and buffer untrusted files fully in memory
﻿## Summary

The GitHub image downloader accepts the entire response body with ``arrayBuffer()`` and writes it to disk without checking ``Content-Length``, enforcing a maximum size, or streaming with a bounded buffer.

## Reproduction

1. Process an issue or pull request containing one or more GitHub user-attachment images.
2. Make a signed attachment response advertise or return a very large body, or include many large attachments in the context.
3. Observe that ``downloadCommentImages`` reads each complete body into memory before it can decide whether to save it.

A deterministic unit test can mock ``response.arrayBuffer()`` to return a body larger than the configured maximum and assert that it is rejected before being written.

## Expected behavior

Image downloads should have a documented maximum size and reject oversized responses before consuming unbounded memory or disk space. A single oversized attachment should be skipped without preventing the rest of the context from being processed.

## Actual behavior

At ``src/github/utils/image-downloader.ts:244-267``, the code awaits ``fetch``, then calls ``arrayBuffer()`` for the complete response and writes the resulting buffer. No response-size check or streaming limit exists.

## Why this matters

Attachment URLs originate in untrusted issue, PR, and comment content. A large or repeated set of attachments can exhaust the runner's memory or temporary disk space during context preparation, causing the action to fail before Claude starts.

## Suggested fix

Enforce a maximum using ``Content-Length`` when available and a bounded streaming reader for responses without that header. Add tests for both an oversized declared length and a stream that exceeds the limit.

## Related

This is separate from #702 and #1351, which concern image MIME detection, #1588, which concerns asset matching, and #1623, which adds a time bound for stalled downloads.

abhinavkr26104:
## Summary

- enforce a documented 10 MiB maximum per image download
- reject oversized `Content-Length` responses before reading them
- consume streamed responses incrementally and cancel when the limit is exceeded
- skip oversized images through the existing per-image error handling
- add declared-length and chunked-response regression tests

Fixes #1628

## Why

Image attachment responses are external input. The downloader previously awaited `arrayBuffer()` without a size check, allowing a large attachment to consume unbounded memory and disk space. The new streaming reader keeps accumulation bounded and never writes an oversized response.

## Testing

- `bun test test/image-downloader.test.ts -t "size|streamed"`
- `bun run typecheck`
- targeted Prettier check for changed files
- `git diff --check`
- `bun test` — 851 passed, 44 pre-existing Windows/platform/environment failures
- `bun run format:check` — pre-existing repository-wide formatting failures in 182 unchanged files; changed files pass targeted Prettier checks

### DIFF
--- a/src/github/utils/image-downloader.ts
+++ b/src/github/utils/image-downloader.ts
@@ -28,6 +28,9 @@ function extractAssetGuid(url: string): string | undefined {
 
 const SIGNED_URL_HOST = "private-user-images.githubusercontent.com";
 
+/** Maximum size of one downloaded image, enforced before it is written. */
+export const MAX_IMAGE_SIZE_BYTES = 10 * 1024 * 1024;
+
 // Signed download URLs have the shape /<owner-id>/<asset-id>-<guid>.<ext>.
 // The GUID must come from the resolved filename, not from anywhere in the raw
 // string, so text that merely embeds a GUID cannot claim another asset.
@@ -248,8 +251,7 @@ export async function downloadCommentImages(
             );
           }
 
-          const arrayBuffer = await imageResponse.arrayBuffer();
-          const buffer = Buffer.from(arrayBuffer);
+          const buffer = await readImageResponse(imageResponse);
 
           // GitHub user-attachment URLs (/user-attachments/assets/<uuid>) carry
           // no file extension, so the URL-based guess silently falls back to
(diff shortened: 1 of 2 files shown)


abhinavkr26104:
Closing as superseded by d721746, which landed the bounded image-download implementation and regression coverage on main.

PR #1637, opened 2026-08-11
fix(base-action): add ~/.local/bin to GITHUB_PATH after default install

### LINKED ISSUE #1634: claude-code-base-action's default install path never adds ~/.local/bin to $GITHUB_PATH (custom-executable branch does)
### What'\''s wrong

`action.yml`'\''s "Install Claude Code" step has two branches. The custom-executable branch correctly adds the binary'\''s directory to `$GITHUB_PATH`:

    else
      echo "Using custom Claude Code executable: $PATH_TO_CLAUDE_CODE_EXECUTABLE"
      CLAUDE_DIR=$(dirname "$PATH_TO_CLAUDE_CODE_EXECUTABLE")
      echo "$CLAUDE_DIR" >> "$GITHUB_PATH"

The default auto-install branch (curl -fsSL https://claude.ai/install.sh | bash) does not:

    if [ -z "$PATH_TO_CLAUDE_CODE_EXECUTABLE" ]; then
      ...
      curl -fsSL https://claude.ai/install.sh | bash -s -- "$CLAUDE_CODE_VERSION"
      ...
      echo "Claude Code installed successfully"
      # nothing added to $GITHUB_PATH

install.sh installs the native binary to ~/.local/bin and prints:

    ⚠ Setup notes:
      ● Native installation exists but ~/.local/bin is not in your PATH. Run:
        echo '\''export PATH="$HOME/.local/bin:$PATH"'\'' >> your shell config file && source your shell config file

That instruction does nothing in a GitHub Actions job, each step runs as `bash --noprofile --norc`, no dotfile is ever sourced, and PATH additions only persist across steps via $GITHUB_PATH. The very next step ("Run Claude Code Action") then fails:

    error: Executable not found in $PATH: "claude"

### Reproduction

Any workflow using the action'\''s default auto-install (no `path_to_claude_code_executable` set) on a runner where ~/.local/bin isn'\''t already on PATH by some other means. We hit this on a self-hosted ARC-based runner, pinned to `v0.0.63` (`e8132bc5e637a42c27763fc757faa37e1ee43b34`).

### Suggested fix

Mirror the custom-executable branch: after the curl install succeeds, add:

    echo "$HOME/.local/bin" >> "$GITHUB_PATH"

(or wherever install.sh actually places the binary, if that'\''s not guaranteed stable).

### Related but not identical

#477 and #521 show the same error text but a maintainer suggested a different cause there (a self-installed `bun` conflicting with the action'\''s setup). We don'\''t install bun ourselves and still hit this, so filing separately rather than assuming one root cause explains both. Worth checking whether #477/#521'\''s reporters also hit this same gap independently of their bun setup.

### Impact worth flagging

Because the job step fails but the overall workflow run can still report "success" (depending on how the calling workflow is structured), this failed silently for us across multiple days before being caught by reading a job log directly, rather than any dashboard signal.

AliAltivate:
The default install branch (curl | install.sh) never adds the installed binary's directory to $GITHUB_PATH, unlike the custom-executable branch which appends $CLAUDE_DIR. install.sh places claude in ~/.local/bin, and the next step runs with bash --noprofile --norc, so 'claude' was not on PATH -> 'Executable not found in /Users/ali/.kimi-code/bin:/Users/ali/.bun/bin:/Users/ali/.bun/bin:/Users/ali/bin:/Users/ali/Downloads/google-cloud-sdk/bin:/Users/ali/Coding:/Users/ali/.local/bin:/Users/ali/Library/Python/3.9/bin:/Users/ali/.local/bin:/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/pkg/env/global/bin:/Users/ali/.lmstudio/bin'.

Mirrors the custom-executable branch's existing style. The top-level action already does this in src/entrypoints/run.ts installClaudeCode(); only base-action/action.yml was missing it.

Fixes #1634. YAML parse + prettier pass; full test suite has pre-existing failures unrelated to this YAML-only change (confirmed on pristine upstream/main).

### DIFF
--- a/base-action/action.yml
+++ b/base-action/action.yml
@@ -165,6 +165,8 @@ runs:
             sleep 5
           done
           echo "Claude Code installed successfully"
+          # Add the directory containing the installed binary to PATH (install.sh places it in ~/.local/bin)
+          echo "$HOME/.local/bin" >> "$GITHUB_PATH"
         else
           echo "Using custom Claude Code executable: $PATH_TO_CLAUDE_CODE_EXECUTABLE"
           # Add the directory containing the custom executable to PATH


AliAltivate:
Pulling this back while I pause contribution activity.

PR #1638, opened 2026-08-11
fix: report neutral conclusion when workflow validation is skipped

### LINKED ISSUE #1632: Workflow-validation skip reports conclusion: success — indistinguishable from a real pass (follow-up to #1343)
Follow-up to #1343, which unified the two validation-skip code paths on `success`. That fixed the inconsistency, but the chosen value has a security-relevant side effect: **a skipped review is now indistinguishable from a passing review** — including to branch protection.

#1343 itself proposed the safer alternative in its final paragraph ("both paths could surface as `skipped` (or `neutral`)"). This issue argues that alternative should be preferred.

## Observed

Confirmed on `claude-code-action@v1`, private repo, August 2026 — both the first-add and modification paths now return `success`:

| Run | PR changed | Duration | Conclusion | Log contains |
|---|---|---|---|---|
| `31280931811` | adds a workflow | 14s | `success` | `Exiting due to workflow validation skip` |
| `31448510042` | **modifies** `claude-review.yml` | 16s | `success` | `Exiting due to workflow validation skip` |

For contrast, a genuine review run on the same repo takes ~2m39s–5m39s and reports real `num_turns` / `total_cost_usd`.

## Why `success` is the wrong value here

The anti-tamper check itself is sound — a PR should not be able to rewrite the reviewer that gates it. The problem is only the conclusion it reports.

**Any PR that edits the invoking workflow gets a green `claude-review` check with nothing reviewed — including a PR that weakens or disables the reviewer.** The check is structurally incapable of failing on exactly the PRs where review matters most.

Because the conclusion is `success`, this is invisible to:
- the checks UI (a green ✓ identical to a real pass)
- **branch protection** — `claude-review` as a required status check is satisfied by a run that never executed
- any automation gating on check conclusions (e.g. Dependabot auto-merge)

The only way to distinguish the two today is to open the run log and grep for `Exiting due to workflow validation skip`, or notice the ~10x duration difference. We were caught by this in practice: a workflow-fix PR showed green, and it took reading the log to establish that the check had done nothing.

## Suggested fix

Report the skip as **`neutral`** (GitHub's conclusion for "ran but intentionally reached no verdict") rather than `success`.

This preserves everything #1343 asked for — no spurious red ✗, both paths consistent, no "something is broken" impression — while keeping the outcome distinguishable from a real pass. Concretely, `neutral`:

- does **not** satisfy a required status check, so a workflow-editing PR can no longer merge on the strength of a review that never ran
- renders distinctly in the checks UI
- is already the conventional signal for "skipped by design"

If `neutral` is considered too disruptive for existing users relying on the green check, a lower-impact alternative would be to keep `success` but set a distinguishable check **output/summary** (e.g. title "Skipped — workflow modified") so tooling has something to key on without reading raw logs.

## Reproduction

1. Add `claude-code-action@v1` to a repo and merge it.
2. Open a PR that modifies that workflow file (change `--max-turns`, say).
3. Observe: check reports `success` in seconds; the log contains `Exiting due to workflow validation skip`; no review was performed.
4. Make `claude-review` a required status check and observe that the PR is mergeable on the strength of that skip.

## Workaround for others hitting this

Until upstream changes, a separate workflow filtered on the invoking workflow's path can post a notice on affected PRs. Worth noting the constraint that makes this awkward: the action cannot be fixed or tested from a PR that edits it, because that PR self-skips — the change has to land on the default branch first and be exercised on an unrelated PR.

AliAltivate:
When workflow validation is skipped, the action returned without setting a conclusion, so the run read as an indistinguishable green pass. Now sets conclusion=neutral on the skip path, matching the repo's existing conclusion model (success/failure/neutral). action.yml output description updated accordingly.

Note: an action step can't change the GitHub check-run conclusion itself (only ::error+exit-1 or exit-0); neutral is reported via the action's conclusion output, which is what tooling and downstream steps can consume. If maintainers want the check itself to not satisfy branch protection, that needs a workflow-level conditional.

Fixes #1632. typecheck + targeted tests pass; 7 pre-existing network-timeout flakes in restoreConfigFromBase confirmed on pristine upstream/main.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -168,7 +168,7 @@ inputs:
 
 outputs:
   conclusion:
-    description: "Execution status of Claude Code ('success' or 'failure')"
+    description: "Execution status of Claude Code ('success', 'failure', or 'neutral' when skipped by workflow validation)"
     value: ${{ steps.run.outputs.conclusion }}
   execution_file:
     description: "Path to the Claude Code execution output file"
--- a/src/entrypoints/run.ts
+++ b/src/entrypoints/run.ts
@@ -175,6 +175,9 @@ async function run() {
     } catch (error) {
       if (error instanceof WorkflowValidationSkipError) {
         core.setOutput("skipped_due_to_workflow_validation_mismatch", "true");
+        // Report a distinct conclusion ("neutral") so a skipped review is not
+        // indistinguishable from a real pass. See #1632.
+        core.setOutput("conclusion", "neutral");
         console.log("Exiting due to workflow validation skip");
         return;
       }


AliAltivate:
Pulling this back while I pause contribution activity.

PR #1654, opened 2026-08-14
fix(mcp): support mcp__github aggregate selector

### LINKED ISSUE #723: MCP tool configuration `mcp__github` does not enable GitHub MCP servers as documented
# **Describe the bug**

The [documentation](https://code.claude.com/docs/en/iam) states that `mcp__github` in `allowedTools` will approve ALL GitHub tools. This likely works correctly for **Claude Code CLI**, but **Claude Code Action has a different implementation** that is undocumented.

In Claude Code Action, specifying `mcp__github` does not initialize the GitHub MCP servers. Claude falls back to using the `gh` CLI via Bash instead of MCP tools, or reports that MCP tools are "not available."

After investigating the source code at [`src/mcp/install-mcp-server.ts`](https://github.com/anthropics/claude-code-action/blob/main/src/mcp/install-mcp-server.ts), I found the root cause:

```typescript
const hasGitHubMcpTools = allowedToolsList.some((tool) =>
  tool.startsWith("mcp__github__"),
);
```

The code checks if any tool **starts with** `mcp__github__` (with trailing double underscore). When the user specifies `mcp__github`, the check fails:

```javascript
"mcp__github".startsWith("mcp__github__")  // Returns FALSE
```

This causes `hasGitHubMcpTools` to be `false`, so the GitHub MCP servers are never initialized.

**The core issue:** The documentation at [code.claude.com/docs/en/iam](https://code.claude.com/docs/en/iam) describes behavior for Claude Code CLI, but Claude Code Action has its own MCP server initialization logic that requires a different configuration. This difference is not documented anywhere.

# **To Reproduce**

1. Create a GitHub Actions workflow with `--allowedTools "Bash,mcp__github"`
2. Trigger the workflow on a pull request
3. In the prompt, instruct Claude to use GitHub MCP tools like `mcp__github__get_pull_request_diff` or `mcp__github_inline_comment__create_inline_comment`
4. Observe that Claude either:
   - Reports the MCP tools are "not available", or
   - Falls back to using `gh` CLI commands via Bash

# **Expected behavior**

Based on the documentation at [code.claude.com/docs/en/iam](https://code.claude.com/docs/en/iam):
> ✅ Use: `mcp__github` (approves ALL GitHub tools)

Specifying `mcp__github` should enable all GitHub MCP servers and their tools, including:
- `mcp__github__*` tools (GitHub API operations)
- `mcp__github_inline_comment__*` tools (PR inline comments)
- `mcp__github_comment__*` tools (comment management)

# **Workflow yml file**

```yaml
name: Claude Code Review

on:
  pull_request:
    types: [opened, labeled]

permissions:
  contents: read
  pull-requests: write
  actions: read
  id-token: write

jobs:
  claude-review:
    if: contains(github.event.pull_request.labels.*.name, 'claude-review')
    runs-on: ubuntu-latest
    steps:
      - name: Checkout code
        uses: actions/checkout@v5
        with:
          fetch-depth: 0

      - name: Claude Code Base Action
        with:
          claude_args: |
            --model claude-sonnet-4-20250514
            --allowedTools "Bash,Read,Glob,Grep,mcp__github"
          prompt: |
            Please review this PR using GitHub MCP tools like:
            - mcp__github__get_pull_request_diff
            - mcp__github_inline_comment__create_inline_comment
```

# **API Provider**

[ ] AWS Bedrock
[ ] GCP Vertex

# **Additional context**

## **Suggested Fix (Option A - Fix implementation to match documentation):**

Update the detection logic in `install-mcp-server.ts` to also match `mcp__github` without trailing underscore, making Claude Code Action behavior consistent with Claude Code CLI and the documentation:

```typescript
const hasGitHubMcpTools = allowedToolsList.some((tool) =>
  tool === "mcp__github" || tool.startsWith("mcp__github__"),
);

const hasInlineCommentTools = allowedToolsList.some((tool) =>
  tool === "mcp__github_inline_comment" || tool.startsWith("mcp__github_inline_comment__"),
);
```

## **Suggested Fix (Option B - Add Claude Code Action specific documentation):**

Add documentation specific to Claude Code Action explaining that it has different MCP server initialization requirements:
- `mcp__github__` (not `mcp__github`) or a full tool name to enable GitHub MCP server
- `mcp__github_inline_comment__` or a full tool name to enable inline comment server

# **Documentation Enhancement Request:**

The documentation would also benefit from:
1. Clear distinction between Claude Code CLI and Claude Code Action MCP behavior
2. A complete list of available MCP servers in Claude Code Action and their tool prefixes
3. A list of available tools under each MCP server
4. Clarification on conditional logic (e.g., `github_inline_comment` only available in PR context)
5. Configuration examples for common use cases like PR reviews with inline comments

# **Contribution Offer**

If this issue is confirmed as valid, I would be happy to contribute a PR to fix either the implementation or the documentation (whichever approach is preferred by the maintainers).

binaydhakal:
## Summary

- recognize the documented `mcp__github` aggregate selector when deciding whether to install the GitHub MCP server
- preserve that selector when tag mode forwards user-approved GitHub MCP tools
- add regression coverage for agent mode, tag mode, qualified tool names, and look-alike names

Fixes #723.

## Testing

- `bun test` (913 passed)
- `bun run typecheck`
- `bun run format:check`
- `git diff --check`

### DIFF
--- a/src/mcp/install-mcp-server.ts
+++ b/src/mcp/install-mcp-server.ts
@@ -87,8 +87,8 @@ export async function prepareMcpConfig(
       tool.startsWith("mcp__github_comment__"),
     );
 
-    const hasGitHubMcpTools = allowedToolsList.some((tool) =>
-      tool.startsWith("mcp__github__"),
+    const hasGitHubMcpTools = allowedToolsList.some(
+      (tool) => tool === "mcp__github" || tool.startsWith("mcp__github__"),
     );
 
     const hasInlineCommentTools = allowedToolsList.some((tool) =>
--- a/src/modes/tag/index.ts
+++ b/src/modes/tag/index.ts
@@ -121,8 +121,8 @@ export async function prepareTagMode({
   );
 
   const userClaudeArgs = process.env.CLAUDE_ARGS || "";
-  const userAllowedMCPTools = parseAllowedTools(userClaudeArgs).filter((tool) =>
-    tool.startsWith("mcp__github_"),
+  const userAllowedMCPTools = parseAllowedTools(userClaudeArgs).filter(
+    (tool) => tool === "mcp__github" || tool.startsWith("mcp__github_"),
   );
 
   const gitPushWrapper = `${process.env.GITHUB_ACTION_PATH}/scripts/git-push.sh`;
--- a/test/install-mcp-server.test.ts
+++ b/test/install-mcp-server.test.ts
@@ -154,6 +154,39 @@ describe("prepareMcpConfig", () => {
     );
   });
 
+  test("should include github MCP server when mcp__github is allowed", async () => {
+    const result = await prepareMcpConfig({
+      githubToken: "test-token",
+      owner: "test-owner",
+      repo: "test-repo",
+      branch: "test-branch",
+      baseBranch: "main",
+      allowedTools: ["mcp__github"],
+      mode: "agent",
+      context: mockContext,
+    });
+
+    const parsed = JSON.parse(result);
+    expect(parsed.mcpServers.github).toBeDefined();
+    expect(parsed.mcpServers.github.command).toBe("docker");
+  });
+
+  test("should not treat look-alike tool names as github MCP tools", async () => {
+    const result = await prepareMcpConfig({
+      githubToken: "test-token",
+      owner: "test-owner",
+      repo: "test-repo",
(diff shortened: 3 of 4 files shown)


Flawed-it:
Yeah its still bad

PR #1658, opened 2026-08-15
fix(restore): gracefully handle git fetch failures when credentials unavailable

### LINKED ISSUE #1559: git fetch in restoreConfigFromBase fails when use_commit_signing=true and persist-credentials=false
## Summary

When `use_commit_signing: true` is set and the workflow checks out with `persist-credentials: false`, agent mode fails during the PR-head config restore step with:

```text
Restoring .claude, .mcp.json, ... from origin/main (PR head is untrusted)
fatal: could not read Username for 'https://github.com': No such device or address
Action failed with error: Command failed: git fetch origin main --depth=1 --no-recurse-submodules
```

This is reproducible on private/internal repositories (anonymous `git fetch` is rejected).

## Root cause

1. On PR contexts, `restoreConfigFromBase` unconditionally runs `git fetch origin <baseBranch>` (`src/github/operations/restore-config.ts`).
2. With `use_commit_signing: true`, agent mode sets `useApiCommitSigning = true` and **does not** call `configureGitAuth` (`src/modes/agent/index.ts`), so no git credentials are configured by the action.
3. If the workflow also uses `actions/checkout` with `persist-credentials: false`, the checkout-provided `includeIf` credentials are removed.
4. Result: no credentials for `git fetch`. `execFileSync` has no try/catch, so the whole action fails even though a safe fallback (leave sensitive paths deleted) already exists for missing base paths.

## Why this is surprising

Commit signing mode intentionally avoids embedding tokens into the git remote because commits go through the GitHub API. The restore path still requires a git-level network operation against a private remote, which contradicts that design.

## Suggested fixes (any one would help)

- Call a minimal git-auth setup (or set `http.extraheader` / credential helper from `GITHUB_TOKEN`) before `restoreConfigFromBase` even when commit signing is enabled.
- Or avoid `git fetch` when credentials are unavailable and fall back to the already-safe “paths stay deleted” behavior instead of fatal-exiting.
- Or document that `persist-credentials: false` is incompatible with PR-triggered runs under `use_commit_signing: true`.

## Environment

- Action: `anthropics/claude-code-action@v1` (resolved SHA around `be7b93b`, Claude Code 2.1.220)
- Trigger: `pull_request` closed/merged (agent mode with explicit `prompt`)
- Inputs: `use_commit_signing: true`, `base_branch: main`
- Checkout: `actions/checkout` with `persist-credentials: false`
- Repo visibility: internal/private

## Workaround used downstream

Keep checkout credentials (default `persist-credentials: true`) so the restore fetch can authenticate with the workflow `GITHUB_TOKEN`, and keep writes on an App token via `github_token`.

anishesg:
## Problem

When `use_commit_signing: true` is set and the workflow checks out with `persist-credentials: false`, a mode fails during config restore with a fatal git fetch error on private repositories. The action has a design inconsistency: commit signing mode intentionally avoids embedding tokens into git remotes (commits go through GitHub API), but the config restore step in `restoreConfigFromBase` still performs a git-level network operation that requires authentication.

## Fix

Wrapped the git fetch call in `src/github/operations/restore-config.ts` with try/catch error handling. When git fetch fails due to missing credentials, the action now logs a warning and returns early, falling back to the already-safe behavior of leaving sensitive paths deleted rather than fatal-exiting. This preserves the security guarantee (PR-controlled sensitive config never executes) while preventing the action from crashing in environments where git credentials are unavailable.

## Testing

Added a regression test that simulates fetch failure by removing the git remote before calling `restoreConfigFromBase`. The test verifies that the function completes without throwing and leaves sensitive paths deleted while preserving the PR snapshot in `.pr/`.

Fixes #1559

### DIFF
--- a/src/github/operations/restore-config.ts
+++ b/src/github/operations/restore-config.ts
@@ -304,20 +304,35 @@ export function restoreConfigFromBase(baseBranch: string): void {
 
   // --no-recurse-submodules: explicitly suppress submodule fetching regardless of
   // fetch.recurseSubmodules config. Defense-in-depth alongside the delete above.
-  execFileSync(
-    "git",
-    [
-      "fetch",
-      "origin",
-      baseBranch,
-      ...fetchDepthArgs(1),
-      "--no-recurse-submodules",
-    ],
-    {
-      stdio: "inherit",
-      env: process.env,
-    },
-  );
+  //
+  // Wrap fetch in try/catch: when credentials are unavailable (e.g., commit
+  // signing mode + persist-credentials=false), git fetch will fail. In that case,
+  // fall back to the already-safe behavior of leaving sensitive paths deleted
+  // rather than fatal-exiting the action.
+  try {
+    execFileSync(
+      "git",
+      [
+        "fetch",
+        "origin",
+        baseBranch,
+        ...fetchDepthArgs(1),
+        "--no-recurse-submodules",
+      ],
+      {
+        stdio: "inherit",
+        env: process.env,
+      },
+    );
+  } catch (error) {
+    console.warn(
+      `Warning: git fetch origin/${baseBranch} failed (likely missing credentials). ` +
+        `Sensitive paths will remain deleted (safe fallback).`,
+    );
+    // When fetch fails, skip checkout attempts below — base content is unavailable,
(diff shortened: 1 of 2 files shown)


anishesg:

PR #1659, opened 2026-08-15
fix(examples): change pull-requests permission from read to write in PR review examples

### LINKED ISSUE #1562: install-github-app generates read-only permissions, so the action posts nothing and the run still goes green
## Summary

The workflow files generated by `/install-github-app` declare `pull-requests: read` and `issues: read`. With those permissions the action runs to completion, the job goes green, and no comment is ever posted — because it has no permission to write one.

This is a silent failure: there is no error in the logs, no annotation, and no indication that anything is wrong. The natural assumption is that the token, the App installation or the trigger is misconfigured, and the actual cause is two words in a file the installer wrote.

## Reproduce

1. Run `/install-github-app` in Claude Code against a repository
2. Merge the PR it opens
3. Comment `@claude hello` on any issue or PR

Expected: Claude replies.
Actual: a workflow run appears in the Actions tab, completes successfully, and nothing is posted.

## Generated permissions

Both files ship the same block. From `.github/workflows/claude.yml`:

```yaml
    permissions:
      contents: read
      pull-requests: read      # cannot post a comment
      issues: read             # cannot post a comment
      id-token: write
      actions: read
```

And `.github/workflows/claude-code-review.yml`:

```yaml
    permissions:
      contents: read
      pull-requests: read      # cannot post a review
      issues: read
      id-token: write
```

Changing both to `write` fixes it immediately.

## Why this is worth fixing in the template

The principle of least privilege is the right instinct, but here it conflicts with what the action is for. `claude.yml` exists to answer `@claude` mentions, and `claude-code-review.yml` exists to post reviews — writing comments is the entire purpose of both. Shipping them read-only guarantees that the first run of a fresh installation does nothing.

Anyone installing the integration for the first time hits this, and the symptom points away from the cause.

## Suggested fixes, in order of preference

1. **Generate `write` for `pull-requests` and `issues`** in both templates. This is what the workflows need to function.
2. If read-only defaults are deliberate, **fail loudly**: have the action detect that it lacks write permission and emit a workflow error rather than exiting successfully. A red run with "cannot post comment: missing pull-requests: write" would cost users nothing to diagnose.
3. At minimum, **say so in the generated file** — a comment above the block along the lines of `# change to write for Claude to post comments` would be enough.

Option 2 has value on its own even if the templates change, since users hand-edit these files and can reintroduce the same state.

## Environment

- Installed via `/install-github-app`
- Authentication: `claude_code_oauth_token`
- Action: `anthropics/claude-code-action@v1`
- Private repositories, personal account

Happy to open a PR against the templates if that would help.

anishesg:
## Problem

The PR review example workflows in `examples/pr-review-filtered-authors.yml` and `examples/pr-review-filtered-paths.yml` declare `pull-requests: read` permissions. These workflows use `gh pr comment` bash commands to post review feedback, which requires write access to pull requests. With read-only permissions, the workflows complete successfully but fail silently when attempting to post comments.

## Fix

Changed `pull-requests: read` to `pull-requests: write` in both example files. This matches the permissions in other review examples like `pr-review-comprehensive.yml` and ensures the bash tools (`gh pr comment`) can actually post their output.

## Testing

Verified the permission blocks now match the working examples and align with the action's requirement for write access when posting comments or reviews.

Fixes #1562

### DIFF
--- a/examples/pr-review-filtered-authors.yml
+++ b/examples/pr-review-filtered-authors.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      pull-requests: read
+      pull-requests: write
       id-token: write
     steps:
       - name: Checkout repository
--- a/examples/pr-review-filtered-paths.yml
+++ b/examples/pr-review-filtered-paths.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      pull-requests: read
+      pull-requests: write
       id-token: write
     steps:
       - name: Checkout repository


anishesg:
closing for now to keep the PR list tidy. reopen anytime.

PR #1672, opened 2026-08-15
fix: resync collect-inputs map with action.yml and guard against drift

### LINKED ISSUE #1664: collect-inputs.ts input map has drifted from action.yml: 11 removed inputs tracked, 19 current inputs missing (including prompt)
## Summary

The input map in [`src/entrypoints/collect-inputs.ts`](https://github.com/anthropics/claude-code-action/blob/d721746d683d812e669ce117cebe55a85fbd9c3e/src/entrypoints/collect-inputs.ts#L2-L34) has drifted out of sync with the inputs actually declared in `action.yml`. It currently tracks **11 inputs that no longer exist** and **omits 19 that do**, including `prompt` — the input that determines whether the action runs in tag mode or agent mode.

Because this function's output becomes the `GITHUB_ACTION_INPUTS` environment variable ([`base-action/src/parse-sdk-options.ts#L280-L282`](https://github.com/anthropics/claude-code-action/blob/d721746d683d812e669ce117cebe55a85fbd9c3e/base-action/src/parse-sdk-options.ts#L280-L282)), the usage signal it produces is currently incomplete and partly meaningless.

## Reproduction

Diffing the declared inputs in `action.yml` against the keys in `inputDefaults`:

```console
$ awk '/^inputs:/,/^outputs:/' action.yml | grep -E "^  [a-z_]+:" | tr -d ' :' | sort > declared.txt
$ grep -oE '^    [a-z_]+:' src/entrypoints/collect-inputs.ts | tr -d ' :' | sort > collected.txt

$ comm -13 declared.txt collected.txt   # tracked but no longer declared
allowed_tools
anthropic_model
claude_env
custom_instructions
direct_prompt
disallowed_tools
fallback_model
max_turns
mode
model
override_prompt

$ comm -23 declared.txt collected.txt   # declared but not tracked
allowed_non_write_users
bot_id
bot_name
branch_name_template
claude_args
display_report
exclude_comments_by_actor
include_comments_by_actor
include_fix_links
path_to_bun_executable
path_to_claude_code_executable
plugin_marketplaces
plugins
prompt
show_full_output
track_progress
use_bedrock
use_foundry
use_vertex
```

All 11 stale entries are v0.x inputs removed in the v1.0 migration — they are listed as deprecated in `docs/usage.md` and `docs/configuration.md`.

## Impact

`collectActionInputsPresence()` reports, for each input, whether the user set it to something other than its default. The current map means that signal cannot observe:

- **`prompt`** — the single input that selects agent mode over tag mode (`src/modes/detector.ts`). Presence of a prompt is arguably the most significant thing about any given invocation, and it is not tracked.
- **`claude_args`** — the primary v1.0 configuration surface, which replaced `allowed_tools`, `disallowed_tools`, `max_turns`, `model`, and `fallback_model`. All five predecessors are still tracked; their replacement is not.
- **`use_bedrock` / `use_vertex` / `use_foundry`** — which cloud provider the run targets.
- **`track_progress`**, **`plugins`**, **`plugin_marketplaces`**, **`allowed_non_write_users`** — all newer features.

Meanwhile the 11 stale keys can never be anything but `false`, since `ALL_INPUTS` (populated from `toJson(inputs)` at `action.yml#L317`) will never contain a key that `action.yml` does not declare. They are pure noise in the output.

## Related: `inputs.mode` is referenced but never declared

Same root cause. [`action.yml#L290`](https://github.com/anthropics/claude-code-action/blob/d721746d683d812e669ce117cebe55a85fbd9c3e/action.yml#L290) sets:

```yaml
MODE: ${{ inputs.mode }}
```

but there is no `mode:` entry in the `inputs:` block — it was removed when mode detection became automatic. The expression resolves to the empty string, and nothing in `src/` reads `process.env.MODE`. It should be deleted along with the stale map entries.

## Why this keeps happening

There is no mechanism tying the two lists together, so any input added to or removed from `action.yml` silently diverges from `collect-inputs.ts`. The drift above accumulated across the whole v1.0 migration without surfacing.

## Suggested fix

1. Regenerate `inputDefaults` from the current `action.yml`, dropping the 11 stale keys and adding the 19 missing ones with their declared defaults.
2. Remove the `MODE` env line from `action.yml`.
3. Add a drift-guard test that parses `action.yml` and asserts the key sets match exactly, so this cannot silently recur. [`test/action-metadata.test.ts`](https://github.com/anthropics/claude-code-action/blob/d721746d683d812e669ce117cebe55a85fbd9c3e/test/action-metadata.test.ts) is the natural home — it already reads and asserts against `action.yml`, and currently contains a single test.

A question for maintainers before I open a PR: should the guard assert exact set equality, or should there be an explicit opt-out list for inputs deliberately excluded from the presence signal (secrets such as `anthropic_api_key` and `github_token` are currently tracked, so I assume exact equality is intended — but I'd rather confirm than guess).

Happy to submit the fix and the guard test together.

## Environment

GautamSharma99:
Fixes #1664.

The input map in `src/entrypoints/collect-inputs.ts` had drifted across the v1.0 migration — tracking 11 inputs that no longer exist and omitting 19 that do. Since its output becomes `GITHUB_ACTION_INPUTS`, the usage signal could not observe `prompt`, the input that selects agent mode over tag mode.

## Changes

**Resynced the map** with all 39 declared inputs and their declared defaults, in `action.yml` declaration order so the two can be diffed by eye. Removed: `mode`, `model`, `anthropic_model`, `fallback_model`, `allowed_tools`, `disallowed_tools`, `custom_instructions`, `direct_prompt`, `override_prompt`, `claude_env`, `max_turns`. Added: `prompt`, `claude_args`, `track_progress`, `plugins`, `plugin_marketplaces`, `use_bedrock`, `use_vertex`, `use_foundry`, `allowed_non_write_users`, `include_fix_links`, `branch_name_template`, `bot_id`, `bot_name`, `display_report`, `show_full_output`, `include_comments_by_actor`, `exclude_comments_by_actor`, `path_to_claude_code_executable`, `path_to_bun_executable`.

Defaults were extracted from `action.yml` programmatically rather than transcribed by hand.

**Promoted the map to an exported `INPUT_DEFAULTS`** so it can be asserted against. `collectActionInputsPresence` is otherwise untouched — the only change inside the function is the identifier.

**Removed `MODE: ${{ inputs.mode }}`** from `action.yml`. That input was dropped when mode detection became automatic; the expression resolved to the empty string and nothing reads `process.env.MODE`.

**Added drift guards** to `test/action-metadata.test.ts`:

| Guard | Catches |
|---|---|
| Every declared input appears in `INPUT_DEFAULTS` | A new input added to `action.yml` and not tracked |
| No entry in `INPUT_DEFAULTS` is undeclared | An input removed from `action.yml` and left behind |
| Values match `action.yml`'s defaults exactly | A drifted default, which silently inverts the presence signal |
| Every `inputs.*` expression refers to a declared input | Exactly what the stale `MODE` reference violated |
| The parser finds >30 inputs and spot-checks three | The parser itself breaking and making the rest vacuously pass |

## On the question I raised in the issue

I asked whether the guard should assert exact set equality or allow an opt-out list for deliberately-excluded inputs. I went with **exact equality**, because the existing map already tracked `anthropic_api_key` and `github_token` — so secrets were evidently intended to be in scope, and the signal is presence-only (a boolean), never a value. Easy to relax if you'd rather exclude some.

I also assert the **default values**, not just the key set. That wasn't in the issue, but a default drifting in `action.yml` without the map following would flip that input's reported presence, which seemed worth closing at the same time.

## Verification

- `bun test` — 925 pass, 0 fail
- `bun run typecheck` — clean
- `bun run format:check` — clean

The guards were mutation-tested rather than assumed:

| Mutation | Result |
|---|---|
| Removed `claude_args` from the map | 2 failed |
| Re-added stale `direct_prompt` key | 2 failed |
| Drifted `label_trigger` default to `"triage"` | 1 failed |
| Reintroduced `MODE: ${{ inputs.mode }}` | 1 failed |

## Notes for review

- **Parsed with regexes, not a YAML library.** The repo has no YAML dependency, the pre-existing assertion in this file already matches against raw text, and `ci.yml` pins Bun 1.2.12 — which predates `Bun.YAML`. Adding a parser dependency for one test seemed the worse trade; happy to switch if you'd prefer it.
- **No behavioural change** to `collectActionInputsPresence` beyond the map contents it iterates.
- Unrelated, so deliberately left out of this PR: `bun.lock` still records `shell-quote@^1.8.3` while `package.json` was bumped to `^1.8.4` in 5da4c76. Running `bun install` resyncs it. Happy to send that separately if it's not already known.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -287,7 +287,6 @@ runs:
           run ${GITHUB_ACTION_PATH}/src/entrypoints/run.ts
       env:
         # Prepare inputs
-        MODE: ${{ inputs.mode }}
         PROMPT: ${{ inputs.prompt }}
         TRIGGER_PHRASE: ${{ inputs.trigger_phrase }}
         ASSIGNEE_TRIGGER: ${{ inputs.assignee_trigger }}
--- a/src/entrypoints/collect-inputs.ts
+++ b/src/entrypoints/collect-inputs.ts
@@ -1,38 +1,58 @@
-export function collectActionInputsPresence(): string {
-  const inputDefaults: Record<string, string> = {
-    trigger_phrase: "@claude",
-    assignee_trigger: "",
-    label_trigger: "claude",
-    base_branch: "",
-    branch_prefix: "claude/",
-    allowed_bots: "",
-    mode: "tag",
-    model: "",
-    anthropic_model: "",
-    fallback_model: "",
-    allowed_tools: "",
-    disallowed_tools: "",
-    custom_instructions: "",
-    direct_prompt: "",
-    override_prompt: "",
-    additional_permissions: "",
-    claude_env: "",
-    settings: "",
-    anthropic_api_key: "",
-    claude_code_oauth_token: "",
-    anthropic_federation_rule_id: "",
-    anthropic_organization_id: "",
-    anthropic_service_account_id: "",
-    anthropic_workspace_id: "",
-    anthropic_oidc_audience: "",
-    github_token: "",
-    max_turns: "",
-    use_sticky_comment: "false",
-    classify_inline_comments: "true",
-    use_commit_signing: "false",
-    ssh_signing_key: "",
-  };
+/**
+ * Declared defaults for every input in action.yml, in declaration order.
+ *
+ * Used to report which inputs a workflow actually set (see
+ * collectActionInputsPresence). An input is considered "set" when its value
+ * differs from the default declared here, so both the key set and the values
+ * must track action.yml — a default that drifts silently inverts the signal
+ * for that input.
+ *
+ * test/action-metadata.test.ts parses action.yml and asserts this map matches
+ * it exactly. Update both together.
+ */
+export const INPUT_DEFAULTS: Record<string, string> = {
+  trigger_phrase: "@claude",
+  assignee_trigger: "",
+  label_trigger: "claude",
+  base_branch: "",
+  branch_prefix: "claude/",
+  branch_name_template: "",
+  allowed_bots: "",
+  allowed_non_write_users: "",
+  include_comments_by_actor: "",
+  exclude_comments_by_actor: "",
+  prompt: "",
+  settings: "",
+  anthropic_api_key: "",
+  claude_code_oauth_token: "",
+  anthropic_federation_rule_id: "",
+  anthropic_organization_id: "",
+  anthropic_service_account_id: "",
+  anthropic_workspace_id: "",
+  anthropic_oidc_audience: "",
+  github_token: "",
+  use_bedrock: "false",
+  use_vertex: "false",
+  use_foundry: "false",
+  claude_args: "",
+  additional_permissions: "",
+  use_sticky_comment: "false",
+  classify_inline_comments: "true",
+  use_commit_signing: "false",
+  ssh_signing_key: "",
+  bot_id: "41898282",
+  bot_name: "claude[bot]",
+  track_progress: "false",
+  include_fix_links: "true",
+  path_to_claude_code_executable: "",
+  path_to_bun_executable: "",
+  display_report: "false",
+  show_full_output: "false",
+  plugins: "",
+  plugin_marketplaces: "",
+};
 
+export function collectActionInputsPresence(): string {
   const allInputsJson = process.env.ALL_INPUTS;
   if (!allInputsJson) {
     console.log("ALL_INPUTS environment variable not found");
@@ -49,7 +69,7 @@ export function collectActionInputsPresence(): string {
 
   const presentInputs: Record<string, boolean> = {};
 
-  for (const [name, defaultValue] of Object.entries(inputDefaults)) {
+  for (const [name, defaultValue] of Object.entries(INPUT_DEFAULTS)) {
     const actualValue = allInputs[name] || "";
 
     const isSet = actualValue !== defaultValue;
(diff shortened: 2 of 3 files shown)


Chirag6722:
I duplicated this by accident — opened #1680 against the same issue before spotting yours, and have closed it. Yours came first and is the better patch, so this is just a note in passing.

Asserting the **defaults** as well as the names is the part mine got wrong: I derived the tracked names from `collectActionInputsPresence()`'s output, which meant a default silently diverging from `action.yml` would still have passed. Exporting `INPUT_DEFAULTS` and comparing both is the stronger contract, and keeping the map in declaration order makes the two diffable by eye. The note on why not to add a YAML dependency is worth having written down.

One thing that might be additive, entirely optional: `collectActionInputsPresence()` has two early returns that appear to be uncovered — `ALL_INPUTS` absent, and `ALL_INPUTS` set to something unparseable. Both return `"{}"`, and the second logs through `console.error`. They are the paths that fire when the wiring in `action.yml` breaks, which is exactly when you would want the signal to fail loudly rather than quietly report nothing.

```ts
test("reports an empty object when ALL_INPUTS is absent", () => {
  delete process.env.ALL_INPUTS;
  expect(JSON.parse(collectActionInputsPresence())).toEqual({});
});

test("reports an empty object when ALL_INPUTS is not valid JSON", () => {
  process.env.ALL_INPUTS = "{not json";
  expect(JSON.parse(collectActionInputsPresence())).toEqual({});
});
```

Save/restore the original env in an `afterEach` if you take it, since other tests in the file may depend on it.

Take it or leave it — your PR stands on its own either way.

PR #1680, opened 2026-08-16
fix(inputs): sync collect-inputs map with action.yml and pin it (#1664)

### LINKED ISSUE #1664: collect-inputs.ts input map has drifted from action.yml: 11 removed inputs tracked, 19 current inputs missing (including prompt)

Chirag6722:
Closes #1664.

Reproduced the reported drift exactly before changing anything, using the commands from the issue:

```
declared=39  tracked=31
tracked but no longer declared (11): allowed_tools anthropic_model claude_env custom_instructions
  direct_prompt disallowed_tools fallback_model max_turns mode model override_prompt
declared but not tracked (19): allowed_non_write_users bot_id bot_name branch_name_template
  claude_args display_report exclude_comments_by_actor include_comments_by_actor include_fix_links
  path_to_bun_executable path_to_claude_code_executable plugin_marketplaces plugins prompt
  show_full_output track_progress use_bedrock use_foundry use_vertex
```

After:

```
declared=39  tracked=39   stale=0   missing=0
```

`bot_id` picks up its real default of `"41898282"` — its line in `action.yml` carries a trailing comment, which is worth knowing if anyone regenerates this by hand.

## The part that matters more than the resync

The map is hand-maintained and `action.yml` is the only source of truth, with nothing holding the two together. That is *why* it drifted, so a straight resync just resets the clock. `test/collect-inputs.test.ts` reads `action.yml` at test time and compares it against the input names the exported function actually reports, following the `readFileSync(new URL("../action.yml", import.meta.url))` idiom already used in `test/action-metadata.test.ts`.

It asserts against `collectActionInputsPresence()`'s output rather than the private `inputDefaults` constant, so the guard is about observable behaviour — the thing that becomes `GITHUB_ACTION_INPUTS` — not an implementation detail.

I verified the guard fails in **both** drift directions rather than only the one I introduced:

| Drift | Result |
|---|---|
| baseline | 6 pass, 0 fail |
| an input removed from the map (a new input never added) | **3 fail** |
| a stale input left in the map | **2 fail** |

A guard I had not watched redden would not be worth adding.

`prompt` gets its own named test, since its absence is what made the signal actively misleading rather than merely incomplete. I also covered the two existing early-return paths (`ALL_INPUTS` absent, and unparseable), which had no coverage.

## Verification

```
bun test test/collect-inputs.test.ts   6 pass, 0 fail
bun run typecheck                      clean
bun test (full)                        882 pass / 44 fail
```

**Those 44 failures are pre-existing.** I stashed the change and re-ran on clean `HEAD`: `876 pass / 44 fail` — identical failures, and my change adds 6 passes and no new failures. Happy to share the list if it is useful.

Similarly `bun run format:check` reports ~189 files, but clean `HEAD` reports **191** on this machine. That is a `core.autocrlf=true` artifact on Windows, not something this PR introduces. Both files I touched are prettier-clean (`npx prettier --check` passes on them specifically), and the diff is 20 insertions / 12 deletions with no line-ending churn — `git diff` and `git diff -w` are identical.

## Note

I regenerated the map from `action.yml` rather than reading it at runtime. Reading it at runtime would remove the duplication entirely, but the action executes from a bundled `dist`, so I did not want to assume `action.yml` is resolvable there. If it is, that would be the stronger fix and I am happy to redo it that way.


Chirag6722:
Closing this — #1672 covers the same issue, was opened about eight hours before it, and is the better patch. Maintainers should review that one, not this.

My mistake: I checked #1664 for assignees and comments, found none, and started. I did not check for an open PR already referencing the issue, which is the check that would have caught it in seconds.

For what it is worth, #1672 is stronger where it counts. It exports `INPUT_DEFAULTS` and asserts the **names and the default values** both match `action.yml`; mine derived names from the function's output and so could not check defaults at all. It also keeps the map in `action.yml` declaration order, which makes the two diffable by eye. Its regex-parsing rationale is written down properly, including why not to reach for a YAML dependency.

One small thing from this branch that may be additive there, offered as a suggestion rather than a request — I have left it on #1672 rather than repeating it here.

Apologies for the duplicate noise, @GautamSharma99.

PR #1697, opened 2026-08-19
feat: package repo as a Claude Code plugin

Flawed-it:
## Summary

Makes this repository installable as a Claude Code plugin, so the MCP servers the action already configures can be used from an interactive Claude Code session rather than only inside a workflow run.

No runtime code changes — this is additive configuration plus docs.

## What's added

- **`.claude-plugin/plugin.json`** — plugin manifest.
- **`.mcp.json`** — five stdio servers mirroring `prepareMcpConfig`:
  - the four `bun` servers in `src/mcp/` (`github_comment`, `github_inline_comment`, `github_file_ops`, `github_ci`), launched with the same flags `bunServerArgs` uses (`--no-env-file`, `--config` pinned to `bunfig.toml`) so runtime config resolves from the plugin directory
  - GitHub's upstream `github` server, pinned to the same `ghcr.io/github/github-mcp-server:sha-23fa0dd` digest `install-mcp-server.ts` pins
- **`docs/claude-code-plugin.md`** — tool inventory per server, full env-var reference, and the local-session caveats.
- **`.gitignore`** — ignore `.claude/worktrees/`.

## Notes

- Env vars use `${VAR:-default}` where the code has a default, so an unset optional var can't leak a literal `${VAR}` into a server process.
- The four `bun` servers were written for the Actions runtime and depend on injected repo/PR context (`CLAUDE_COMMENT_ID`, `PR_NUMBER`, `BRANCH_NAME`, …). The docs are explicit about which ones need what and where the limits are locally, rather than implying they all work standalone.
- The digest is now pinned in two places (`.mcp.json` and `prepareMcpConfig`); the docs call out that they need bumping together.

## Testing

- `prettier --check` — clean on all changed files.
- `bun test` — 887 pass / 44 fail, **identical to the same run on `main` without this commit** (verified by testing `upstream/main` detached). The 44 failures are pre-existing Windows-environment failures, unrelated to this change. On Linux CI this should be green.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,0 +1,13 @@
+{
+  "name": "claude-code-action",
+  "version": "1.0.0",
+  "description": "Bundles the GitHub MCP servers that power the Claude Code GitHub Action: PR/issue comment updates, inline review comments, commit-signed file operations, and CI status inspection.",
+  "author": {
+    "name": "Anthropic",
+    "url": "https://github.com/anthropics/claude-code-action"
+  },
+  "homepage": "https://github.com/anthropics/claude-code-action",
+  "repository": "https://github.com/anthropics/claude-code-action",
+  "license": "MIT",
+  "keywords": ["github", "mcp", "pull-request", "code-review", "ci"]
+}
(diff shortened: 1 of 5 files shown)


Flawed-it (commented):
Solid

PR #1702, opened 2026-08-20
fix: use paths in delete_files prompt example

### LINKED ISSUE #1665: Tag-mode prompt documents delete_files with a "files" parameter, but the tool schema declares "paths"
## Summary

The tag-mode prompt tells Claude to call `mcp__github_file_ops__delete_files` with a `files` parameter, but the tool's schema declares that parameter as `paths`. Every first attempt to delete a file under `use_commit_signing: true` therefore fails schema validation.

## The mismatch

Prompt instruction — [`src/create-prompt/index.ts#L827`](https://github.com/anthropics/claude-code-action/blob/d721746d683d812e669ce117cebe55a85fbd9c3e/src/create-prompt/index.ts#L827):

```
  - mcp__github_file_ops__delete_files: {"files": ["path/to/old.js"], "message": "chore: remove deprecated file"}
```

Tool schema — [`src/mcp/github-file-ops-server.ts#L414-L425`](https://github.com/anthropics/claude-code-action/blob/d721746d683d812e669ce117cebe55a85fbd9c3e/src/mcp/github-file-ops-server.ts#L414-L425):

```ts
server.tool(
  "delete_files",
  "Delete one or more files from a repository in a single commit",
  {
    paths: z
      .array(z.string())
      .describe(
        'Array of file paths to delete relative to repository root (e.g. ["src/old-file.js", "docs/deprecated.md"])',
      ),
    message: z.string().describe("Commit message"),
  },
  async ({ paths, message }) => {
```

The handler destructures `paths`. A call shaped as `{"files": [...]}` fails zod validation before the handler body executes.

## How the inconsistency arose

The sibling tool in the same file genuinely does use `files`. [`commit_files`](https://github.com/anthropics/claude-code-action/blob/d721746d683d812e669ce117cebe55a85fbd9c3e/src/mcp/github-file-ops-server.ts#L197-L208) declares:

```ts
server.tool(
  "commit_files",
  ...
  {
    files: z.array(z.string()).describe(...),
    message: z.string().describe("Commit message"),
  },
  async ({ files, message }) => {
```

So the two tools take differently named array parameters, and the prompt example for `delete_files` was written against `commit_files`' shape. The line immediately above it in the prompt is correct:

```
  - mcp__github_file_ops__commit_files: {"files": ["path/to/file1.js", "path/to/file2.py"], "message": "feat: add new feature"}
```

## Impact

Scoped to `use_commit_signing: true`, which is the path where `delete_files` is allowlisted ([`src/modes/tag/index.ts#L155-L160`](https://github.com/anthropics/claude-code-action/blob/d721746d683d812e669ce117cebe55a85fbd9c3e/src/modes/tag/index.ts#L155-L160)). In that configuration:

- The model follows the instruction it was given, the call is rejected, and a turn is spent on a validation error.
- Recovery depends on the model inferring the correct parameter name from the error. Not fatal, but it is avoidable latency and token spend on a documented, always-present instruction.

This lives in the block that `CLAUDE.md` identifies as the most important part of the action:

> **Prompt construction**: ... The prompt includes issue/PR body, comments, diff, and CI status. This is the most important part of the action — it's what Claude sees.

## Suggested fix

Change `"files"` to `"paths"` in the `delete_files` example at `src/create-prompt/index.ts:827`.

The one-word fix is obvious, but the class of bug is worth guarding: the prompt hard-codes JSON examples for MCP tools whose schemas live in a different file, with nothing asserting they agree. I'd suggest pairing the fix with a test that imports the tool schemas and asserts every parameter name appearing in a prompt tool-usage example exists in the corresponding schema. Happy to include that in the PR if it's wanted — or to keep the change to the single word if you'd rather not take on the test.

## Environment

WeAreResilience:
## Summary

The tag-mode prompt told the model to call `mcp__github_file_ops__delete_files` with `files`. The tool schema and handler expect `paths`, so the first signed-commit delete fails Zod validation.

This changes the prompt example to `paths` and adds a test so the example cannot drift back to `files`.

Fixes #1665

## Test plan

- [x] `bun test test/create-prompt.test.ts` (49 pass)
- [x] `bun run typecheck`

From [RESILIENCE Agentic Solutions](https://we-are-resilience.com).

### DIFF
--- a/src/create-prompt/index.ts
+++ b/src/create-prompt/index.ts
@@ -824,7 +824,7 @@ ${
     ? `- Use mcp__github_file_ops__commit_files for making commits (works for both new and existing files, single or multiple). Use mcp__github_file_ops__delete_files for deleting files (supports deleting single or multiple files atomically), or mcp__github__delete_file for deleting a single file. Edit files locally, and the tool will read the content from the same path on disk.
   Tool usage examples:
   - mcp__github_file_ops__commit_files: {"files": ["path/to/file1.js", "path/to/file2.py"], "message": "feat: add new feature"}
-  - mcp__github_file_ops__delete_files: {"files": ["path/to/old.js"], "message": "chore: remove deprecated file"}`
+  - mcp__github_file_ops__delete_files: {"paths": ["path/to/old.js"], "message": "chore: remove deprecated file"}`
     : `- Use git commands via the Bash tool for version control (remember that you have access to these git commands):
   - Stage files: Bash(git add <files>)
   - Commit changes: Bash(git commit -m "<message>")
--- a/src/mcp/github-file-ops-schemas.ts
+++ b/src/mcp/github-file-ops-schemas.ts
@@ -1,0 +1,24 @@
+import { z } from "zod";
+
+/** Raw shape passed to `server.tool` for commit_files. */
+export const commitFilesInputSchema = {
+  files: z
+    .array(z.string())
+    .describe(
+      'Array of file paths relative to repository root (e.g. ["src/main.js", "README.md"]). All files must exist locally.',
+    ),
+  message: z.string().describe("Commit message"),
+};
+
+/** Raw shape passed to `server.tool` for delete_files. */
+export const deleteFilesInputSchema = {
+  paths: z
+    .array(z.string())
+    .describe(
+      'Array of file paths to delete relative to repository root (e.g. ["src/old-file.js", "docs/deprecated.md"])',
+    ),
+  message: z.string().describe("Commit message"),
+};
+
+export const commitFilesPayloadSchema = z.object(commitFilesInputSchema);
+export const deleteFilesPayloadSchema = z.object(deleteFilesInputSchema);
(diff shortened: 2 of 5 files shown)


WeAreResilience:
Verification against the live `delete_files` Zod schema (the same shape `server.tool` registers):

| Check | Result |
| --- | --- |
| Old prompt payload `{ files, message }` | rejected (`paths` required) |
| New prompt payload `{ paths, message }` | accepted |
| Generated tag-mode example parses the live schema | pass |
| `paths` as a string, not an array | rejected |
| `paths` present, `message` omitted | rejected |
| Extra `files` key beside valid `paths` | accepted (unknown keys stripped) |
| `commit_files` example still uses `files` | unchanged |
| Generated `delete_files` keys are exactly `paths` + `message` | pass |

These live in `test/delete-files-prompt-schema.test.ts` (8 tests). Happy to trim the extra coverage if you would rather keep the PR to the one-word prompt change.

ashwin-ant (approved):

PR #1703, opened 2026-08-20
fix: restore AGENTS.md from the PR base branch

### LINKED ISSUE #1696: AGENTS.md is not in SENSITIVE_PATHS — the documented @AGENTS.md import loads the PR head's instructions despite the CLAUDE.md restore
## Summary

`SENSITIVE_PATHS` (src/github/operations/restore-config.ts, ~line 26 at v1.0.195 / `d40ddef`) restores `CLAUDE.md`, `CLAUDE.local.md`, `.claude`, `.mcp.json`, `.claude.json`, `.gitmodules`, `.ripgreprc`, and `.husky` from the base branch on PR contexts, described in-file as "Paths that are both PR-controllable and read from cwd at CLI startup" — so a PR cannot inject instructions into the run.

`AGENTS.md` is not on that list. But the Claude Code memory docs explicitly recommend keeping shared instructions in `AGENTS.md` and importing them from `CLAUDE.md` via `@AGENTS.md` (https://code.claude.com/docs/en/memory, "AGENTS.md" section). In a repo using that documented pattern, the restored base-branch `CLAUDE.md` still contains the `@AGENTS.md` import — and the import resolves against the working tree, which is the **PR head** checkout (src/github/operations/branch.ts checks out `pull/N/head`, including fork PRs). The import therefore loads the PR author's `AGENTS.md` into the run's instructions, reopening the injection path the CLAUDE.md restore was built to close.

## Impact

Bounded by the trigger gate (write access required by default, `allowed_non_write_users` empty), so exploitation requires a maintainer commenting `@claude` on a PR whose head modifies or adds `AGENTS.md`. That is precisely the scenario `SENSITIVE_PATHS` exists to defend — a maintainer invoking the action on an untrusted contribution — and the file it misses is the one the product's own docs tell users to put their instructions in. Nested `AGENTS.md` files in subdirectories (also part of the agents.md convention) have the same property when imported.

## Reproduction sketch

1. Repo with `CLAUDE.md` containing `@AGENTS.md` (the documented pattern) and both files on the default branch.
2. External PR modifies `AGENTS.md` to include adversarial instructions.
3. Maintainer comments `@claude` on the PR.
4. The action restores `CLAUDE.md` from base (correct), but the `@AGENTS.md` import inside it resolves to the PR head's file — the modified instructions load.

## Suggested fix

Add `AGENTS.md` (and arguably `AGENTS.local.md` / nested `**/AGENTS.md`, matching whatever the import + native-discovery surface grows to) to `SENSITIVE_PATHS`. More generally: any file reachable from a restored `CLAUDE.md` via `@` imports inherits CLAUDE.md's trust requirements — a recursive restore of the import closure would close the class rather than the instance, but the single filename covers the documented pattern today.

Found while auditing our own repo's adoption of the `@AGENTS.md` import pattern (action pinned at v1 → v1.0.195, installing Claude Code 2.1.235; confirmed 2.1.235 does not natively discover AGENTS.md, so the import is the sole loader — which is exactly why it escapes the restore).

WeAreResilience:
## Summary

`restoreConfigFromBase` already replaces `CLAUDE.md` from the PR base. The documented `@AGENTS.md` import still resolved against the working tree, so a PR that rewrote `AGENTS.md` could inject instructions after that restore.

This adds `AGENTS.md` and `AGENTS.local.md` to `SENSITIVE_PATHS`. The PR-head copies are snapshotted into `.claude-pr/` for review agents, then replaced from base (or deleted if they only exist on the head).

Fixes #1696

## Test plan

- [x] `bun test test/restore-config.test.ts` (20 pass, run three times)
- [x] `bun run typecheck`
- [x] Existing symlink-to-`AGENTS.md` restore tests still pass

From [RESILIENCE Agentic Solutions](https://we-are-resilience.com).

Made with [Cursor](https://cursor.com)


tarunag10 (commented):
The root-file fix and tests cover the documented `@AGENTS.md` pattern cleanly. One residual trust-boundary case is worth making explicit before merge: a restored base `CLAUDE.md` can import a nested path such as `@docs/AGENTS.md`, and that path still resolves from the PR-head checkout because `SENSITIVE_PATHS` only snapshots the two root filenames. If the intended scope is the documented root convention, I suggest adding a comment/test that records this limitation; if the intended invariant is “restored instructions cannot import PR-controlled content,” the robust fix is to resolve and restore the base `@`-import closure rather than enumerate filenames. I would avoid a broad `**/AGENTS.md` glob unless restore-config can preserve directory semantics and snapshot every matched path safely.

WeAreResilience:
Agreed — the right invariant is that restored instructions cannot import PR-controlled content.

This follow-up walks the **base** `@`-import closure after `SENSITIVE_PATHS` are restored (documented syntax, relative to the importing file, max four hops, code spans/fences skipped). Extra repo-relative files are snapshotted into `.claude-pr/`, then checked out from base or deleted if they are absent there. Absolute/`~`/`..` escapes are ignored, and directory targets are not restored — no `**/AGENTS.md` glob.

The `@docs/AGENTS.md` case you named is now a test, along with PR-only nested deletes, backtick literals, path escapes, a two-hop cycle, and the fifth-hop depth bound.

PR #1722, opened 2026-08-24
fix: clean checkout includeIf credentials

### LINKED ISSUE #1721: Checkout credential cleanup is still a no-op: checkout persists via includeIf.gitdir, not include.path (follow-up to #1510)
## Summary

The fix for #1510 walks `include.path` entries of the local git config, but `actions/checkout`
does not use `include.path` in the normal (non-global) code path — it writes `includeIf.gitdir:`
entries. So `replaceCheckoutCredentials` still removes nothing, and the checkout credential
(the workflow `GITHUB_TOKEN`) stays active for every git operation in the job, overriding the
app token the action puts into the remote URL.

Observed with `anthropics/claude-code-action@v1.0.199` and `actions/checkout@v7.0.1`.

## Evidence

Job log of a real tag-mode run (`@claude` on an issue comment, `use_commit_signing: false`):

```
Configuring git user...
Removing existing git authentication headers...
No existing authentication headers to remove      <-- cleanup found nothing
Updating remote URL with authentication...
✓ Updated remote URL with authentication token
```

The agent then committed and pushed successfully — but under `GITHUB_TOKEN`, not under the app
token, which we could confirm from the outside: the push produced **no** `pull_request:
synchronize` event, so neither CI nor our review workflow ran on the commit the agent added to
its own PR. GitHub suppresses events for actions performed with `GITHUB_TOKEN`, and that is only
consistent with the header winning over the credentials embedded in the remote URL.

For contrast, the same repository on `actions/checkout@v5.1.0` (which still wrote the header into
`.git/config`) logged `✓ Removed existing authentication headers`, and pushes from the agent did
raise `synchronize` — CI and review ran on bot-pushed commits. The regression appeared for us
exactly when Dependabot bumped checkout from v5.1.0 to v7.0.1.

## Why the current cleanup misses it

`replaceCheckoutCredentials` (`src/github/operations/git-config.ts`, v1.0.199) unsets
`http.<server>/.extraheader` from the local config and then from every file listed in
`git config --local --get-all include.path`.

`actions/checkout` v7.0.1 (`src/git-auth-helper.ts`, `configureToken`) only writes `include.path`
in the `globalConfig` branch. In the branch that runs for a normal checkout it writes, into the
repository-local config:

```ts
const hostIncludeKey = `includeIf.gitdir:${gitDir}.path`
await this.git.config(hostIncludeKey, credentialsConfigPath)

const hostWorktreeIncludeKey = `includeIf.gitdir:${gitDir}/worktrees/*.path`
await this.git.config(hostWorktreeIncludeKey, credentialsConfigPath)
```

plus equivalent container-path variants. None of these are `include.path`, so the loop iterates
over an empty list and the header survives.

## Impact

Two shapes, depending on what the workflow grants `GITHUB_TOKEN`:

1. **`contents: write` granted** — the push succeeds under `GITHUB_TOKEN` and silently raises no
   events. Downstream workflows never run on commits the agent adds to an existing PR, while the
   PR still shows the checks from the previous SHA, so it looks verified. This is the failure we
   hit; it is silent, which makes it worse than a hard error.
2. **`contents: write` not granted** — the push fails with `403 Write access to repository not
   granted`, which appears to be what #907 describes.

## Suggested fix

Collect every include target rather than just `include.path`, e.g.

```
git config --local --get-regexp '^include(If\..+)?\.path$'
```

and unset `http.<server>/.extraheader` from each referenced file (the same
`git config --file <path> --unset-all` call already used). Optionally also drop the
`includeIf.gitdir:*` keys themselves once their credential file no longer holds the header.

A regression test would need to cover the non-global branch specifically — the current one only
exercises `include.path`, which checkout uses solely for global config.

## Workaround for others hitting this

Give `actions/checkout` a token that is not `GITHUB_TOKEN` (e.g. from
`actions/create-github-app-token`), so whichever credential wins is still one whose actions raise
events. `persist-credentials: false` also avoids the conflict, but it has separate reported
issues (#1236, #1711).

hamedrabah:
## Summary

- discover active checkout authorization headers from their Git config origins
- remove headers contributed through both `include.path` and checkout v7 `includeIf.gitdir` files
- preserve the include entries while replacing the effective Git credential
- add regression coverage for ordinary and conditional include layouts, including config paths with spaces

## Why

`actions/checkout@v7` persists its credential through `includeIf.gitdir:*`, so scanning only `include.path` leaves the workflow `GITHUB_TOKEN` active. Git can then prefer that header over the action token, suppressing downstream workflow events for agent pushes.

## Verification

- `bun test test/git-config.test.ts` (6 passed)
- `bun build src/github/operations/git-config.ts --target bun`
- `git diff --check`

Closes #1721.

### DIFF
--- a/src/github/operations/git-config.ts
+++ b/src/github/operations/git-config.ts
@@ -60,12 +60,12 @@ export async function configureGitAuth(
  * actions/checkout < v6 stored the header directly in the repo-local config,
  * where `git config --unset-all` removes it. Since v6.0.0 (backported to
  * v5.0.1 and v4.3.1) the header is written to a separate file under
- * RUNNER_TEMP that the repo config pulls in via `include.path`; `--unset-all`
- * on the local config cannot touch an include-provided value, so the removal
- * was a silent no-op and the checkout credential (typically the workflow
- * GITHUB_TOKEN) stayed usable by git for the rest of the job. Clear the
- * header from the local config AND from every included file so it can no
- * longer authenticate while Claude runs.
+ * RUNNER_TEMP that the repo config pulls in via `include.path` or
+ * `includeIf.gitdir:*.path`; `--unset-all` on the local config cannot touch an
+ * include-provided value, so the removal was a silent no-op and the checkout
+ * credential (typically the workflow GITHUB_TOKEN) stayed usable by git for
+ * the rest of the job. Clear the header from every config file that contributes
+ * an active value so it can no longer authenticate while Claude runs.
  */
 export async function replaceCheckoutCredentials(
   githubToken: string,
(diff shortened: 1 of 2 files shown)


hamedrabah:
@ashwin-ant when you have a chance, could you take a look? This fixes the checkout v7 `includeIf.gitdir` credential-cleanup gap from #1721 and adds regression coverage for both conditional and ordinary include layouts. The focused suite is green (6/6), and the PR is conflict-free. Thanks!

hamedrabah:
@km-anthropic, could you review this credential-cleanup fix for #1721 when convenient? It handles the includeIf.gitdir entries that the current cleanup misses and includes regression coverage.

hamedrabah:
Could a maintainer approve the pending fork workflows for this PR? GitHub reports the branch as mergeable; the CI and integration workflows are waiting for approval.

PR #1733, opened 2026-08-26
fix: restore package-manager config (.pnpmfile.cjs/.npmrc/.yarnrc.yml) from base to close install-time RCE

arpe1618:
## Summary

`restoreConfigFromBase` reverts PR-controlled config that gets auto-loaded or
auto-executed from `cwd` back to the reviewed base version, so a fork PR can't
run code before/while the agent works. The list already covers `.husky`
(git hooks that run on the routine `git commit` the agent performs) and
`.gitmodules` (read on fetch), but it omits the package-manager equivalents.

This adds `.pnpmfile.cjs`, `.npmrc`, and `.yarnrc.yml` — the same three files
Claude Code itself already recognizes as sensitive config.

## Why these are the same class as `.husky`

Installing dependencies is a routine, expected agent step in JS repos ("set up
the project", "run the tests", "add a dep"). A fork PR that plants one of these
gets code execution the moment that install runs:

- **`.pnpmfile.cjs`** — pnpm loads and runs its top-level JS on install/resolve.
  It's *config*, not a lifecycle script, so it executes **even under
  `--ignore-scripts`**. "ignore-scripts makes install safe" does not hold for
  this file. Reproduced on pnpm 11.21.0 / Node 24.19.0 — in an empty directory
  containing only:

  ```jsonc
  // package.json
  {"name":"t","version":"1.0.0"}
  ```
  ```js
  // .pnpmfile.cjs
  require('fs').writeFileSync('pnpmfile-ran','1'); module.exports = {};
  ```
  ```console
  $ pnpm install --ignore-scripts --offline
  Already up to date

  Done in 638ms using pnpm v11.21.0
  $ ls pnpmfile-ran
  pnpmfile-ran
  ```

  The marker file is created, so the pnpmfile ran. Renaming `.pnpmfile.cjs` away
  and re-running leaves no marker, confirming that file is the source. The
  fixture has no dependencies at all, so this is the pnpmfile itself executing,
  not a package lifecycle script.
- **`.yarnrc.yml`** — `yarnPath` / `plugins` point to JS that Yarn loads and
  executes on any command, install included.
- **`.npmrc`** — can redirect the registry (supply-chain) and re-enable install
  scripts.

This is the same pattern that justified `.husky`: a PR-planted file that runs
arbitrary code when the agent performs an ordinary operation (`git commit` →
`pnpm install`).

## Scope / honesty

This is **defense-in-depth for the fork-PR vector**, not a claim of a new
zero-click auto-trigger — nothing here runs at CLI startup on its own. It closes
the gap where a maintainer relies on the model to notice a disguised
`.pnpmfile.cjs` before running `install`; model judgment is probabilistic and
model-dependent, so a deterministic revert-to-base is the right backstop. The
tradeoff is identical to the existing entries: the PR-authored version is still
snapshotted to `.claude-pr/` for review agents to inspect (never executed), and
only PR-head *changes* are reverted — a legitimate base `.npmrc`/`.yarnrc.yml`
is preserved.

## Changes

- Add `.pnpmfile.cjs`, `.npmrc`, `.yarnrc.yml` to `SENSITIVE_PATHS`.
- Expand the block comment so the list's rationale stays accurate (covers files
  executed by tooling the agent routinely runs, not only CLI-startup reads).
- Tests: a PR-added `.pnpmfile.cjs` is removed from the working tree but kept in
  `.claude-pr/` for review; a PR-modified `.npmrc` is restored to base.

## Testing

- Added two unit tests in `test/restore-config.test.ts` covering the new paths (a PR-added `.pnpmfile.cjs` is removed and snapshotted to `.claude-pr/`; a PR-modified `.npmrc` is restored to base).
- `bun run typecheck` passes locally.
- Note: the full `bun test` suite has pre-existing symlink/autocrlf failures on Windows unrelated to this change (they reproduce on `main` too); the new tests are written in the same style as the existing ones and rely on CI (Linux) as the real signal.

Happy to trim to just `.pnpmfile.cjs` (the clearest-cut case) if you'd prefer
the most minimal change.

### DIFF
--- a/src/github/operations/restore-config.ts
+++ b/src/github/operations/restore-config.ts
@@ -31,11 +31,11 @@ import { fetchDepthArgs } from "./fetch-depth";
 //                       install scripts.
 //
 // Deliberately excluded from the CLI's broader auto-edit blocklist:
-// .git/ — not tracked by git; PR commits cannot place files there.
-// Restoring it would also undo the PR checkout entirely.
-// .gitconfig — git reads ~/.gitconfig and .git/config, never cwd/.gitconfig.
-// .bashrc etc. — shells source these from $HOME; checkout cannot reach $HOME.
-// .vscode/.idea— IDE config; nothing in the CLI's startup path reads them.
+//   .git/        — not tracked by git; PR commits cannot place files there.
+//                  Restoring it would also undo the PR checkout entirely.
+//   .gitconfig   — git reads ~/.gitconfig and .git/config, never cwd/.gitconfig.
+//   .bashrc etc. — shells source these from $HOME; checkout cannot reach $HOME.
+//   .vscode/.idea— IDE config; nothing in the CLI's startup path reads them.
 export const SENSITIVE_PATHS = [
   ".claude",
   ".mcp.json",
--- a/src/github/operations/restore-config.ts
+++ b/src/github/operations/restore-config.ts
@@ -27,8 +27,13 @@ import { fetchDepthArgs } from "./fetch-depth";
 //                       `--ignore-scripts`.
 //       .yarnrc.yml   — `yarnPath` / `plugins` point to JS that Yarn loads and
 //                       executes on any command, including install.
+//       .yarnrc       — Yarn Classic's `yarn-path` points to JS that Yarn 1
+//                       runs as its binary on any command, including install —
+//                       the pre-Berry equivalent of `.yarnrc.yml`.
 //       .npmrc        — can redirect the registry (supply-chain) and re-enable
 //                       install scripts.
+//       bunfig.toml   — `preload` runs JS on `bun run`/`bun test`, and
+//                       `[install].registry` can redirect the registry.
 //
 // Deliberately excluded from the CLI's broader auto-edit blocklist:
 //   .git/        — not tracked by git; PR commits cannot place files there.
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
Could `.yarnrc` be restored as well as `.yarnrc.yml`? Yarn Classic supports `yarn-path` in `.yarnrc`, which can point at PR-controlled JavaScript that Yarn executes when the agent runs a Yarn command, so a fork PR can retain essentially the same install-time code-execution path this change is closing for modern Yarn.

arpe1618:
@sylvesterkaczmarek good catch — `.yarnrc` (Yarn Classic) is the pre-Berry sibling of `.yarnrc.yml`: `yarn-path` there points to JS that Yarn 1 runs as its binary on any yarn command, so a Classic-repo fork keeps exactly the install-time exec path this closes for Berry.

Added it in 6a32e57. Also pulled in `bunfig.toml` while there — same class (`preload` runs JS on `bun run`/`bun test`, which the action itself calls; `[install]` can repoint the registry). Both come with tests mirroring the `.pnpmfile.cjs`/`.npmrc` ones.

sylvesterkaczmarek (commented):
Re-reviewed the Yarn Classic gap I raised. `.yarnrc` is now included alongside `.yarnrc.yml`, with regression coverage showing a PR-added `yarn-path` config is removed from the agent worktree and preserved in the review snapshot. My previous blocker is resolved.

PR #1736, opened 2026-08-26
fix: preserve repeated plugin-dir arguments

### LINKED ISSUE #1728: claude_args: repeated --plugin-dir collapses to last-one-wins (not in ACCUMULATING_FLAGS), silently dropping plugins
TL;DR: `base-action/src/parse-sdk-options.ts` parses `claude_args` into a record where only `ACCUMULATING_FLAGS` may repeat. `plugin-dir` isn't in the set, so `--plugin-dir A --plugin-dir B` reaches the CLI as `B` alone, silently dropping `A`, even though the CLI accepts the flag repeated. One-line fix: add `plugin-dir` to `ACCUMULATING_FLAGS`.

Repro: pass two `--plugin-dir` flags via `claude_args`; only the last plugin installs, no warning. Confirmed by transcribing the parser from source; present at v1.0.183 and v1.0.205.

Impact: any workflow loading >1 local plugin via `claude_args` loses all but the last, silently. In our org-wide PR-review pipeline this meant review agents ran for weeks without the org rulesets living in sibling plugins — invisible, because a `<plugin>:<skill>` reference into an uninstalled plugin fails without error. Two engineers independently root-caused it to the same line.

Workaround we shipped: the action's native `plugins/plugin_marketplaces` inputs (bypass `claude_args` parsing). They work well, but the silent collapse remains a trap for anyone reaching for the documented CLI flag.

aaravxsingh15:
## Summary

- add `plugin-dir` to `ACCUMULATING_FLAGS`
- preserve repeated `--plugin-dir` values instead of keeping only the final value
- add a focused regression test covering multiple plugin directories

Fixes #1728.

## Testing

- Focused parser suite: 42 passed, 0 failed
- TypeScript typecheck passed
- Prettier check passed
- Full Windows run: 916 passed; remaining failures occurred outside the focused parser suite

### DIFF
--- a/base-action/src/parse-sdk-options.ts
+++ b/base-action/src/parse-sdk-options.ts
@@ -20,6 +20,7 @@ const ACCUMULATING_FLAGS = new Set([
   "disallowed-tools",
   "mcp-config",
   "add-dir",
+  "plugin-dir",
 ]);
 
 // Delimiter used to join accumulated flag values
--- a/base-action/test/parse-sdk-options.test.ts
+++ b/base-action/test/parse-sdk-options.test.ts
@@ -280,7 +280,19 @@ describe("parseSdkOptions", () => {
       expect(result.sdkOptions.disallowedTools).toEqual(["Bash", "Write"]);
     });
   });
+  describe("plugin-dir accumulation", () => {
+    test("should preserve repeated --plugin-dir flags", () => {
+      const options: ClaudeOptions = {
+        claudeArgs: "--plugin-dir ./plugin-a --plugin-dir ./plugin-b",
+      };
 
+      const result = parseSdkOptions(options);
+
+      expect(result.sdkOptions.extraArgs?.["plugin-dir"]).toBe(
+        "./plugin-a\x00./plugin-b",
+      );
+    });
+  });
   describe("mcp-config merging", () => {
     test("should pass through single mcp-config in extraArgs", () => {
       const options: ClaudeOptions = {


farmer-data:
Heads up on a case this may not cover. I applied just the `ACCUMULATING_FLAGS` change locally against `main` (`a874e9e`, agent-sdk 0.3.251) and probed `parseSdkOptions`:

```
before:  extraArgs['plugin-dir'] = "./plugin-b"
after:   extraArgs['plugin-dir'] = "./plugin-a./plugin-b"
```

Accumulation joins with `ACCUMULATE_DELIMITER` (`\x00`), but unlike the other members of that set, nothing unpacks `plugin-dir` downstream — `add-dir` splits into `additionalDirectories` (L210–216), `allowedTools`/`disallowed-tools` split at L228–254, `mcp-config` splits and re-merges at L271–274. So the joined value stays in `extraArgs`, and the SDK emits one `--flag value` pair per entry with `String(v)` and no splitting:

```js
for (let [k, v] of Object.entries(extraArgs))
  if (v === null) argv.push(`--${k}`); else oC(argv, k, v);
```

Node then refuses the spawn:

```
ERR_INVALID_ARG_VALUE - The argument 'args[1]' must be a string
without null bytes. Received './plugin-a\x00./plugin-b'
```

The test asserts `toBe("./plugin-a\x00./plugin-b")`, which is why CI is green — it pins the joined intermediate rather than the delivered args.

The SDK has a first-class `plugins?: SdkPluginConfig[]` option (`sdk.d.ts:1860`) and already emits one `--plugin-dir` per entry from it, so unpacking there mirrors the `add-dir` handling. `--plugin-dir-no-mcp` has the same last-one-wins bug and maps to `skipMcpDiscovery`.

Happy to share a patch if useful, or leave it with you.

Issue #1543, opened 2026-07-23
Automatic PR review flags/misscopes to the working-tree reversal left by restoreConfigFromBase


jasonevines:
## Summary

When a PR modifies any path in `restoreConfigFromBase`'s `SENSITIVE_PATHS` (`.claude`, `.mcp.json`, `.claude.json`, `.gitmodules`, `.ripgreprc`, `CLAUDE.md`, `CLAUDE.local.md`, `.husky`), the security restore reverts those paths to the base-branch version in the working tree, leaving **uncommitted reversals** of the PR's own edits — the documented "known limitation" in [`restore-config.ts`](https://github.com/anthropics/claude-code-action/blob/main/src/github/operations/restore-config.ts). An automatic PR-review workflow that invokes the CLI-native `/code-review` — whose default scope includes "any uncommitted changes in the working tree" ([docs](https://code.claude.com/docs/en/code-review)) — then reviews those reversals **as if they were part of the PR**, producing false findings; on large base↔head divergence it misscopes the entire review to them.

## Distinct from existing issues

- [#1134](https://github.com/anthropics/claude-code-action/issues/1134) / [#1172](https://github.com/anthropics/claude-code-action/pull/1172): the PR's `.claude/` changes being *invisible* to review (fixed via the `.claude-pr/` snapshot). This is the opposite-facing symptom — the review *does* see the reversal and treats it as a finding.
- [#1275](https://github.com/anthropics/claude-code-action/issues/1275) / [#1277](https://github.com/anthropics/claude-code-action/pull/1277): the `.claude-pr/` snapshot being *committed* by a later `git add -A`. Also distinct — this is about the review, not a commit.

## Reproduction

1. A PR that edits, e.g., `.claude/settings.json` or `CLAUDE.md`.
2. An automatic review workflow (`pull_request: synchronize`) that runs the action and invokes the bundled `/code-review` at a chosen effort level, with no explicit diff target.
3. The review reports the base-vs-PR difference on the edited sensitive path as an uncommitted working-tree change and comments on it as a PR finding.

Observed with `anthropics/claude-code-action@v1`; the job log shows `Restoring .claude, … from origin/<base> (PR head is untrusted)` shortly before the review runs.

## Impact

[...]

PR #1755, opened 2026-08-29
fix(git): run git against the checkout, not the ambient working directory

### LINKED ISSUE #1591: restoreConfigFromBase / configureGitAuth fail with "not a git repository" right after actions/checkout succeeds (not the #1560 dubious-ownership error)
## Summary

`restoreConfigFromBase` (and the earlier `configureGitAuth`, same failure signature) fails with a plain "no git repo here" error, not the previously-reported "dubious ownership" error (#1560, fixed 2026-07-28). This happens even though `actions/checkout@v6` completed successfully immediately before, in the same job.

## Observed logs

First (non-fatal, caught by try/catch in `prepareAgentMode`):

```
Configuring git user...
Setting git user as claude[bot]...
fatal: not in a git directory
Failed to configure git authentication: ...
ShellError: Failed with exit code 128
    stderr: "fatal: not in a git directory\n"
    at configureGitAuth (.../src/github/operations/git-config.ts:41:9)
    at prepareAgentMode (.../src/modes/agent/index.ts:60:13)
```

Second (fatal, kills the job — no `structured_output` is ever produced):

```
Restoring .claude, .mcp.json, .claude.json, .gitmodules, .ripgreprc, CLAUDE.md, CLAUDE.local.md, .husky from origin/main (PR head is untrusted)
fatal: not a git repository (or any of the parent directories): .git
##[error]Action failed with error: Command failed: git fetch origin main --depth=1 --no-recurse-submodules
```

This second one is the `execFileSync("git", ["fetch", "origin", baseBranch, "--depth=1", "--no-recurse-submodules"])` call in `restoreConfigFromBase` (`src/github/operations/restore-config.ts`), run with no explicit `cwd`.

## Environment

- Action: `anthropics/claude-code-action@v1`, resolved to commit `be7b93b1907a4abad570368f3c74b6fe3807510b` (v1.0.183, published 2026-07-25) — unchanged since that date, confirmed via the GitHub API.
- `actions/checkout@v6` (last released 2026-07-20, also unchanged).
- Trigger: `pull_request` (`opened`/`synchronize`), `runs-on: ubuntu-latest`, GitHub-hosted.
- Since neither dependency shipped anything new on the day this started, the likely trigger is a `ubuntu-latest` hosted-runner image update changing some git/environment default this action's restore step implicitly depends on (same general class as #1560, different manifestation).

## Timeline

- Started appearing 2026-08-03 across multiple unrelated private repos/workflows using this same action, all pinned to the floating `@v1` tag.
- Hit rate was high and inconsistent per-PR: one PR's review completed cleanly twice in a row; four other PRs in the same short window failed identically 3 times each in a row (original run + 2 manual reruns via `gh run rerun --failed`), all with the exact same two-stage error above.
- `gh run rerun --failed` is not a reliable workaround for this — most reruns reproduced the identical failure.

## Impact

Because the workflow's gate is (correctly, by design) fail-closed on "no verdict produced," every PR hit by this shows as a review **block**, indistinguishable at the check-status level from a real content finding — the only way to tell them apart is reading the raw run log for this exact signature. That's an expensive manual step for every occurrence.

## Suggested fix

1. Ensure `configureGitAuth`/`restoreConfigFromBase`'s git subprocess calls always run with an explicit `cwd: process.env.GITHUB_WORKSPACE` (or equivalent) rather than relying on ambient `process.cwd()`, in case the composite step's working directory isn't always what's assumed.
2. Consider surfacing a distinguishable output (e.g. a specific error code or output field) when the failure is in the action's own git-bootstrap plumbing versus a genuine review-parsing failure, so consumers of this action can tell "infra hiccup, safe to retry" apart from "something is actually wrong with the diff."

Happy to provide additional log excerpts if useful — redacted the repo/PR specifics above since they're from private repos, but the failure signature is exact and directly copy-pasted.

HuzaifaAbdulRehman:
Fixes #1591.

## API surface touched, up front

`fetchDepthArgs(depth: number)` becomes `fetchDepthArgs(depth: number, cwd?: string)`. Existing calls compile and behave identically. `repo-dir.ts` is a new internal module. Nothing in `base-action/` is touched, so the published `@anthropic-ai/claude-code-base-action` surface is unchanged.

## What breaks

`configureGitAuth` and `restoreConfigFromBase` run git with no `cwd`, so both inherit whatever directory the composite step happens to be in. When that directory is not the checkout, git exits 128:

```
Configuring git user...
fatal: not in a git directory
Failed to configure git authentication: ...
    at configureGitAuth (src/github/operations/git-config.ts:41:9)

Restoring .claude, .mcp.json, ... from origin/main (PR head is untrusted)
fatal: not a git repository (or any of the parent directories): .git
##[error]Action failed with error: Command failed: git fetch origin main --depth=1 --no-recurse-submodules
```

The first failure is swallowed by the `try/catch` in `prepareAgentMode`. The second kills the job, so no `structured_output` is produced. For anyone whose review gate is fail-closed on a missing verdict, that reads as a review block, and it looks identical to a real finding. The reporter of #1591 hit it on four PRs in one window, three consecutive runs each including `gh run rerun --failed`, and had to read raw run logs to tell infrastructure failures apart from review results.

## Root cause

The rest of the action locates the checkout through `GITHUB_WORKSPACE`. `src/mcp/install-mcp-server.ts:142` hands the file-ops server `GITHUB_WORKSPACE || cwd` as its `REPO_DIR`, and the CLI permission model scopes edits to `$GITHUB_WORKSPACE` (`src/create-prompt/index.ts:33`, `src/modes/tag/index.ts:132`). These two files are the exception. They take the repository from ambient `process.cwd()` instead.

Every git call in them lacked a working directory:

- `src/github/operations/git-config.ts:41,42,81,88,93,126,127,133,170-172`
- `src/github/operations/restore-config.ts:53,210,310,327,338`
- `src/github/operations/fetch-depth.ts:25`, reached from the restore's fetch

`restoreConfigFromBase` also does its filesystem work relative to `process.cwd()`: `rmSync(".claude-pr")`, the `lstatSync(p)` loop over `SENSITIVE_PATHS`, and `realpathSync(process.cwd())` at line 281.

## What the issue gets wrong

The follow-up comment groups #907, #1236, #1509, #1510, #1559 and #1543 with this one as a single class of "raw git subprocess calls run without an explicit cwd". I read all of them, and that grouping does not hold. #907 and #1236 are credential problems: an installation token without write access, and `git fetch` running before the action configures its own auth under `persist-credentials: false`. Neither turns on the working directory. #1591 is the only report in that list whose symptom is git failing to find a repository.

## What changed

`repositoryDir()` in a new `src/github/operations/repo-dir.ts` returns `process.env.GITHUB_WORKSPACE || process.cwd()`, matching what `install-mcp-server.ts` already does. Both files now pass it explicitly to git.

The filesystem operations in `restoreConfigFromBase` move with the git calls. Anchoring only the git side would leave a restore that fetches in the checkout while snapshotting and deleting somewhere else, which is worse than the crash it replaces. The security restore would report success while the CLI went on reading PR-controlled `.claude/` from the real checkout. That is a silent failure where there is currently a loud one. `shouldSnapshotContent` and `snapshotSensitivePath` take the literal working-tree directory alongside the resolved real path, since the containment checks need the real path and `relative()` needs the literal one.

`fetchDepthArgs` takes an optional `cwd`. `restore-config.ts` passes its directory; `branch.ts` calls it unchanged and keeps inheriting the process directory, so its behaviour is untouched.

## What I did not change

The issue proposes a second change: a distinguishable output so a consumer can tell an infrastructure failure in the action's git plumbing from a real review failure. That is not in here. It needs a decision about what the action exposes, whether a new output, an exit code or an annotation, and it deserves its own PR rather than riding along behind a working-directory fix. This change reduces how often that situation arises, which is not the same as solving it.

`branch.ts:146`, `branch-cleanup.ts` and `fetcher.ts:495` invoke git the same way, with no `cwd`. They are outside the two functions this issue names, `setupBranch` has its own open reports, and moving them is a larger blast radius than one bug fix should carry. Leaving `fetchDepthArgs`'s parameter optional keeps `branch.ts` internally consistent: today it probes and fetches in the same directory, and forcing the probe onto the workspace while the fetch stayed on cwd would break that.

## Behavioural difference

When `GITHUB_WORKSPACE` is set and differs from the process working directory, these operations now act on the workspace rather than on cwd. That is the point of the change, and it is a real difference, not a no-op. It also covers `setupSshSigning`, which writes `gpg.format`, `user.signingkey` and `commit.gpgsign` into the repository config: those land in the workspace checkout now. When the variable is unset, or set to the directory the process is already in, behaviour is byte-for-byte what it was, which is the path every existing test exercises.

I could not reproduce the hosted-runner conditions that make the two diverge, and I am not claiming to know what changed on `ubuntu-latest` on 2026-08-03. What I can show is that the action contains two subsystems that disagree about where the repository is. Point git at a directory that is not the checkout and it fails exactly as reported, with both of the issue's error strings reproduced in a unit test. If a workflow checks out to a subdirectory with the `path:` input, `GITHUB_WORKSPACE` is not the repository either and this does not help that case.

## Test Plan

Two regression tests, one per file. Both set `GITHUB_WORKSPACE` to the temp repository and `chdir` somewhere else, which is the condition the issue describes. The restore test also asserts that nothing is written next to the unrelated working directory, so it fails if the fix anchors git but leaves the filesystem work behind.

Existing tests now delete an ambient `GITHUB_WORKSPACE` in `beforeEach`. Actions sets that variable, so without this the suite would resolve to the action's own checkout instead of the temp repository it builds.

Environment: Ubuntu 26.04 under WSL2, bun 1.4.0, git 2.53.0. CI pins bun 1.2.12 for the test and typecheck jobs.

New tests against unmodified source:

```
$ bun test test/restore-config.test.ts test/git-config.test.ts
(fail) restoreConfigFromBase > restores the checkout named by GITHUB_WORKSPACE, not the process cwd
(fail) git-config > working directory > configures the checkout named by GITHUB_WORKSPACE, not the process cwd
 20 pass
 2 fail

error: Command failed: git diff --name-only -z --relative --ignore-submodules HEAD --
fatal: not in a git directory
```

That is the same command and the same message quoted in the issue.

Full suite, pristine `main` at a874e9e:

```
$ bun test
 957 pass
 9 fail
Ran 966 tests across 57 files.
```

Full suite with this branch:

```
$ bun test
 959 pass
 9 fail
Ran 968 tests across 57 files.
```

The nine failures are identical in both runs, all in `GitHub API client endpoint routing`, each timing out at roughly 5000 ms against the network. They are unrelated to this change and fail the same way before it.

```
$ bun run typecheck
$ tsc --noEmit
(clean)

$ bun run format:check
$ prettier --check .
All matched files use Prettier code style!
```

### DIFF
--- a/src/github/operations/fetch-depth.ts
+++ b/src/github/operations/fetch-depth.ts
@@ -16,16 +16,17 @@ import { execFileSync } from "child_process";
  * lose, so the limit still applies there and large repositories keep the fetch
  * savings it was added for.
  */
-export function fetchDepthArgs(depth: number): string[] {
-  return isShallowRepository() ? [`--depth=${depth}`] : [];
+export function fetchDepthArgs(depth: number, cwd?: string): string[] {
+  return isShallowRepository(cwd) ? [`--depth=${depth}`] : [];
 }
 
-function isShallowRepository(): boolean {
+function isShallowRepository(cwd?: string): boolean {
   try {
     const output = execFileSync(
       "git",
       ["rev-parse", "--is-shallow-repository"],
       {
+        cwd,
         encoding: "utf8",
         stdio: ["ignore", "pipe", "pipe"],
       },
--- a/src/github/operations/git-config.ts
+++ b/src/github/operations/git-config.ts
@@ -11,6 +11,7 @@ import { join } from "path";
 import { homedir } from "os";
 import type { GitHubContext } from "../context";
 import { GITHUB_SERVER_URL } from "../api/config";
+import { repositoryDir } from "./repo-dir";
 
 const SSH_SIGNING_KEY_PATH = join(homedir(), ".ssh", "claude_signing_key");
 
@@ -25,6 +26,7 @@ export async function configureGitAuth(
   user: GitUser,
 ) {
   console.log("Configuring git authentication for non-signing mode");
+  const repoDir = repositoryDir();
 
   // Determine the noreply email domain based on GITHUB_SERVER_URL
   const serverUrl = new URL(GITHUB_SERVER_URL);
@@ -38,8 +40,10 @@ export async function configureGitAuth(
   const botName = user.login;
   const botId = user.id;
   console.log(`Setting git user as ${botName}...`);
-  await $`git config user.name "${botName}"`;
-  await $`git config user.email "${botId}+${botName}@${noreplyDomain}"`;
+  await $`git config user.name "${botName}"`.cwd(repoDir);
+  await $`git config user.email "${botId}+${botName}@${noreplyDomain}"`.cwd(
+    repoDir,
+  );
   console.log(`✓ Set git user as ${botName}`);
 
   await replaceCheckoutCredentials(githubToken, context);
@@ -72,25 +76,29 @@ export async function replaceCheckoutCredentials(
   context: GitHubContext,
 ) {
   const serverUrl = new URL(GITHUB_SERVER_URL);
+  const repoDir = repositoryDir();
 
   // Remove the authorization header that actions/checkout sets
   console.log("Removing existing git authentication headers...");
   const extraheaderKey = `http.${GITHUB_SERVER_URL}/.extraheader`;
   let removedHeader = false;
   try {
-    await $`git config --unset-all ${extraheaderKey}`;
+    await $`git config --unset-all ${extraheaderKey}`.cwd(repoDir);
     removedHeader = true;
   } catch {
     // No extraheader in the local config (expected on the v6+ include layout).
   }
   try {
-    const includePaths =
-      await $`git config --local --get-all include.path`.text();
+    const includePaths = await $`git config --local --get-all include.path`
+      .cwd(repoDir)
+      .text();
     for (const includePath of includePaths.split("\n")) {
       const path = includePath.trim();
       if (!path) continue;
       try {
-        await $`git config --file ${path} --unset-all ${extraheaderKey}`;
+        await $`git config --file ${path} --unset-all ${extraheaderKey}`.cwd(
+          repoDir,
+        );
         removedHeader = true;
       } catch {
         // This include does not define the header; leave it untouched.
@@ -123,14 +131,14 @@ export async function replaceCheckoutCredentials(
       { mode: 0o700 },
     );
     const cleanUrl = `https://${serverUrl.host}/${context.repository.owner}/${context.repository.repo}.git`;
-    await $`git remote set-url origin ${cleanUrl}`;
-    await $`git config credential.helper ${helperPath}`;
+    await $`git remote set-url origin ${cleanUrl}`.cwd(repoDir);
+    await $`git config credential.helper ${helperPath}`.cwd(repoDir);
     console.log("✓ Configured credential helper");
   } else {
     // Update the remote URL to include the token for authentication
     console.log("Updating remote URL with authentication...");
     const remoteUrl = `https://x-access-token:${githubToken}@${serverUrl.host}/${context.repository.owner}/${context.repository.repo}.git`;
-    await $`git remote set-url origin ${remoteUrl}`;
+    await $`git remote set-url origin ${remoteUrl}`.cwd(repoDir);
     console.log("✓ Updated remote URL with authentication token");
   }
 }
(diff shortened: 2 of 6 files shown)


sylvesterkaczmarek (commented):
`GITHUB_WORKSPACE` is not always the checkout directory: `actions/checkout` supports `path: subdir`, in which case the repo is `$GITHUB_WORKSPACE/subdir`. This change would force git and the restore filesystem operations to the parent workspace and reproduce the `not a git repository` failure (or restore the wrong tree), even if ambient cwd was correct. Could `repositoryDir()` derive the actual checkout root rather than assuming workspace root == repo root?

HuzaifaAbdulRehman:
Yes. That's in the PR body already. `path:` is still a miss. MCP file-ops already sets `REPO_DIR` to `GITHUB_WORKSPACE || cwd`, so it has the same hole.

This PR only points `configureGitAuth` and `restoreConfigFromBase` at that directory. If `repositoryDir()` went looking for a nested checkout, those would run in a different tree than MCP. I don't want to do that here.

HuzaifaAbdulRehman:
@ashwin-ant, could you take a look and approve the workflow runs when you have a chance? They are still marked as `action_required`.

Issue #1577, opened 2026-07-31
`--max-turns` enforced exactly at low values, but silently exceeded (60 configured → 73 turns, subtype: success) on a real long-running session


leinardi:
**Describe the bug**

`--max-turns` (passed via `claude_args`) enforces correctly and precisely at low values — verified with a controlled test at `--max-turns 1`. But a real production run configured with `--max-turns 60` completed with `num_turns: 73` and `subtype: "success"` — 13 turns past the configured cap, with no error, no `error_max_turns`, nothing indicating the limit was ever consulted. So the flag appears to fail *open* (silently exceed the budget) rather than fail *closed* (hard error) once a session is long/complex enough, even though the same mechanism is airtight for trivial sessions.

This matters because `--max-turns` is the primary per-run cost/runaway-loop guardrail for automated Claude Code Action usage — a mode where it silently doesn't apply, with no error signal, is an unbounded-cost risk that's invisible to the caller.

**To Reproduce**

Control test (works correctly):
1. Run `anthropics/claude-code-action@v1` (via a thin composite-action wrapper — see workflow yml below) with `claude_args` including `--max-turns 1`, `use_bedrock: true`, on a trivial PR review triggered by an `issue_comment` event.
2. Result: `"num_turns": 2, "subtype": "error_max_turns", "terminal_reason": "max_turns", "errors": ["Reached maximum number of turns (1)"]`. Enforced exactly as documented.

Production run (fails to enforce):
1. Same action, same `claude_args` mechanism, `--max-turns 60`, `use_bedrock: true`.
2. Triggered on a genuinely large review task: a 20-file, ~650-line-changed PR, checked against a large repo-specific review checklist (`REVIEW.md`) injected via `--append-system-prompt-file`, routed to an Opus Bedrock inference profile.
3. Session ran ~16 minutes, made 72 sequential tool calls (confirmed via full JSONL transcript — no batching, single continuous session, one `session_id`, one `system: init` event).
4. Final `result` event: `"num_turns": 73, "subtype": "success", "stop_reason": "end_turn", "terminal_reason": "completed"` — no error at all.

**Expected behavior**

[...]

PR #1762, opened 2026-09-01
fix(base-action): don't reject successful runs when num_turns exceeds maxTurns

### LINKED ISSUE #1758: Post-result num_turns > maxTurns guard can reject batched-tool traces within the round-trip limit
# Post-result num_turns > maxTurns guard can reject batched-tool traces within the model round-trip limit

Follow-up to #1607 and #1577. This report questions the post-result counter comparison, not the typed maxTurns mapping or the need to fail on genuine SDK errors.

At the inspected main revision `a874e9ecd7bb36efdad65429c6b35815f5a08f10` (Claude Code 2.1.251 / declared Agent SDK dependency `^0.3.251`), [run-claude-sdk.ts](https://github.com/anthropics/claude-code-action/blob/a874e9ecd7bb36efdad65429c6b35815f5a08f10/base-action/src/run-claude-sdk.ts#L238-L250) rejects success when `result.num_turns > sdkOptions.maxTurns`. Official agent-loop docs define `maxTurns` as tool-use round trips. Those quantities can differ with batched tool calls.

## Existing local evidence

An existing direct CLI 2.1.251 run requested `--max-turns 14` and returned `success / is_error:false / num_turns:27`. Metadata from the original request only shows:

- one user input, no subagents or resumed input in the analyzed interval;
- 14 distinct main assistant response IDs;
- 13 responses containing tool calls, plus the final response;
- 26 distinct tool calls and 26 matching tool results.

It fits 14 even counting the final response. Counts were obtained by grouping assistant records by `message.id`, deduplicating `tool_use.id`, and matching tool-result IDs within the original request interval. The subsequent resumed request was excluded.

The action's exact predicate rejects this result as an overrun. We reproduced that predicate behavior offline without another model request. This is local CLI metadata plus a wrapper-predicate reproduction, not a freshly instrumented hosted SDK run. The equality of `num_turns` to tool calls plus one in this trace is an observation, not a proposed universal conversion formula.

## Hosted symptom and limits of attribution

[A hosted run at the same action pin](https://github.com/goformx/goformx/actions/runs/33340103680) logs typed `maxTurns:16`, `success / is_error:false / num_turns:28`, zero permission denials and 328658 ms runtime. The native final review is posted, then the guard fails the job. That run's raw trace was not retained, so its exact model-round count remains unknown; it is consistent with the mismatch but is not the proof of the 14/27 counting above.

## Requested correction

The [current regression test](https://github.com/anthropics/claude-code-action/blob/a874e9ecd7bb36efdad65429c6b35815f5a08f10/base-action/test/run-claude-sdk.test.ts#L221-L277) mocks `num_turns:73 / maxTurns:60` without model-response/tool-call structure. Please add a batching-aware fixture and establish which result/stream metric is comparable to `maxTurns`. Preserve typed `maxTurns` enforcement, genuine `error_max_turns`, `is_error` handling and timeouts; do not resolve this by raising budgets or blindly accepting failures.

Reference contract: [official agent-loop documentation](https://code.claude.com/docs/en/agent-sdk/agent-loop). Its turns/budget section describes tool-use round trips, and its result section distinguishes success from `error_max_turns`.

No credentials or raw repository/model transcripts are included. A fresh SDK/model-backed reproduction has not been run as part of this report.

gunjanjaswal:
Fixes #1758.

`run-claude-sdk.ts` re-checked the SDK's result with `num_turns > maxTurns` and threw on a run the SDK had already reported as `subtype: "success", is_error: false`. The two quantities aren't comparable: `maxTurns` bounds model round trips, while `num_turns` counts every turn in the transcript, and batched tool calls inflate it well past the round-trip count. The issue has a concrete trace, a `--max-turns 14` run returning `num_turns: 27`, reported as a success by the SDK and then failed by this guard, leaving the tracking comment stuck.

Dropping the guard doesn't lose genuine enforcement. A real overrun is reported by the SDK as a non-success subtype (`error_max_turns`), which the existing `isSuccess` check (`subtype === "success" && !is_error`) already treats as a failure and throws on. So typed `maxTurns` enforcement, `is_error` handling, and timeouts all stay intact; the only thing removed is the second-guess against a metric that doesn't line up.

### Changes
- Remove the post-result `num_turns > maxTurns` throw in `base-action/src/run-claude-sdk.ts`, with a comment on why the comparison isn't valid.
- Tests: the old fixture asserted the wrong behavior (`num_turns: 73 / maxTurns: 60` rejected). Replaced it with a batching-aware fixture (`num_turns: 27 / maxTurns: 14` stays a success, matching the issue's trace), and added a test that an `error_max_turns` result still fails.

### Testing
- `bun test base-action/test/run-claude-sdk.test.ts` — 5 pass
- `bun run typecheck` clean, `prettier --check` clean

Follow-up to #1607 / #1577. Per the issue, this keeps genuine `error_max_turns` enforcement rather than raising budgets or accepting failures.

### DIFF
--- a/base-action/src/run-claude-sdk.ts
+++ b/base-action/src/run-claude-sdk.ts
@@ -238,16 +238,13 @@ export async function runClaudeWithSdk(
     throw new Error("No result message received from Claude");
   }
 
-  if (
-    resultMessage.subtype === "success" &&
-    !resultMessage.is_error &&
-    sdkOptions.maxTurns !== undefined &&
-    resultMessage.num_turns > sdkOptions.maxTurns
-  ) {
-    const message = `Claude reported a successful result after ${resultMessage.num_turns} turns, exceeding the configured maximum of ${sdkOptions.maxTurns}`;
-    core.error(message);
-    throw new Error(message);
-  }
+  // maxTurns bounds model round trips, but result.num_turns counts every turn in
+  // the transcript, and batched tool calls inflate that well past the round-trip
+  // count. Comparing the two rejected runs the SDK itself reported as successful
+  // within the limit (a --max-turns 14 run returning num_turns 27). A genuine
+  // overrun is reported by the SDK as a non-success subtype (error_max_turns),
+  // which the isSuccess check below already treats as failure, so there is
+  // nothing to re-check against num_turns here.
 
   // subtype "success" with is_error:true means the run errored without producing
   // a real result — treat it as failure so CI does not show a misleading green check.
--- a/base-action/test/run-claude-sdk.test.ts
+++ b/base-action/test/run-claude-sdk.test.ts
@@ -218,15 +218,11 @@ describe("runClaudeWithSdk", () => {
     }
   });
 
-  test("fails closed when a successful result exceeds maxTurns", async () => {
+  test("accepts a successful result whose num_turns exceeds maxTurns (batched tool calls)", async () => {
     const consoleErrorSpy = spyOn(console, "error").mockImplementation(
       () => {},
     );
     const consoleLogSpy = spyOn(console, "log").mockImplementation(() => {});
-    const coreErrorSpy = spyOn(
-      await import("@actions/core"),
-      "error",
-    ).mockImplementation(() => {});
 
     tempDir = await mkdtemp(join(tmpdir(), "claude-sdk-"));
     process.env.RUNNER_TEMP = tempDir;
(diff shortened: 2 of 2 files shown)


base-action/src/run-claude-sdk.ts
```diff
@@ -238,16 +238,13 @@ export async function runClaudeWithSdk(
...
-    const message = `Claude reported a successful result after ${resultMessage.num_turns} turns, exceeding the configured maximum of ${sdkOptions.maxTurns}`;
-    core.error(message);
-    throw new Error(message);
-  }
+  // maxTurns bounds model round trips, but result.num_turns counts every turn in
+  // the transcript, and batched tool calls inflate that well past the round-trip
+  // count. Comparing the two rejected runs the SDK itself reported as successful
+  // within the limit (a --max-turns 14 run returning num_turns 27). A genuine
+  // overrun is reported by the SDK as a non-success subtype (error_max_turns),
+  // which the isSuccess check below already treats as failure, so there is
```
aj-saison: @claude Explain this to me

PR #1772, opened 2026-09-01
fix: tolerate non-JSON error bodies in app token exchange

### LINKED ISSUE #1764: exchangeForAppToken parses error responses as JSON, masking non-JSON failures and skipping workflow-validation skip detection
**Type:** bug (error handling)
**Severity:** medium
**Area:** `src/github/token.ts`
**Effort:** trivial

## Summary

On a non-OK response the token exchange unconditionally parses the body as JSON.
If the body is not JSON - an HTML error page from a corporate proxy, an empty
`502`/`504`, a rate-limit page - `response.json()` throws a `SyntaxError`. That
`SyntaxError` replaces the real failure, so the HTTP status is never reported and
the retry/skip classification never runs.

## Affected code

`src/github/token.ts:125-138`

```ts
if (!response.ok) {
  const responseJson =
    (await response.json()) as AppTokenExchangeErrorResponse;   // <-- can throw

  if (isWorkflowValidationError(response.status, responseJson)) { ... }

  const message = getAppTokenExchangeErrorMessage(responseJson);
  console.error(
    `App token exchange failed: ${response.status} ${response.statusText} - ${message}`,
  );
  throw new Error(message);
}
```

## Impact

1. **Diagnostics are lost.** The user sees
   `Unexpected token '<', "<html>..." is not valid JSON` instead of
   `App token exchange failed: 502 Bad Gateway`. The status code and
   `statusText` - the only actionable information - are never logged.
2. **`WorkflowValidationSkipError` detection is skipped.** `isWorkflowValidationError`
   never gets a chance to run, so a workflow-validation `401` served with a
   non-JSON body is treated as a hard failure instead of the documented graceful
   skip (`run.ts` sets `skipped_due_to_workflow_validation_mismatch`).
3. **Retry semantics change.** `retryWithBackoff`'s `shouldRetry` only
   special-cases `WorkflowValidationSkipError`, so the `SyntaxError` is retried
   three times with backoff against an endpoint that will keep returning HTML.

## Suggested fix

Read the body once as text and parse defensively:

```ts
if (!response.ok) {
  const rawBody = await response.text();
  let responseJson: AppTokenExchangeErrorResponse = {};
  try {
    responseJson = JSON.parse(rawBody) as AppTokenExchangeErrorResponse;
  } catch {
    // Non-JSON error body (proxy HTML page, empty 5xx). Fall back to the
    // status line, which is the only thing worth reporting.
    responseJson = { message: rawBody.slice(0, 500) || response.statusText };
  }

  if (isWorkflowValidationError(response.status, responseJson)) { ... }
  ...
}
```

Also consider including `response.status` in the thrown `Error` message - right
now only the `console.error` line carries it, and that line is skipped whenever
the parse throws.

rajarshidattapy:
Closes #1764.

## Problem

On a non-OK response, `exchangeForAppToken` unconditionally does `await response.json()`. When the body is not JSON — an HTML error page from a proxy, an empty `502`/`503` — that call throws a `SyntaxError` which **replaces** the real failure. Three things go wrong at once:

1. **The status is lost.** The user sees `Unexpected token '<', "<html>..." is not valid JSON` instead of `502 Bad Gateway`. The `console.error` line that carries `response.status`/`statusText` sits *after* the parse, so it never runs.
2. **The workflow-validation skip is missed.** `isWorkflowValidationError` never gets to inspect the body, so a `401` whose body is plain text rather than JSON is treated as a hard failure instead of the intended graceful skip (`skipped_due_to_workflow_validation_mismatch`).
3. **It retries pointlessly.** `retryWithBackoff`'s `shouldRetry` only excludes `WorkflowValidationSkipError`, so the `SyntaxError` is retried three times with backoff against an endpoint that will never return JSON.

## Change

- Add `readErrorResponseBody`: read the body once as text, parse it as JSON if it parses, otherwise fall back to surfacing the body as a `message` string (capped at 500 chars, since a gateway can return an entire HTML page). A body that fails even to read yields `{}`.
- Keep `response.status` and `statusText` in the **thrown** error, not only in the log line. For an empty body the status is the only information available, so without this the failure surfaces as a bare `Unknown error`.

The happy path is untouched — `response.json()` on a successful response is unchanged.

## Effect on the retry path

`isWorkflowValidationError` already inspects `responseJson.message`, so a `401` with a plain-text `workflow validation failed` body now takes the skip path and stops retrying, which is what the surrounding code intends. Genuine `4xx`/`5xx` failures still retry exactly as before.

## Tests

Three tests added to `test/token.test.ts`, all of which **fail on `main`** (verified by stashing the source change and re-running):

- non-JSON `502` body → rejects with `App token exchange failed: 502 Bad Gateway`, retried 3x
- empty `503` body → rejects with the status rather than a bare `Unknown error`
- plain-text `401` workflow-validation body → rejects with `WorkflowValidationSkipError` after a single attempt (previously unreachable, since the parse threw first)

The existing tests are unaffected: they assert on substrings of the thrown message, which the status prefix preserves.

## Verification

- `bun run format:check` — passes on both changed files
- `bun test` — 928 pass / 41 fail, against a `main` baseline of 925 pass / 41 fail. The +3 are the new tests; the 41 failures are pre-existing and specific to my Windows machine (`EPERM ... symlink` needing elevation, POSIX path separators in expectations, `0o600` mode assertions), unrelated to this change.

## Note on the error text

The non-JSON fallback puts a slice of the response body into the error message. That message reaches `core.setFailed` and the tracking comment through paths that already run it through `redactSecrets` (`run.ts`, `comment-logic.ts`), and the 500-char cap bounds how much of a stray page can be echoed.

### DIFF
--- a/src/github/token.ts
+++ b/src/github/token.ts
@@ -53,6 +53,44 @@ function isWorkflowValidationError(
   );
 }
 
+// Cap how much of a non-JSON error body is surfaced. A gateway can answer with
+// a whole HTML page; the opening lines are enough to identify what replied.
+const MAX_ERROR_BODY_CHARS = 500;
+
+/**
+ * Read a failed response's body without letting a non-JSON body mask the
+ * failure.
+ *
+ * The exchange endpoint answers with JSON, but a proxy or gateway in front of
+ * it can return an HTML error page or an empty body. `response.json()` then
+ * throws a SyntaxError that replaces the real failure: the status and
+ * statusText are never logged, `isWorkflowValidationError` never gets to run,
+ * and `retryWithBackoff` retries a response that will never parse.
+ */
+async function readErrorResponseBody(
+  response: Response,
+): Promise<AppTokenExchangeErrorResponse> {
+  let rawBody: string;
+  try {
+    rawBody = await response.text();
+  } catch {
+    return {};
+  }
+
+  try {
+    const parsed: unknown = JSON.parse(rawBody);
+    if (parsed !== null && typeof parsed === "object") {
+      return parsed as AppTokenExchangeErrorResponse;
+    }
+  } catch {
+    // Not JSON. Fall through and surface the body as plain text instead, so
+    // the status line below still reaches the user.
+  }
+
+  const snippet = rawBody.trim().slice(0, MAX_ERROR_BODY_CHARS);
+  return snippet ? { message: snippet } : {};
+}
+
 async function getOidcToken(): Promise<string> {
   try {
     const oidcToken = await core.getIDToken("claude-code-github-action");
@@ -123,8 +161,7 @@ async function exchangeForAppToken(
   );
 
   if (!response.ok) {
-    const responseJson =
-      (await response.json()) as AppTokenExchangeErrorResponse;
+    const responseJson = await readErrorResponseBody(response);
 
     if (isWorkflowValidationError(response.status, responseJson)) {
       const message = getAppTokenExchangeErrorMessage(responseJson);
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
`readErrorResponseBody()` accepts any JSON object and casts it to the expected error shape, but `isWorkflowValidationError()` later calls `.toLowerCase()` on `message`. A 401 gateway response such as `{"message":{"error":"unauthorized"}}` therefore still throws a `TypeError` and masks the HTTP status, which is the failure this PR is trying to prevent. Could the parsed fields be type-checked as strings before use, with malformed JSON shapes falling back to the generic/snippet path?

rajarshidattapy:
Confirmed the issue: malformed JSON shapes can either crash on .toLowerCase() or silently drop the error message, masking the actual HTTP status. I’ll validate the response fields before using them and fall back to the raw body when there’s no usable message, with tests covering these cases.

@sylvesterkaczmarek I'll put a commit fixing them.

PR #1775, opened 2026-09-01
fix: require positive integer line numbers for inline comments

### LINKED ISSUE #1767: create_inline_comment accepts line/startLine of 0, silently collapsing a multi-line comment to single-line
**Type:** bug (input validation)
**Severity:** low-medium
**Area:** `src/mcp/github-inline-comment-server.ts`
**Effort:** trivial

## Summary

`line` and `startLine` are validated with zod's `.nonnegative()`, which permits
`0`. Git diff line numbers are 1-based, so `0` is never valid - and the code
downstream uses plain falsiness checks, so a `0` is silently reinterpreted rather
than rejected.

## Affected code

`src/mcp/github-inline-comment-server.ts:51-66`

```ts
line: z.number().nonnegative().optional()
  .describe("Line number for single-line comments (required if startLine is not provided)"),
startLine: z.number().nonnegative().optional()
  .describe("Start line for multi-line comments (use with line parameter for the end line)"),
```

The two consumers both treat `0` as "absent":

```ts
// line 97
if (!line && !startLine) {
  throw new Error("Either 'line' ... or both 'startLine' and 'line' ... must be provided");
}

// line 137
const isSingleLine = !startLine;
```

## Failure scenarios

**A. `startLine: 0, line: 12`** - schema accepts it. `isSingleLine` evaluates to
`true`, so `params.start_line` and `params.start_side` are never set. The request
the model asked for (a comment spanning lines 0-12) is posted as a *single-line*
comment on line 12. The tool reports success. The model has no way to detect that
its range was discarded.

**B. `line: 0`** - schema accepts it, then line 97 rejects it with
`"Either 'line' for single-line comments or both 'startLine' and 'line' ... must be provided"`,
even though `line` *was* provided. The message sends the model looking for a
missing argument rather than an out-of-range one.

**C.** The same `!startLine` falsiness bug is repeated in the replay path,
`src/entrypoints/post-buffered-inline-comments.ts:126` (`if (c.startLine)`), so a
buffered comment with `startLine: 0` collapses to single-line there too.

Neither `line` nor `startLine` is constrained to an integer either, so `line: 3.5`
is accepted by the schema and rejected by the GitHub API as a 422.

## Suggested fix

Tighten the schema so invalid values are rejected at the boundary with a clear
zod message, rather than being silently reinterpreted:

```ts
line: z.number().int().positive().optional()
  .describe("Line number for single-line comments (required if startLine is not provided)"),
startLine: z.number().int().positive().optional()
  .describe("Start line for multi-line comments (use with line parameter for the end line)"),
```

With `0` and non-integers excluded at the schema, the existing `!line` /
`!startLine` checks become correct as written, so no downstream change is
required.

rajarshidattapy:
Closes #1767.

## Problem

`create_inline_comment` validates `line` and `startLine` with `.nonnegative()`, which permits `0`. Diff line numbers are 1-based, so `0` is never valid — and both consumers test the values for **falsiness** rather than presence, so a `0` is silently reinterpreted instead of rejected.

```ts
if (!line && !startLine) { throw new Error("Either 'line' ... must be provided"); }
const isSingleLine = !startLine;
```

**`startLine: 0, line: 12`** — accepted by the schema. `isSingleLine` is `true`, so `start_line`/`start_side` are never set and the requested 0–12 range is posted as a **single-line** comment on line 12. The tool reports success; the model has no way to see that its range was dropped.

**`line: 0`** — accepted by the schema, then rejected by the handler with *"Either 'line' for single-line comments or both 'startLine' and 'line' ... must be provided"*, even though `line` **was** provided. That sends the model looking for a missing argument rather than an out-of-range one.

Neither field was constrained to an integer either, so `line: 3.5` passed validation and was rejected by the GitHub API as a 422, long after the model had been told nothing was wrong.

## Change

Constrain both fields to `.int().positive()`. With `0` and non-integers excluded at the boundary, the existing `!line` / `!startLine` checks become correct as written, so no handler change is needed.

To make that testable, the tool's input shape moves into `src/mcp/github-inline-comment-schemas.ts`, mirroring the existing pattern in `src/mcp/github-file-ops-schemas.ts` (raw shape for `server.tool`, plus a `z.object(...)` payload schema for tests). The server imports the shape; the handler is untouched.

The extracted shape is otherwise identical to the original — a `diff` against `main` shows only the two intended `.nonnegative()` → `.int().positive()` changes, plus one cosmetic Prettier reflow of the `commit_id` description onto a single line, which now fits after the two-space de-indent.

## Tests

New `test/inline-comment-schema.test.ts`, seven tests. Three of them **fail on `main`** (verified by temporarily restoring `.nonnegative()` and re-running):

- rejects `line: 0`
- rejects `startLine: 0`
- rejects non-integer line numbers

The other four pin behaviour that must not change: positive integers accepted, multi-line ranges accepted, `side` defaulting to `RIGHT`, and both line fields still permitted to be absent — the schema deliberately does not enforce "at least one", so the handler can keep returning its own explanatory message.

## Note on the replay path

`src/entrypoints/post-buffered-inline-comments.ts` has the same falsiness assumption (`if (c.startLine)`). It is unreachable once the schema rejects `0`, since nothing can write such an entry to the buffer, so it is left alone here rather than widening this PR — worth knowing it exists if the schema is ever loosened again.

## Verification

- `bun run format:check` — passes on all three files
- `bun test` — 932 pass / 41 fail, against a `main` baseline of 925 pass / 41 fail. The +7 are the new tests; the 41 failures are pre-existing and specific to my Windows machine (`EPERM ... symlink` needing elevation, POSIX path separators in expectations, `0o600` mode assertions), unrelated to this change.

### DIFF
--- a/src/mcp/github-inline-comment-schemas.ts
+++ b/src/mcp/github-inline-comment-schemas.ts
@@ -3,11 +3,12 @@ import { z } from "zod";
 /**
  * Raw shape passed to `server.tool` for create_inline_comment.
  *
- * `line` and `startLine` are constrained to positive integers. Diff line
- * numbers are 1-based, so 0 is never valid, and the handler tests them for
- * falsiness (`!line`, `!startLine`) rather than for presence - so a 0 that got
- * past the schema would be read as "absent" and silently change the request
- * rather than be rejected.
+ * `line` and `startLine` are constrained to positive integers: diff line
+ * numbers are 1-based, so 0 and fractions are never valid, and a 0 reaching
+ * the handler would be indistinguishable from an omitted argument.
+ *
+ * Rules that involve both fields live in `createInlineCommentPayloadSchema`
+ * below, which is what the handler parses through.
  */
 export const createInlineCommentInputSchema = {
   path: z
@@ -27,7 +28,7 @@ export const createInlineCommentInputSchema = {
     .positive()
     .optional()
     .describe(
-      "Line number for single-line comments (required if startLine is not provided)",
+      "Line number for the comment. Always required: the line for a single-line comment, or the END line of the range when startLine is given.",
     ),
   startLine: z
     .number()
@@ -35,7 +36,7 @@ export const createInlineCommentInputSchema = {
     .positive()
     .optional()
     .describe(
-      "Start line for multi-line comments (use with line parameter for the end line)",
+      "Start line for multi-line comments. Must be provided together with line (the end line), and must not be greater than it.",
     ),
   side: z
     .enum(["LEFT", "RIGHT"])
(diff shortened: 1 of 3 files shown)


sylvesterkaczmarek (commented):
The schema still allows `startLine` without `line`. That passes the handler's `if (!line && !startLine)` check, is treated as a multi-line comment, and sends `line: undefined` to GitHub even though the contract says both values are required. Since validation is being moved into the schema here, could it enforce `line` whenever `startLine` is present (and ideally `startLine <= line`) before the API call?

rajarshidattapy:
@sylvesterkaczmarek
 
 Good catch -{ startLine: 5 } with no line does slip through and ends up sending line: undefined.

Pushing a fix: added a refine on the payload schema that requires line when startLine is set and checks startLine <= line, and the handler now parses through that schema instead of doing its own !line && !startLine check, so there's one place defining the rule. Validation runs before the buffer write too, so a bad call can't get replayed later.

One thing worth noting: server.tool() takes a raw shape, so the cross-field rule can't show up in the JSON schema the model sees - it has to be a runtime check either way. Updated the descriptions to spell out the pairing.

PR #1789, opened 2026-09-03
Allow semicolons in branch names

rolandsusans:
## Problem

`validateBranchName` rejects semicolons, but they are valid per `git-check-ref-format`:

```
$ git check-ref-format --branch 'FIDA-3971;-adjust-mapping'   # exits 0
```

The consequence is larger than a naming inconvenience, because `setupBranch` validates **the PR's own head and base refs**, not just a branch the action would create:

```ts
// src/github/operations/branch.ts, open-PR path
const branchName = prData.headRefName;
validateBranchName(branchName);        // ~line 195
const baseBranch = prData.baseRefName;
validateBranchName(baseBranch);        // ~line 220
```

`prepareTagMode` calls `setupBranch` unconditionally, before the prompt is built and before the model starts. So a repository whose tickets produce branch names like `FIDA-3971;-adjust-mapping` cannot run the action **at all** on those PRs — including a pure review-only run that never creates a branch, never commits, and never pushes. It fails in the prepare step with:

```
Invalid branch name: "FIDA-3971;-adjust-mapping". Branch names must start with an alphanumeric character...
```

There is no input that skips it, and the head ref belongs to the repository rather than to the action.

We hit this on 16 runs in a month in one repo, whose Jira integration puts the ticket id and title into the branch name with a separator the tracker chooses.

## Change

Adds `;` to the whitelist, following the same reasoning already recorded for `#`, `+`, `,`, `@` and `()`:

> All git calls use `execFileSync` (not shell interpolation), so none of these characters carry injection risk.

**That reasoning turned out to be incomplete, and review caught it.** It covers the git operations but not the prompt, where the branch is interpolated into a suggested `Bash(<wrapper> origin <branch>)` and the live grant prefix-matches — so a semicolon there would have become a second command. The second commit single-quotes it at that site; see the discussion below.

The error message gains `semicolons (;)` alongside the other named characters, and a comment records the reasoning in the same style as its neighbours.

## Verification

Two cases added to `test/validate-branch-name.test.ts`, matching the shape of the existing parentheses test, plus a prompt-generation regression test for the quoting.

Checked that the change does not weaken anything else: the separate metacharacter guard still rejects `~ ^ : ? * [ ] \`, spaces and control characters, and the follow-on guards still reject a leading dash, a leading or trailing period, a trailing slash, `//`, `..`, `.lock`, `@{`, and a bare `@`. A name like `semi;colon~tilde` is still rejected, on the tilde.

I was not able to run `bun test` locally (no Bun on this machine), so CI is the first real run of the new cases.

### DIFF
--- a/src/create-prompt/index.ts
+++ b/src/create-prompt/index.ts
@@ -454,7 +454,7 @@ function getCommitInstructions(
           Bash(git commit -m "<message>\\n\\n${coAuthorLine}")`
             : ""
         }
-        - Push to the remote: Bash(${GIT_PUSH_WRAPPER} origin ${branchName})`;
+        - Push to the remote: Bash(${GIT_PUSH_WRAPPER} origin '${branchName}')`;
     }
   }
 }
--- a/src/github/operations/branch.ts
+++ b/src/github/operations/branch.ts
@@ -74,12 +74,15 @@ export function validateBranchName(branchName: string): void {
   //   "_release/v1.2.3"), which previously failed validation as a PR's base branch.
   // Parentheses are valid per git-check-ref-format and commonly appear in branch names that
   //   use Conventional Commit-style scopes (e.g. "feat(parser)-handle-empty-input").
+  // ; is valid per git-check-ref-format and appears in branch names generated from ticket
+  //   titles by some trackers (e.g. "FIDA-3971;-adjust-mapping"), which previously failed
+  //   validation as a PR's head ref and blocked review-only runs on those PRs entirely.
   // All git calls use execFileSync (not shell interpolation), so none of these characters carry injection risk.
-  const validPattern = /^[a-zA-Z0-9@_][a-zA-Z0-9/_.#+,@()-]*$/;
+  const validPattern = /^[a-zA-Z0-9@_][a-zA-Z0-9/_.#+,@();-]*$/;
 
   if (!validPattern.test(branchName)) {
     throw new Error(
-      `Invalid branch name: "${branchName}". Branch names must start with an alphanumeric character, underscore, or '@' and contain only alphanumeric characters, forward slashes, hyphens, underscores, periods, hashes (#), plus signs (+), commas (,), at signs (@), or parentheses.`,
+      `Invalid branch name: "${branchName}". Branch names must start with an alphanumeric character, underscore, or '@' and contain only alphanumeric characters, forward slashes, hyphens, underscores, periods, hashes (#), plus signs (+), commas (,), at signs (@), parentheses, or semicolons (;).`,
     );
   }
 
--- a/test/create-prompt.test.ts
+++ b/test/create-prompt.test.ts
@@ -181,6 +181,30 @@ describe("generatePrompt", () => {
     ); // from review comments
   });
 
+  test("should single-quote the branch in the push instruction", async () => {
+    // The live grant is Bash(<wrapper>:*), a prefix match, so anything the model appends to
+    // that command is permitted and the shell splits it. An unquoted branch carrying a
+    // metacharacter would turn the suggested push into two commands.
+    const envVars: PreparedContext = {
+      repository: "owner/repo",
+      claudeCommentId: "12345",
+      triggerPhrase: "@claude",
+      eventData: {
+        eventName: "issues",
+        eventAction: "opened",
+        isPR: false,
+        issueNumber: "789",
+        baseBranch: "main",
+        claudeBranch: "feat;-scoped-branch",
+      },
+    };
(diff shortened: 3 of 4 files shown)


sylvesterkaczmarek (commented):
`validateBranchName()` now accepts `;`, but the branch is also interpolated unquoted into `Bash(${GIT_PUSH_WRAPPER} origin ${branchName})` in `create-prompt/index.ts`. A branch such as `foo;/usr/bin/env` is valid under this validator and turns that push instruction into a second shell command, so the `execFileSync` safety argument does not cover the Bash path. Please quote/escape the branch there (or keep `;` rejected) and add a prompt-generation regression test.

rolandsusans:
You are right, and my justification was incomplete. Fixed in `d85e9a1`.

I traced the chain to be sure I was fixing the real thing rather than the symptom:

1. `create-prompt/index.ts:457` interpolates the branch into a suggested `Bash(${GIT_PUSH_WRAPPER} origin ${branchName})`.
2. The **live** grant is `modes/tag/index.ts:152` — `` `Bash(${gitPushWrapper}:*)` ``. That is a prefix match, so `<wrapper> origin foo;/usr/bin/env` is *permitted*, not just suggested. (`buildAllowedToolsString` in `create-prompt/index.ts` looks like the grant but is dead — its own NOTE says nothing reads `ALLOWED_TOOLS`.)
3. The Bash tool runs it through a shell, which splits on `;`.

So the `execFileSync` comment covers the git operations and not this path, and `validateBranchName` was the thing standing in the way. My PR would have removed it.

Reachability, for the record: `branchName` there is `claudeBranch || baseBranch`. `claudeBranch` is action-generated and safe; `baseBranch` is the PR's base ref, which is validated by the same call I was loosening (`branch.ts:220`). So a repo where contributors can push branches gives an attacker the input.

**The fix** single-quotes the branch at the interpolation site. That is airtight rather than best-effort here: a quote character cannot appear in a valid branch name — it is outside the whitelist *and* caught by the separate metacharacter guard — so there is nothing to escape past. Verified both directions:

```
branch with a single quote accepted by whitelist?  false
injection branch accepted by whitelist?            true   # hence the quoting
```

Regression test added in `test/create-prompt.test.ts`, asserting the quoted form is emitted and the bare form is not.

Two things I did **not** do, deliberately:

- **I did not touch the `Bash(<wrapper>:*)` grant.** It admits an arbitrary suffix regardless of branch names, so a model that composes its own command is unconstrained by anything in this PR. That looks worth narrowing on its own merits, but it is a pre-existing surface and a bigger change than this one should carry. Happy to open it separately if you want it.
- **I did not quote line 443** (`origin HEAD`) or line 831 (the `<branch>` placeholder), since neither interpolates a ref.

If you would rather keep `;` rejected and drop this entirely, that is a fine outcome too — say so and I will close it. The motivation is narrow: one repo whose tracker generates branch names like `FIDA-3971;-adjust-mapping`, where `setupBranch` validates the head and base refs before the prompt is built, so those PRs cannot get a review-only run at all.

I could not run `bun test` locally (no Bun on this machine), so CI is the first real execution of both new tests.

rolandsusans:
Thanks for the look. I went digging, and it turns out neither of us had the permission model right.

Claude Code doesn't treat `Bash(<wrapper>:*)` as a plain prefix match. It splits the command on shell operators first — `&&`, `||`, `;`, `|`, `|&`, `&`, newlines — and every part has to match an allow rule on its own. So `<wrapper> origin foo;/usr/bin/env` is denied at `/usr/bin/env`. A `;` in a branch name can't turn the push instruction into a second command.

I pushed the quoting fix anyway (`d85e9a1`). It costs nothing, a quote can't appear in a valid branch name so there's nothing to escape past, and it stops the prompt suggesting a command that would only get denied. Regression test asserts the quoted form is emitted and the bare one isn't.

One thing that might be useful either way: `branchName` at that site is `claudeBranch || baseBranch`. `claudeBranch` is action-generated; `baseBranch` is the PR's base ref, validated by the same call this PR loosens. That's why a review-only run on a `;` branch never gets started.

If you'd rather keep `;` rejected, happy to close this. The motivation is narrow — one repo where the issue tracker generates the branch names, and those PRs get no review at all today.

sylvesterkaczmarek (approved):
Re-reviewed `d85e9a1`. The accepted semicolon is now single-quoted in the generated push instruction, and the validator still excludes single quotes, so the accepted branch-name surface cannot break out of that quoting. The prompt regression covers the semicolon case directly, alongside the validator coverage. My earlier concern is resolved; no blocker from me.

PR #1790, opened 2026-09-03
fix: gate mention-triggered workflow examples on author_association

### LINKED ISSUE #1481: `/install-github-app` and `docs/usage.md` examples lack `author_association` guard — self-trigger loop and denial-of-wallet exposure at scale
### Problem

The workflow generated by `/install-github-app` and the example in [`docs/usage.md`](https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md) do not include an `author_association` filter or a `sender.type != 'Bot'` guard in the `if:` condition. This causes two issues:

1. **Self-trigger loop :** The agent's reply comment can contain `@claude`, re-triggering the workflow indefinitely.

2. **Denial-of-wallet surface:** On public repos, any user can comment `@claude` to trigger the workflow. The action's internal `checkWritePermissions` gate (in [`src/github/validation/permissions.ts`](https://github.com/anthropics/claude-code-action/blob/main/src/github/validation/permissions.ts)) blocks non-collaborators from write actions, but the workflow still runs and may consume API tokens before that gate fires at the tool-execution layer.

### Current pattern (from `docs/usage.md`)

The [`docs/usage.md`](https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md) example does not even include an `if:` condition:

```yaml
name: Claude Assistant
on:
  issue_comment:
    types: [created]
jobs:
  claude-response:
    runs-on: ubuntu-latest
    steps:
        with:
```

The `/install-github-app` generated workflow adds a content check but no actor filter:

```yaml
if: |
  (github.event_name == 'issue_comment' && contains(github.event.comment.body, '@claude')) || ...
```

### Suggested addition

```yaml
if: |
  contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association) &&
  github.event.sender.type != 'Bot' &&
  (github.event_name == 'issue_comment' && contains(github.event.comment.body, '@claude')) || ...
```

This prevents both self-trigger loops (bot comments are excluded) and non-collaborator triggers (only OWNER/MEMBER/COLLABORATOR can invoke).

linhongyu510:
## Summary

The shipped `@claude` workflows have no actor filter, so on a public repository any user who can comment can start a run. `checkWritePermissions` does reject the actor — but only after the job has taken a runner and reached the action, so the guard belongs in the workflow `if:` as well.

This adds an `OWNER`/`MEMBER`/`COLLABORATOR` guard to every branch that reads a mention body, plus a regression test.

Fixes #1481.

## Current state

| File | Actor filter today |
| --- | --- |
| `docs/usage.md` | none — the example has no `if:` at all |
| `docs/configuration.md` (CI helper) | none |
| `examples/claude.yml` | trigger-phrase check only |
| `examples/claude-wif.yml` | trigger-phrase check only |
| `.github/workflows/claude.yml` (this repo) | trigger-phrase check only |

## Two things worth flagging

**1. The guard cannot be factored out in front of the chain.** `&&` binds tighter than `||` in [GitHub Actions expressions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions), so a single leading guard attaches only to the first branch:

```
G && (issue_comment...) || (pull_request_review_comment...) || ...
→ (G && issue_comment...) || (pull_request_review_comment...) || ...
```

Every later branch then runs unguarded — i.e. the shape suggested in #1481 would still let an outside user trigger via a review comment. So the guard is repeated inside each branch instead. The added test asserts this directly (guard count == mention-branch count), so the factored-out form cannot be reintroduced silently.

Each event also carries its trigger's association on a different field, which is the second reason a single check does not work:

| Event | Association field |
| --- | --- |
| `issue_comment` | `github.event.comment.author_association` |
| `pull_request_review_comment` | `github.event.comment.author_association` |
| `pull_request_review` | `github.event.review.author_association` |
| `issues` (`opened`) | `github.event.issue.author_association` |

**2. `issues.assigned` / `issues.labeled` are left out of the generic example.** Those triggers key off who assigned or labelled rather than the issue author, so gating them on `issue.author_association` would break `assignee_trigger` / `label_trigger` on issues filed by outside contributors. But shipping an unconditional `action == 'assigned'` branch is worse: `assigneeTrigger` has no default, so `checkContainsTrigger` can never match an assigned event out of the box, and the branch would start a runner for every assignment only to no-op. Both are now omitted from the default `if:` and the `issues:` types, with a comment showing how to opt in — qualified by the configured trigger (`github.event.assignee.login == '...'`) rather than left bare. Thanks to @sylvesterkaczmarek for catching this.

## Scope

Left unchanged on purpose:

- `examples/issue-deduplication.yml`, `base-action/examples/issue-triage.yml` — `prompt`-driven automations that are *meant* to run on outside-authored issues; an insider guard would defeat their purpose.
- `examples/agent-approval-check.yml` — itself a security gate, and its existing `if:` is load-bearing.
- `docs/custom-automations.md` — prose list of supported events, not a workflow.

Also not included: `checkHumanActor` already blocks bot actors by default (`allowed_bots` is opt-in), so the self-trigger-loop half of #1481 is largely handled in code today. I did not add a `sender.type != 'Bot'` clause, since that would duplicate an existing runtime gate rather than close a gap. Happy to add it if you'd prefer defence in depth at the workflow layer.

## Verification

- `bun test` — **981 pass / 0 fail** (2263 assertions, 58 files).
- `bun run format:check` — the 6 files it flags (`ROADMAP.md`, `src/github/token.ts`, and 4 others) are untouched by this PR and already flagged on `main`; every file changed here is clean under Prettier.
- Both new invariants were confirmed to fail without their fix: reverting `examples/claude.yml` to the unguarded form turns 15 pass into 11 pass / 4 fail, and re-adding a bare `assigned` branch turns it into 14 pass / 1 fail.

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -6,20 +6,20 @@ on:
   pull_request_review_comment:
     types: [created]
   issues:
-    types: [opened, assigned]
+    types: [opened]
   pull_request_review:
     types: [submitted]
 
 jobs:
   claude:
     # Only run for repository insiders; see examples/claude.yml for why the
-    # association guard is repeated per branch and why issues.assigned is exempt.
+    # association guard is repeated per branch, and for how to opt into
+    # assignee_trigger / label_trigger.
     if: |
       (github.event_name == 'issue_comment' && contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association) && contains(github.event.comment.body, '@claude')) ||
       (github.event_name == 'pull_request_review_comment' && contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association) && contains(github.event.comment.body, '@claude')) ||
       (github.event_name == 'pull_request_review' && contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.review.author_association) && contains(github.event.review.body, '@claude')) ||
-      (github.event_name == 'issues' && github.event.action == 'opened' && contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.issue.author_association) && (contains(github.event.issue.body, '@claude') || contains(github.event.issue.title, '@claude'))) ||
-      (github.event_name == 'issues' && github.event.action == 'assigned')
+      (github.event_name == 'issues' && github.event.action == 'opened' && contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.issue.author_association) && (contains(github.event.issue.body, '@claude') || contains(github.event.issue.title, '@claude')))
     runs-on: ubuntu-latest
     permissions:
       contents: write
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -12,11 +12,14 @@ on:
 
 jobs:
   claude:
+    # Only run for repository insiders; see examples/claude.yml for why the
+    # association guard is repeated per branch and why issues.assigned is exempt.
     if: |
-      (github.event_name == 'issue_comment' && contains(github.event.comment.body, '@claude')) ||
-      (github.event_name == 'pull_request_review_comment' && contains(github.event.comment.body, '@claude')) ||
-      (github.event_name == 'pull_request_review' && contains(github.event.review.body, '@claude')) ||
-      (github.event_name == 'issues' && (contains(github.event.issue.body, '@claude') || contains(github.event.issue.title, '@claude')))
+      (github.event_name == 'issue_comment' && contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association) && contains(github.event.comment.body, '@claude')) ||
+      (github.event_name == 'pull_request_review_comment' && contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association) && contains(github.event.comment.body, '@claude')) ||
+      (github.event_name == 'pull_request_review' && contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.review.author_association) && contains(github.event.review.body, '@claude')) ||
+      (github.event_name == 'issues' && github.event.action == 'opened' && contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.issue.author_association) && (contains(github.event.issue.body, '@claude') || contains(github.event.issue.title, '@claude'))) ||
+      (github.event_name == 'issues' && github.event.action == 'assigned')
     runs-on: ubuntu-latest
     permissions:
       contents: write
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -160,6 +160,9 @@ permissions:
 
 jobs:
   claude-ci-helper:
+    # Restrict triggering to repository insiders so an outside comment cannot
+    # start a billable run before the action's write-access check rejects it.
+    if: contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association) && contains(github.event.comment.body, '@claude')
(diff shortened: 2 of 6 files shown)


sylvesterkaczmarek (commented):
The new unconditional `issues && action == 'assigned'` branch makes the shipped example start a runner for every issue assignment, even though `assignee_trigger` is commented out/unset. Before this PR, an assignment without `@claude` in the issue title/body did not pass the job `if`. That broadens runner usage while this change is meant to reduce it. Could the generic example omit `assigned` unless an assignee trigger is actually configured, or gate this branch on the configured trigger?

linhongyu510:
@sylvesterkaczmarek You're right, and thank you — that branch made the example strictly worse on the one axis this PR is supposed to improve. Fixed in 0b4f5d0.

Confirming the regression you spotted, comparing the shipped `if:` before this PR against my first commit:

| Scenario | before PR | first commit | |
| --- | --- | --- | --- |
| `assigned`, no `@claude` in issue | skipped | **runs** | widened |
| `assigned`, has `@claude` | runs | runs | same |
| `opened`, no `@claude` | skipped | skipped | same |
| `opened`, has `@claude` | runs | runs | same |

And it was worse than just broader, because those runs could never do anything. In `checkContainsTrigger`:

```ts
if (isIssuesAssignedEvent(context)) {
  let triggerUser = assigneeTrigger.replace(/^@/, "");
  const assigneeUsername = context.payload.assignee?.login || "";
  if (triggerUser && assigneeUsername === triggerUser) { ... return true; }
}
```

`assigneeTrigger` has no default in `action.yml`, so `triggerUser` is empty and this never returns true. Every assignment would have taken a runner to reach a guaranteed no-op.

I took your first option — the generic example omits it. `assigned` is dropped from the `if:` and from the `issues:` types (and `labeled` too in `docs/usage.md`, which had the same shape), replaced with a comment showing how to opt in, qualified by the configured trigger:

```yaml
#   (github.event_name == 'issues' && github.event.action == 'assigned' &&
#    github.event.assignee.login == 'claude-bot')
```

Two tests replace the assertion that previously required an `assigned` branch:

- a bare `assigned`/`labeled` branch not qualified by `assignee.login` / `label.name` fails;
- every action listed in `issues: types:` must have a branch that can match it, so the `on:` block and the `if:` can't drift apart again — which is the class of mistake I made here.

Verified the second test actually catches it: re-adding exactly the branch you flagged turns 15 pass into 14 pass / 1 fail.

One correction to my own verification note above, since it's no longer accurate: I'd reported 9 pre-existing failures in `test/github-graphql-url.test.ts`. Those were a stale local `node_modules` — a dependency re-resolution cleared them, and the suite is now **981 pass / 0 fail**, typecheck clean, Prettier clean on every file this PR touches. The failures were never related to this change, but "9 fail" in the description was a local artifact rather than a property of `main`, and I'd rather flag that than leave it standing.

sylvesterkaczmarek (approved):
The shipped examples no longer subscribe to `issues.assigned` by default, so an unrelated assignment cannot start a runner when `assignee_trigger` is unset. The tests also pin that assignment without an explicit configured branch stays skipped. This resolves the runner-broadening issue I raised.

linhongyu510:
Following up — this one is small and has been sitting since 09-06 with no CI ever having run on it.

@sylvesterkaczmarek caught a real regression in the first pass: my original example added an unconditional `issues && action == 'assigned'` branch, which meant the *shipped* example would start a runner on every issue assignment even with `assignee_trigger` unset — strictly worse on the exact axis this PR is meant to improve. Fixed in `0b4f5d0`, and re-reviewed as resolved on 09-06.

The PR only touches workflow **examples**, gating mention-triggered ones on `author_association` so a drive-by comment from a non-collaborator cannot start a run.

Current state: **0 check runs** on head `0b4f5d00a` — the fork workflow gate has not been triggered, so there is nothing red, just nothing run. An "Approve and run" plus a maintainer look would move it.

PR #1792, opened 2026-09-03
fix: post buffered inline comments as a single review

### LINKED ISSUE #433: Seperate code reviews are started for every inline comment on experimental mode
**Describe the bug**
When using the experimental code review mode, every inline comment has it's own code review "instance", instead of being under one

**To Reproduce**
Create a pull request that causes claude to create >= 2 review comments.

**Expected behavior**
A single review (`github-actions [bot] reviewed 5 minutes ago` ) is created containing multiple inline comments

**Screenshots**

Image

**Workflow yml file**
[Default example experimental workflow](https://github.com/anthropics/claude-code-action/blob/main/examples/claude-experimental-review-mode.yml)

**API Provider**

[ ] AWS Bedrock
[ ] GCP Vertex

**Additional context**

linhongyu510:
## Summary

Buffered inline comments are posted with one `createReviewComment` call each, and GitHub wraps every such call in its own review. A PR with N inline comments therefore shows N separate `reviewed` entries rather than one review containing N comments — the screenshot in #433.

The comments are already collected in a buffer (`/tmp/inline-comments-buffer.jsonl`) and posted together at the end, so only the final call needs to change.

Fixes #433.

## Cause

`src/entrypoints/post-buffered-inline-comments.ts` loops over the batch:

```ts
for (const c of toPost) {
  if (await postComment(octokit, owner, repo, pull_number, headSha, c)) { ... }
}
```

and `postComment` calls `octokit.rest.pulls.createReviewComment` once per entry. That endpoint has no notion of a containing review, so each call creates a standalone one.

## Change

One `pulls.createReview` with a `comments[]` array. Two details that are easy to get wrong:

- **`event: "COMMENT"` requires a non-empty `body`.** The [API docs](https://docs.github.com/en/rest/pulls/reviews?apiVersion=2022-11-28#create-a-review-for-a-pull-request) list `body` as *"**Required** when using `REQUEST_CHANGES` or `COMMENT`"*, but the Octokit typings mark it optional, so omitting it type-checks and then fails at runtime with 422. `REVIEW_BODY` supplies the summary line, and a test asserts it stays non-empty.
- **Per-entry fields differ from the single-comment API.** The review payload takes `path` / `body` / `line` / `side` (+ `start_line` / `start_side` for multi-line hunks) and must *not* repeat `commit_id` or `pull_number`, which are set once at the review level.

## Fallback is conditional, not unconditional

`createReview` is atomic — one unmappable line (outside the diff, or stale after a force-push) rejects the whole batch. Dropping all N comments because one is unmappable would be a worse failure than the bug being fixed, so there is a per-comment fallback. But it only runs when the failure proves nothing was created:

| Failure | Created? | Action |
| --- | --- | --- |
| 4xx except 408/429 | definitively not | replay immediately |
| 5xx, 408, 429, transport error, no status | unknown | reconcile, then replay only if absent |

Reconciliation looks for a review on the head SHA carrying `REVIEW_BODY`. If that lookup itself fails, it does not replay: posting nothing is recoverable by re-running, double-posting N comments is not. Thanks to @sylvesterkaczmarek for catching that the original fallback was unconditional.

## Verification

- `bun test` — **983 pass / 0 fail** (2273 assertions, 59 files); 17 new tests.
- `bun run format:check` — clean on all three files touched here.
- Five of the new tests drive `main()` with a stubbed Octokit and assert on whether the per-comment path runs: response lost + review exists → no replay; ambiguous + no review → replay; 422 → replay with no reconciliation call; reconciliation fails → no replay; success → one review, zero individual posts.
- `main()` is behind `import.meta.main` and exported for those tests. `action.yml:446` runs this file directly with `bun run`, where `import.meta.main` is true, so the production path is unchanged; this matches `prepare.ts`, `run.ts` and 3 other entrypoints.

I could not exercise the real API from here, so the batched call is verified against the documented schema and by unit tests rather than end-to-end; the `body`-required behaviour is the part most worth a maintainer's eye.

An earlier revision of this PR claimed `postComment` was broken on `main` via a doubled `.rest`. That was wrong — `@octokit/rest` self-references, so both forms work — and it was an artifact of an unfaithful test stub. Retracted in f6cc6a7; the stub now mirrors the real client and the change to `github-inline-comment-server.ts` was reverted.

🤖 Written with [Claude Code](https://claude.ai/code).

### DIFF
--- a/src/entrypoints/post-buffered-inline-comments.ts
+++ b/src/entrypoints/post-buffered-inline-comments.ts
@@ -15,6 +15,10 @@ import { redactSecrets } from "../github/utils/sanitizer";
 
 const BUFFER_PATH = "/tmp/inline-comments-buffer.jsonl";
 
+// Summary line for the batched review. event=COMMENT requires a non-empty body,
+// and this is what appears above the grouped inline comments on the PR.
+export const REVIEW_BODY = "Claude finished reviewing this pull request.";
+
 type BufferedComment = {
   ts: string;
   path: string;
@@ -107,6 +111,32 @@ async function classifyComments(bodies: string[]): Promise<boolean[] | null> {
     );
     return null;
   }
+}
+
+/**
+ * Build the `comments[]` payload for a single batched review.
+ *
+ * Exported for testing: the review payload uses a different shape from
+ * createReviewComment (no commit_id/pull_number per entry, and multi-line
+ * hunks use start_line + line), so the mapping is worth pinning down.
+ */
+export function buildReviewComments(
+  comments: BufferedComment[],
+): Array<Record<string, unknown>> {
+  return comments.map((c) => {
+    const side = c.side || "RIGHT";
+    const entry: Record<string, unknown> = {
+      path: c.path,
+      body: redactSecrets(c.body),
+      side,
+    };
+    if (c.startLine) {
+      entry.start_line = c.startLine;
+      entry.start_side = side;
+    }
+    entry.line = c.line;
+    return entry;
+  });
 }
 
 async function postComment(
@@ -218,6 +248,34 @@ async function main() {
   const headSha = pr.data.head.sha;
 
   console.log(`Posting ${toPost.length} classified-as-real comment(s)`);
+
+  // Post as one review so the PR shows a single "reviewed" entry containing
+  // every inline comment, rather than one standalone review per comment.
+  try {
+    await octokit.pulls.createReview({
+      owner,
+      repo,
+      pull_number,
+      commit_id: headSha,
+      event: "COMMENT",
+      // Required by the API whenever event is COMMENT: a review submitted with
+      // event=COMMENT and no body is rejected with 422. The typings mark body
+      // optional, so this is not caught at compile time.
+      body: REVIEW_BODY,
+      comments: buildReviewComments(toPost) as never,
+    });
+    console.log(`Posted ${toPost.length}/${toPost.length} in a single review`);
+    return;
+  } catch (e) {
+    // createReview is atomic: one unmappable line (e.g. outside the diff, or
+    // stale after a force-push) rejects the whole batch. Rather than drop
+    // every comment, fall back to posting them individually — the pre-batch
+    // behaviour — so partial delivery still beats none.
+    console.log(
+      `::warning::Batched review failed (${e instanceof Error ? e.message : String(e)}) — falling back to individual comments`,
+    );
+  }
+
   let posted = 0;
   for (const c of toPost) {
     if (await postComment(octokit, owner, repo, pull_number, headSha, c)) {
@@ -228,7 +286,11 @@ async function main() {
   console.log(`Posted ${posted}/${toPost.length}`);
 }
 
-main().catch((e) => {
-  console.error("post-buffered-inline-comments failed:", e);
-  process.exit(1);
-});
+// Guarded so the module can be imported by tests without running the posting
+// flow, matching the other entrypoints (prepare.ts, run.ts, ...).
+if (import.meta.main) {
+  main().catch((e) => {
+    console.error("post-buffered-inline-comments failed:", e);
+    process.exit(1);
+  });
+}
--- a/src/entrypoints/post-buffered-inline-comments.ts
+++ b/src/entrypoints/post-buffered-inline-comments.ts
@@ -139,6 +139,44 @@ export function buildReviewComments(
   });
 }
 
+/**
+ * Whether a failed createReview definitively means nothing was created.
+ *
+ * 4xx responses (except 408/429) are request-validation failures: GitHub
+ * rejected the payload and committed nothing, so replaying individually is
+ * safe. Transport errors, timeouts and 5xx are ambiguous — the review may have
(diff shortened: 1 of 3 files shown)


sylvesterkaczmarek (commented):
The fallback runs for every `createReview` failure, including ambiguous transport errors after GitHub may already have committed the review. A lost response can therefore make the fallback post every comment a second time; restrict fallback to deterministic request-validation failures (for example 422), or reconcile whether the review was created before replaying individual comments, and add an error-after-success regression test.

linhongyu510:
@sylvesterkaczmarek Good catch — that was a real hole, and fixing it turned up something worse underneath. Addressed in d807385.

**The fallback is now conditional.** Failures are split by whether they prove nothing was created:

| Failure | Created? | Action |
| --- | --- | --- |
| 4xx except 408/429 (e.g. 422 unmappable line) | definitively not | replay immediately, no extra call |
| 5xx, 408, 429, transport error, no status | unknown | reconcile first |

Reconciliation looks for a review on the head SHA whose body carries `REVIEW_BODY`, and skips the replay if it's there. I went with reconcile rather than a strict 422-only allowlist because the 502/timeout case is exactly the one where the comments *did* land and a replay would double-post — an allowlist alone would turn that into "post nothing" rather than "post correctly".

If the reconciliation lookup itself fails, it does **not** replay. Posting nothing is recoverable by re-running; double-posting N comments onto a PR isn't, so unknown state resolves to the safe side.

**Writing the error-after-success test you asked for surfaced a pre-existing bug on `main`.** `postComment` is handed `createOctokit(githubToken).rest` and then dereferences `octokit.rest.pulls.createReviewComment` — a second `.rest`, which is `undefined` at runtime:

```
failed a.ts:1: undefined is not an object (evaluating 'octokit.rest.pulls')
```

So **every individual inline-comment post has been throwing**, with the error swallowed per comment by the surrounding try/catch and reported only as `Posted 0/N`. That's on `main` today, not something this PR introduced — `git show main:...` line 137 has it. It matters here because my fallback would have been dead on arrival, so I've corrected it to `octokit.pulls.*` and the tests now cover that path executing.

Worth flagging separately: this may be a contributing cause of #433 rather than just a neighbouring bug, since it means the buffered path has not been successfully posting at all.

Five new tests drive `main()` with a stubbed Octokit and assert on whether the per-comment path runs: response lost + review exists → no replay; ambiguous + no review → replay; 422 → replay with no reconciliation call; reconciliation fails → no replay; success → exactly one review and zero individual posts. `main()` is exported for this.

**983 pass / 0 fail**, typecheck clean, Prettier clean on all three files.

linhongyu510:
Correction to my previous comment: **the `.rest` bug I reported does not exist.** I was wrong, and I've reverted that part in f6cc6a7.

`@octokit/rest` exposes a self-referencing `.rest`, so both forms resolve to the same function:

```
r.pulls.createReviewComment                                  // function
r.rest.pulls.createReviewComment                             // function
r.pulls.createReviewComment === r.rest.pulls.createReviewComment  // true
```

The `undefined is not an object (evaluating 'octokit.rest.pulls')` error was produced by **my own test stub**, which returned a plain object with no `.rest` self-reference. Correct production code looked broken because the double was unfaithful, not because the code was wrong. I should have checked the real client before concluding, and I regret that it went out as a claim about `main` rather than a question.

Retracting the two consequences of that mistake:

- **`main` is not broken.** Individual inline-comment posting works today, on this file and on `src/mcp/github-inline-comment-server.ts`. I reverted my change to that second file entirely — it was never broken, and I should not have touched it.
- **It is not a contributing cause of #433.** That claim was already impossible on the dates: this file was added 2026-03-12 (#1048), while #433 was filed 2025-08-09. I should have checked that before posting.

What's left in this PR is unchanged and stands on its own:

- batching into one `pulls.createReview` (the actual #433 fix);
- `REVIEW_BODY`, since `event: "COMMENT"` requires a non-empty body;
- the conditional fallback + reconciliation from your review.

The stub now mirrors the real client's `.rest` self-reference, and all five error-after-success tests still pass against it — so the reconciliation behaviour was never dependent on the faulty double. `octokit.pulls.*` is kept in `postComment` purely for consistency with `pulls.get` and `createReview` in the same file; the comment now says so instead of claiming a fix.

**983 pass / 0 fail**, typecheck clean, Prettier clean. PR description updated to remove the incorrect section.

sylvesterkaczmarek (approved):
The ambiguous createReview failure path now reconciles against reviews on the head commit before deciding whether to replay comments, while deterministic 4xx rejections can fall back directly. The error-after-success tests pin the duplicate-posting case. My previous concern is resolved.

linhongyu510:
Following up — approved by @sylvesterkaczmarek on 09-06, and still with **0 check runs** on head `f6cc6a7ed`.

The substance: buffered inline comments were posted as separate API calls, so a partial failure could leave some comments posted and others lost with no way to tell which. They now go out as a single review. The review's remaining concern was the ambiguous `createReview` failure path — the head-commit reconciliation was added so an error that actually succeeded server-side is not retried into a duplicate review.

Nothing red on it; the fork workflow gate simply has not been triggered. An "Approve and run" plus a maintainer pass would unblock it.

PR #1793, opened 2026-09-04
docs: fix checkout fetch-depth example in FAQ

z0rimo:
## Summary

Fixes the full-history checkout example in `docs/faq.md`.

`actions/checkout` uses the `fetch-depth` input under `with`, rather than a top-level `depth` field. The existing example would therefore not configure a full-history checkout as intended.

## Test plan

- Ran `git diff --check`
- Verified `actions/checkout` defines `fetch-depth`, where `0` fetches all history
- Documentation-only change

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -125,7 +125,8 @@ If you need full history, you can configure this in your workflow before calling
 
 ```
 - uses: actions/checkout@v6
-  depth: 0 # will fetch full repo history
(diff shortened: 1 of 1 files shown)


z0rimo:
Hi @ashwin-ant, just checking if you’d have a chance to take a look at this small docs fix. Happy to adjust anything if needed. Thanks!

PR #1794, opened 2026-09-04
fix: read --mcp-config file paths instead of silently dropping them

### LINKED ISSUE #1004: --mcp-config file path in claude_args is silently dropped when action's inline JSON config is present
## Summary

When a user passes `--mcp-config /path/to/file.json` inside `claude_args`, the file path is silently dropped and the custom MCP servers never start. The agent appears to have the tools available (they show up in `allowedTools`) but the MCP server is absent from `mcp_servers` entirely — so the tools are uncallable.

## Root Cause (traced to source)

In `base-action/src/parse-sdk-options.ts`, `mergeMcpConfigs` has two branches:

- **Inline JSON** (starts with `{`) → parsed and merged into `merged.mcpServers`
- **File path** → stored as `lastFilePath`, returned only if `merged.mcpServers` is empty

The action always prepends its own built-in servers (e.g. `github_comment`) as **inline JSON** before user `claude_args` are processed. This means `merged.mcpServers` is never empty by the time user's file path is evaluated. The condition on line 69:

```typescript
if (Object.keys(merged.mcpServers!).length === 0 && lastFilePath) {
  return lastFilePath;
}
```

is never true, so the function falls through to `return JSON.stringify(merged)` — returning only the action's own inline-JSON servers, with the user's file path silently lost.

The code comment acknowledges this but proposes an impossible workaround:

> _"If user passes a file path, they should ensure it includes all needed servers."_

This is not feasible since users cannot know the action's internal server configuration at workflow-write time, and it changes between action versions.

Note also that the function-level docstring contradicts the implementation:

> _"For file paths, they are kept as-is (user's file takes precedence and is used last)."_

## Evidence From CI Logs

Session init (`extraArgs["mcp-config"]`) shows only the action's built-in server — the user's file path is absent:

```json
"mcp-config": "{\"mcpServers\":{\"github_comment\":{...}}}"
```

`mcp_servers` at connection time: custom server is not listed as `FAILED` or `needs-auth` — it is completely absent, confirming the config referencing it was never passed to the SDK.

The `allowedTools` array does include the custom tools (because `--allowedTools` is a separate flag that passes through verbatim), creating a misleading state: tools appear allowed but the server never started.

## Steps to Reproduce

1. Write a custom MCP config to a temp file in a workflow step:
   ```yaml
   - name: Write MCP config
     run: |
       cat > /tmp/my-mcp-config.json <<'EOF'
       {"mcpServers":{"my_server":{"command":"uvx","args":["..."]}}}
       EOF
   ```
2. Pass it via `claude_args`:
   ```yaml
     with:
       claude_args: --mcp-config /tmp/my-mcp-config.json --allowedTools "mcp__my_server__*"
   ```
3. Inspect session init logs — `my_server` will be absent from `mcp_servers`.

## Expected Behavior

The function docstring says:
> _"For file paths, they are kept as-is (user's file takes precedence and is used last)."_

Both the action's built-in inline JSON servers and the user's file-path server should be active in the session.

## Workaround

Pass the custom MCP config as **inline JSON** instead of a file path. Inline JSON is merged correctly:

```yaml
  with:
    claude_args: >-
      --mcp-config '{"mcpServers":{"my_server":{"command":"uvx","args":["..."]}}}'
      --allowedTools "mcp__my_server__*"
```

The "Write MCP config" step can be removed entirely when using this approach.

## Suggested Fix

Read the file at merge time — `existsSync` and `readFileSync` are already imported in `run.ts` and available in the runner environment. When a file path is encountered in `mergeMcpConfigs`, read and parse the file content and merge its `mcpServers` with the rest, rather than storing it as `lastFilePath` and conditionally returning it.

linhongyu510:
## Summary

`--mcp-config /path/to/file.json` inside `claude_args` is silently dropped, so the user's MCP servers never start. Because `--allowedTools` passes through verbatim, the tools appear allowed but are uncallable — the server is simply absent from `mcp_servers`.

Fixes #1004.

## Cause

`mergeMcpConfigs` kept a file path in `lastFilePath` and returned it only if nothing else had been merged:

```ts
if (Object.keys(merged.mcpServers!).length === 0 && lastFilePath) {
  return lastFilePath;
}
```

The action always prepends its own servers (`github_comment`, …) as **inline JSON** before user `claude_args` are parsed, so `merged.mcpServers` is never empty by then. The condition can't fire, and the function falls through to `JSON.stringify(merged)` — returning only the action's servers.

The existing comment acknowledged the gap but suggested users "ensure the file includes all needed servers", which isn't possible: the action's internal server set isn't knowable at workflow-authoring time and changes between versions. The docstring also already claimed the opposite of the behaviour ("*file paths are kept as-is … user's file takes precedence*").

## Reproduced locally

Driving `parseSdkOptions` with the action's real argument shape (single-quoted inline JSON, as `modes/tag/index.ts` builds it) plus a user file path:

```
before:  {"mcpServers":{"github_comment":{…}}}          ← my_server gone
after:   {"mcpServers":{"github_comment":{…},"my_server":{…}}}
```

## Change

Read and parse the file at merge time and merge its `mcpServers` with the rest; later values win on key collisions, matching how inline configs already merge.

Merging is all-or-nothing. When multiple `--mcp-config` values require action-side merging, every inline JSON value and every file must parse successfully; an unreadable or malformed value throws with its source instead of silently running with a partial server set. A single `--mcp-config /path/to/file.json` is still passed through unchanged, preserving the CLI-owned resolution and error path.

## Verification

- Root suite: **973 pass / 0 fail** (2240 assertions, 57 files).
- `base-action`: **171 pass / 0 fail** (345 assertions, 11 files).
- Confirmed the fail-closed tests are load-bearing: reverting only the implementation produces exactly 4 failures.
- New cases cover file + inline, file listed first, collision precedence, two file paths, missing files, malformed JSON, repeated unreadable paths, and single-path CLI passthrough.
- `bun run typecheck` and `bun run format:check` pass in both the root package and `base-action`.

`base-action` is a separately published package; this changes only the internals of an unexported helper, so its public API is unaffected.

🤖 Written with [Claude Code](https://claude.ai/code).

### DIFF
--- a/base-action/src/parse-sdk-options.ts
+++ b/base-action/src/parse-sdk-options.ts
@@ -1,3 +1,4 @@
+import { readFileSync } from "fs";
 import { parse as parseShellArgs } from "shell-quote";
 import type { ClaudeOptions } from "./run-claude";
 import type { Options as SdkOptions } from "@anthropic-ai/claude-agent-sdk";
@@ -60,13 +61,14 @@ type McpConfig = {
 
 /**
  * Merge multiple MCP config values into a single config.
- * Each config can be a JSON string or a file path.
- * For JSON strings, mcpServers objects are merged.
- * For file paths, they are kept as-is (user's file takes precedence and is used last).
+ * Each config can be a JSON string or a file path; both are read and their
+ * mcpServers objects merged, with later values winning on key collisions.
+ * If nothing can be read, the last unreadable path is returned unchanged so
+ * the CLI can report the error against the real path.
  */
 function mergeMcpConfigs(configValues: string[]): string {
   const merged: McpConfig = { mcpServers: {} };
-  let lastFilePath: string | null = null;
+  const unreadableFilePaths: string[] = [];
 
   for (const config of configValues) {
     const trimmed = config.trim();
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
One failure case still looks unsafe: if an explicit `--mcp-config` path is unreadable or malformed, it is warned about and discarded. Because the action normally also contributes built-in inline MCP servers, the merged config can stay non-empty and the run continues without the user's requested server. Could explicit file-path failures remain fatal rather than silently degrading to a partial config?

linhongyu510:
@sylvesterkaczmarek Good catch — the prior behavior was a real fail-open regression. Fixed in 5642bcb.

The merge path is now all-or-nothing:

- a single `--mcp-config path` is still passed through unchanged, so the CLI retains ownership of resolving and reporting that path;
- once there are multiple values and the action must merge them, every inline JSON value and every file must parse successfully;
- an unreadable or malformed explicit file now throws with the file path instead of returning the action's built-in partial config;
- malformed inline JSON also throws rather than being silently skipped.

I replaced the old partial-success assertions with four fail-closed regression cases. Verification:

- root suite: 973 passed; typecheck and format check passed;
- `base-action`: 171 passed; typecheck and format check passed;
- reverting only the implementation while retaining the new tests produces exactly 4 failures.

Thank you for catching this before it shipped.

sylvesterkaczmarek (approved):
Re-reviewed the fail-closed merge path. Once multiple MCP config values must be merged, unreadable files, malformed file JSON, and malformed inline JSON now abort instead of silently producing a partial configuration. A single path is still handed to the CLI unchanged. This resolves the failure mode I raised.

linhongyu510:
Following up — approved by @sylvesterkaczmarek on 09-06, still **0 check runs** on head `5642bcbfa`.

The bug: `--mcp-config` file paths were silently dropped rather than read, so a user passing a config file got no MCP servers and no error explaining why. The review's focus was the fail-closed behaviour once multiple config values have to be merged — an unreadable file now fails loudly instead of degrading to a partial merge, which is the safer default when the alternative is silently running with less configuration than the user asked for.

Same situation as my other two open PRs here: nothing failing, the fork workflow gate just has not fired. An "Approve and run" would let it report.

PR #1802, opened 2026-09-05
fix(base-action): honor RUNNER_DEBUG when deciding to show full output

### LINKED ISSUE #1604: Debug re-runs don't enable full output: isDebugMode checks ACTIONS_STEP_DEBUG, which is never in the step env
## Description

`parseSdkOptions` decides whether to show the full SDK message stream like this ([base-action/src/parse-sdk-options.ts#L195](https://github.com/anthropics/claude-code-action/blob/main/base-action/src/parse-sdk-options.ts#L195)):

```ts
const isDebugMode = process.env.ACTIONS_STEP_DEBUG === "true";
const showFullOutput = options.showFullOutput === "true" || isDebugMode;
```

and the sanitized-output message tells users:

> Rerun in debug mode or enable `show_full_output: true` in your workflow file for full output.

However, re-running a workflow with **"Enable debug logging"** (or `gh run rerun --debug`) does not export an `ACTIONS_STEP_DEBUG` environment variable into step processes. It only enables debug-line rendering runner-side and sets `RUNNER_DEBUG=1` in the step environment — which is what `@actions/core`'s `isDebug()` checks. A code search shows `RUNNER_DEBUG` is not referenced anywhere in this repo.

So the `isDebugMode` branch is effectively dead code, and the log hint sends users down a dead end: they re-run in debug mode expecting a transcript and still get the sanitized stream.

## Reproduction

1. Run any workflow using this action without `show_full_output` (we hit this on `v1.0.182`, pinned by [dyad's PR review workflow](https://github.com/dyad-sh/dyad/blob/main/.github/workflows/claude-pr-review.yml)).
2. Re-run it with "Enable debug logging" checked.
3. The step log shows thousands of `##[debug]` lines (debug mode is genuinely on), but Claude output is still sanitized: `Running Claude Code via SDK (full output hidden for security)...`

Observed in run: https://github.com/dyad-sh/dyad/actions/runs/31030895502 (attempt 4).

## Suggested fix

```ts
const isDebugMode =
  process.env.ACTIONS_STEP_DEBUG === "true" ||
  process.env.RUNNER_DEBUG === "1";
```

(or use `core.isDebug()`). Alternatively, if hiding output on debug re-runs is intentional for security, update the sanitized-output message to stop suggesting a debug re-run will reveal it.

PranshulSoni:
Closes #1604

## Problem

`parseSdkOptions` decides whether a debug re-run should reveal full SDK output with:

```ts
const isDebugMode = process.env.ACTIONS_STEP_DEBUG === "true";
```

but the runner does not export `ACTIONS_STEP_DEBUG` into step environments. Re-running a workflow with **"Enable debug logging"** checked (or `gh run rerun --debug`) sets `RUNNER_DEBUG=1` — the variable `@actions/core`'s own `isDebug()` reads (confirmed in `@actions/core@^1.10.1`: `return process.env['RUNNER_DEBUG'] === '1'`). The `isDebugMode` branch was effectively dead code, and the sanitized-output message pointed users at a dead end: they re-run in debug mode, see thousands of `##[debug]` lines, and still get `full output hidden for security`.

## Solution

Read `RUNNER_DEBUG === "1"` in addition to `ACTIONS_STEP_DEBUG === "true"`. The latter is kept for environments that still export it explicitly, so existing configurations that relied on it do not regress. This also brings the check in line with `@actions/core`'s `isDebug()` semantics.

`show_full_output: true` continues to take precedence as before.

## Testing

- 5 new tests in `base-action/test/parse-sdk-options.test.ts` covering: `RUNNER_DEBUG=1` enables full output, unset/other values do not, `ACTIONS_STEP_DEBUG` still works, and precedence vs. the `show_full_output` input.
- Verified the regression tests fail against the unfixed code (2 failures) and all pass with the fix.
- Full suite: 971 tests pass across 57 files; `bun run typecheck` and `bun run format:check` are clean.

### DIFF
--- a/base-action/src/parse-sdk-options.ts
+++ b/base-action/src/parse-sdk-options.ts
@@ -192,7 +192,12 @@ function parseClaudeArgsToExtraArgs(
  */
 export function parseSdkOptions(options: ClaudeOptions): ParsedSdkOptions {
   // Determine output verbosity
-  const isDebugMode = process.env.ACTIONS_STEP_DEBUG === "true";
+  // The runner exports RUNNER_DEBUG=1 when "Enable debug logging" is checked
+  // (the same variable @actions/core's isDebug() reads). ACTIONS_STEP_DEBUG is
+  // kept for environments where it is still exported explicitly.
+  const isDebugMode =
+    process.env.ACTIONS_STEP_DEBUG === "true" ||
+    process.env.RUNNER_DEBUG === "1";
   const showFullOutput = options.showFullOutput === "true" || isDebugMode;
 
   // Parse claudeArgs into extraArgs for CLI pass-through
--- a/base-action/test/parse-sdk-options.test.ts
+++ b/base-action/test/parse-sdk-options.test.ts
@@ -546,6 +546,88 @@ describe("parseSdkOptions", () => {
     });
   });
 
+  describe("showFullOutput / debug mode", () => {
+    test("should enable full output when RUNNER_DEBUG is 1", () => {
+      // The runner exports RUNNER_DEBUG=1 on re-runs with "Enable debug
+      // logging" checked; ACTIONS_STEP_DEBUG is not set in step envs.
+      const originalEnv = { ...process.env };
+      process.env.RUNNER_DEBUG = "1";
+
+      try {
+        const result = parseSdkOptions({ claudeArgs: "" });
+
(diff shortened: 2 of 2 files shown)


PranshulSoni:
Closing — this duplicates #1612 and #1615, which fix the same issue.

